PR #14, opened 2021-06-11
Installation script for rocksdb

shabbyjoon:


rocksdb_install.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
```
simba-git, line 5: This only works for Linux distributions that use apt-get like ubuntu, maybe check if apt-get exists via something like:
if ! [ -x "$(command -v git)" ]; then

otherwise fallback to the Operating System not supported error

rocksdb_install.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then 
+        sudo apt-get install libgflags-dev
```
simba-git, line 6: all these installs can go on one line:
sudo apt-get install libgflags-dev libsnappy-dev ...

rocksdb_install.sh
```diff
@@ -0,0 +1,14 @@
...
+        sudo apt-get install libsnappy-dev
+        sudo apt-get install zlib1g-dev
+        sudo apt-get install libbz2-dev
+        sudo apt-get install liblz4-dev
+        sudo apt-get install libzstd-dev
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+        brew install rocksdb
```
simba-git, line 13: Let's only brew install the dependencies, we should be able to get them via:
brew deps --tree --1 rocksdb

rocksdb_install.sh
```diff
@@ -0,0 +1,14 @@
...
+fi
```
simba-git, line 14: Let's add an else error case saying that we don't currently support the operating system

simba-git (changes requested):
A few tweaks

rocksdb_install.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$OSTYPE" == "linux-gnu"* ]] && [[-x "$(command -v apt-get)" ]]]; then 
+        sudo apt-get install libgflags-dev libsnappy-dev zlib1g-dev libbz2-dev liblz4-dev libzstd-dev
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+        brew deps --tree --1 rocksdb
```
simba-git, line 8: Not exactly what I meant, run this on your system to get the dependencies, then add a
brew install
line here with all the dependencies that rocksdb needs

We are building rocksdb ourselves in this project, we just need to make sure all its dependencies are installed externally on the system

simba-git (changes requested):

rocksdb_install.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$OSTYPE" == "linux-gnu"* ]] && [[-x "$(command -v apt-get)" ]]]; then
```
simba-git, line 5: Have you tested this? Looks like there is extra brackets

simba-git (changes requested):

simba-git (approved):

PR #32, opened 2021-07-29
Add offline index

simba-git:


sdk/python/offlinehub.py
```diff
@@ -0,0 +1,163 @@
...
+        if key in self._key_to_idx:
+            return self._key_to_idx[key]
+        else:
+            idx = self._next_idx
+            self._next_idx += 1
+            self._key_to_idx[key] = idx
+            self._idx_to_key[idx] = key
+            return idx
+
+    def to_key(self, idx):
```
Sami1309, line 154: perhaps include error handling if index doesn't exist

sdk/python/offlinehub.py
```diff
@@ -0,0 +1,163 @@
...
+        idxs = [idx]
+        # knn_query returns itself, so we get one extra nearest neighbor and drop
+        # idx when mapping the indices.
+        list_results, list_distances = self._idx.knn_query(idxs, num + 1)
+        results = list_results[0]
+        return [
+            self._mapper.to_key(result) for result in results if result != idx
+        ]
+
+    def _add_capacity_to_fit(self, size):
```
Sami1309, line 117: Is self._size changed?

Sami1309 (commented):
Just minor things but I don't understand hnswlib implementation so probably fine.

sdk/python/offlinehub.py
simba-git: What should it do you think? Currently it would just throw an error due to the map

sdk/python/offlinehub.py
simba-git: Great catch, I'll write a test for this as well

sdk/python/offlinehub.py
Sami1309: Would depend on the use case but you're right, Python will just raise a key error and reference the line. Just was unsure about needed robustness of error handling conventions.

simba-git: I think for Python this is fine for now, we could catch and wrap it in our own error or provide another API for a safer to_key that won't throw in the future

Sami1309 (approved):
LGTM!

PR #33, opened 2021-07-29
Fix setup.py to point to the right location

simba-git:
We moved client to sdk/python but didn't update setup.py


Sami1309 (approved):
LGTM

Sami1309 (commented):
lgmt

PR #88, opened 2021-11-03
Local/non local variation for dashboard fetch API

Sami1309:


simba-git (approved):
LGTM, wonder if we can move this into a config file of sorts in the future

PR #93, opened 2021-11-10
Add featureform.com link checker to link checking action

Sami1309:
Makes sure all links in host featureform.com aren't broken. Excludes linkedin links for now due to default 301 response to automatic requests.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -36,11 +36,11 @@ jobs:
           max-depth: 2
   website-link-checker:
     runs-on: ubuntu-latest
-    name: Check featureform website for broken links
+    name: broken-link-check
     steps:
       - name: Check for broken links
         id: link-report
-        uses: celinekurpershoek/link-checker@v1.0.1
+        uses: celinekurpershoek/link-checker@v1.0.0
         with:
           # Required:
           url: "https://www.featureform.com/"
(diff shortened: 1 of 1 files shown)


.github/workflows/link-check.yml
```diff
@@ -31,3 +33,19 @@ jobs:
...
+    steps:
+      - name: Check for broken links
+        id: link-report
+        uses: celinekurpershoek/link-checker@master
+        with:
+          # Required:
+          url: "https://www.featureform.com/"
+          # optional:
+          honorRobotExclusions: false
+          ignorePatterns: "*linkedin*"
```
simba-git, line 48: Can you explain this in a comment

.github/workflows/link-check.yml
```diff
@@ -31,3 +33,19 @@ jobs:
...
+          recursiveLinks: true
+      - name: Print result
+        run: echo "${{steps.link-report.outputs.result}}"
```
simba-git, line 51: Will this just quitely echo the broken links or make it clear to us in some other way?
As in, do we have to keep checking this job?

simba-git (approved):

.github/workflows/link-check.yml
Sami1309: This is to exclude Linkedin links for checking since Linkedin returns an automatic 999 error to web scrapers.

.github/workflows/link-check.yml
Sami1309: The job fails if there are any broken links, so it would be obvious from the outset

PR #94, opened 2021-11-10
Ignore node_modules directory with lint and format

simba-git:
Prior to this, the linter would fail because of C++ files found in
the node_modules directory.


Sami1309 (approved):
LGTM

PR #98, opened 2021-11-15
Basic prometheus charting with chart.js

Sami1309:
Queries prometheus metrics from localhost:9090 and displays on graph.


dashboard/src/components/entitypage/elements/PrometheusGraph.js
```diff
@@ -0,0 +1,51 @@
...
+        plugins: {
+          "datasource-prometheus": {
+            prometheus: {
+              endpoint: "http://localhost:9090",
+            },
+            query: `${query}`,
+            timeRange: {
+              type: "relative",
+
+              start: -parseInt(timeRange.timeRange[0]) * 60 * 1000,
```
simba-git, line 24: Can you make these numbers, named constant, also why is this negative?

simba-git (approved):
Minor change, but looks good otherwise

dashboard/src/components/entitypage/elements/PrometheusGraph.js
Sami1309: type "relative" means relative to time.now, in the past or the future, so historical data would be negative values.

PR #100, opened 2021-11-17
Add metrics API and redefine fetching metrics selection to redux

Sami1309:


dashboard/src/api/resources/Metrics.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  fetchMetrics(instance) {
+    var fetchAddress = `${METRICS_API_URL}/metadata`;
+    return fetch(fetchAddress, {
+      headers: {
+        "Content-Type": "application/json",
+      },
+    })
+      .then((res) => res.json().then((json_data) => json_data))
```
simba-git, line 25: ".then((json_data) => json_data)" this seems like an identity lambda that doesn't do anything, am I missing something?

dashboard/src/api/resources/Metrics.js
```diff
@@ -0,0 +1,30 @@
...
+export default class MetricsAPI {
+  fetchInstances() {
+    var fetchAddress = `${METRICS_API_URL}/label/__name__/values`;
+
+    return fetch(fetchAddress, {
+      headers: {
+        "Content-Type": "application/json",
+      },
+    })
+      .then((res) => res.json().then((json_data) => json_data))
```
simba-git, line 12: Same as below *

simba-git (changes requested):
One question

dashboard/src/api/resources/Metrics.js
Sami1309: yea it's an artifact left over from debugging

PR #104, opened 2021-12-06
Add metrics package and docker compose for full-stack test

Sami1309:


metrics/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /app
+
+COPY go.mod ./
+COPY go.sum ./
+RUN go mod download
+
+COPY *.go ./
+
+RUN go build -o /go-service
```
simba-git, line 11: Minor but can we call it something that isn't this generic :)

metrics/interface.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+//generic interfaces exposed to the user
```
simba-git, line 14: This comment isn't needed

metrics/interface.go
```diff
@@ -0,0 +1,96 @@
+package main
+
+import (
+	//remote packages
```
simba-git, line 4: no need for this comment

metrics/main.go
```diff
@@ -0,0 +1,36 @@
+package main
```
simba-git, line 1: Let rename this file to metrics-example.go or something

metrics/prometheus/prometheus.yml
```diff
@@ -0,0 +1,6 @@
+scrape_configs:
+  - job_name: myapp
+    scrape_interval: 10s
+    metrics_path: /metrics
+    static_configs:
+      - targets: ["host.docker.internal:2112"]
```
simba-git, line 6: What is this host? Is this using the docker network?

simba-git (commented):
Few minor changes and questions

metrics/prometheus/prometheus.yml
Sami1309: It's syntax for how docker for windows exposes localhost. I'll add conventions for mac and linux (which just uses the normal localhost format)

PR #110, opened 2021-12-10
Offline Dataset Working

simba-git:


Sami1309 (approved):
LGTM

ff/serving/dataset_example.py
```diff
@@ -0,0 +1,40 @@
...
+model = tf.keras.models.Sequential([
+    tf.keras.layers.Flatten(input_shape=(28, 28)),
+    tf.keras.layers.Dense(128, activation='relu'),
+    tf.keras.layers.Dropout(0.2),
+    tf.keras.layers.Dense(10)
+])
+
+print("model summary", model.summary())
+predictions = model(x_train[:1]).numpy()
+# print('before',predictions)s
```
shabbyjoon, line 29: We should delete these print statements or update the strings with better descriptions. I can do this later on another branch too.

shabbyjoon (commented):
Left one comment but otherwise looks gucci to me :)

PR #114, opened 2021-12-15
Add metrics instrumentation outline to feature serve

Sami1309:


ff/serving/csv.go
```diff
@@ -45,6 +45,7 @@ func (provider *LocalCSVProvider) GetDatasetReader(key map[string]string) (Datas
 
 func (provider *LocalCSVProvider) ToKey(path string, schema CSVSchema) map[string]string {
 	key := make(map[string]string)
+	logger := provider.Logger.With("Key", key)
```
simba-git, line 48: This doesn't do anything, it logs an empty map

ff/serving/serving.go
```diff
@@ -54,40 +57,50 @@ func NewTrainingDataServer(logger *zap.SugaredLogger) (*TrainingDataServer, erro
 func (serv *TrainingDataServer) TrainingData(req *pb.TrainingDataRequest, stream pb.OfflineServing_TrainingDataServer) error {
 	id := req.GetId()
 	name, version := id.GetName(), id.GetVersion()
+	featureObserver := prom_metrics.BeginObservingFeatureServe(name+" "+version, "training data")
```
simba-git, line 60: run go fmt

simba-git: what happens if name has a space in it or if version has a space in it. Also why make up a string called "training data", not good for to force users to memorize a fixed set of strings that metrics actually uses on the frontend, Try using a set of constants

ff/serving/serving.go
```diff
@@ -54,40 +57,50 @@ func NewTrainingDataServer(logger *zap.SugaredLogger) (*TrainingDataServer, erro
...
 		serv.Logger.Errorw("Dataset error", "Error", err)
+		featureObserver.SetError()
 		return err
 	}
 	return nil
 }
 
 func main() {
 	logger := zap.NewExample().Sugar()
+	prom_metrics = metrics.NewMetrics("test")
```
simba-git, line 101: What's name used for?

metrics/metrics_service_example.go
```diff
@@ -10,7 +10,10 @@ var metrics MetricsHandler
 func testServing() {
 	go func() {
 		for {
+			serveFeature("fixed_acidity", "feature_key")
 			serveFeature("Non-free Sulfur Dioxide", "feature_key")
+			serveFeature("Wine quality set", "Feature Set")
+
```
simba-git, line 16: go fmt

metrics/interface.go
```diff
@@ -1,6 +1,8 @@
 package main
 
 import (
+	//remote packages
```
simba-git, line 4: ?

simba-git (changes requested):

ff/serving/serving.go
Sami1309: A label specifying the instance of metrics observation. In practice should be an environment variable tied to an instance so metrics can be queries based on belonging to an instance or super-set type structure.

shabbyjoon (approved):

PR #120, opened 2022-01-10
Change package name and wrong interface reference in metrics package

Sami1309:
Serving couldn't use the metrics module as the new PR made a reference to a specific struct rather than the FeatureObserver interface. Also renamed package so it isn't read as a program, and moved testing program to its own folder.

Working condition: ff/serving should be able to run and expose prometheus metrics on port 2112.

Package can be pulled and tested in ff/serving via 
`go get github.com/featureform/embeddinghub/metrics@ca2edeff1d10886b748e0983a854ab3dc75a61fb`
long string is the hash of the latest commit

and then run
`go run serving.go csv.go dataset.go metadata.go`


simba-git (changes requested):
Can you run that go get in the directory to make sure it works then commit the go.mod and go.sum in the PR?

PR #128, opened 2022-01-13
Add users page

Sami1309:


dashboard/public/data/lists/wine-data.json
```diff
@@ -415,5 +415,40 @@
...
+      }
+    },
+    {
+      "name": "Shabnam Mohktarani",
+      "default-version": "default-version",
+      "all-versions": ["default-version"],
+      "versions": {
+        "default-version": {
+          "tags": ["Sample model team", "data orchestration"],
+          "permissions": "Edit"
```
simba-git, line 438: No permissions yet

dashboard/public/data/users/Simba Khadder.json
```diff
@@ -0,0 +1,75 @@
...
+            "tags": ["wine", "batch", "sulfur"],
+            "variants": "default"
+          }
+        ],
+        "models": [
+          {
+            "name": "wine random forest",
+            "description": "Random forest wine classification model",
+            "tags": ["wine", "batch", "sulfur"],
+            "variants": "default"
```
simba-git, line 69: Shouldn't variants be a list in this case?

simba-git (changes requested):

dashboard/public/data/users/Simba Khadder.json
Sami1309: Yes. Currently there's no strict implementation of variant select on the object page. Since this change will apply for more than just model I will handle it in another PR.

Sami1309: Added basic variant select to new commit

PR #143, opened 2022-01-28
Dashboard fixes

Sami1309:
Tags should be removed from sublists as well
Labels should have types
timezone should be set to user's time
Revision should be removed everywhere
The inner lists (like when you click a feature) should have the same variant dropdown, note that only variants that the resource uses should be there
Providers do not have variants
Transformations shouldn't be under provider
Datasets need created at
Variant drop down needs to fill in descriptions
Entities do not have variants
Models do not have usage
Lets remove model variants for now
No usage or variant for user
No joined for user
No models subslist under users (models dont have owners)
Add owner to source


simba-git (approved):
Looks good, in the future can you make these separate commits, easier to keep track of

PR #163, opened 2022-02-18
Fix hanging client call on get resources.

simba-git:
The client must CloseSend when its done sending requests, otherwise
the system gets into deadlock.


Sami1309 (approved):
LGTM!

PR #168, opened 2022-02-19
Fix provider notify

simba-git:
Providers only update their list fields if the resource directly
depends on them.


Sami1309 (approved):
LGTM

PR #179, opened 2022-03-11
Integrate typesense search into dashboard

Sami1309:


dashboard/src/api/resources/Resources.js
```diff
@@ -109,17 +109,16 @@ export default class ResourcesAPI {
   }
 
   fetchSearch(query) {
-    const fetchAddress = "/data/lists/wine-data.json";
+    const TypesenseClient = require("./Search.js");
+    const myTypesenseClient = new TypesenseClient("8108", "localhost", "xyz");
+    const myResults = myTypesenseClient.search(
+      { q: query, query_by: "Name" },
```
Sami1309, line 115: Add these to class since they are generic

dashboard/src/api/resources/Resources.js
```diff
@@ -109,17 +109,16 @@ export default class ResourcesAPI {
   }
 
   fetchSearch(query) {
-    const fetchAddress = "/data/lists/wine-data.json";
+    const TypesenseClient = require("./Search.js");
+    const myTypesenseClient = new TypesenseClient("8108", "localhost", "xyz");
```
Sami1309, line 113: no need to use "my"

simba-git, line 113: Better var names, also the API key, port, localhost should be somewhere else

simba-git: Also you should just keep the client created somehwere

dashboard/src/api/resources/Resources.js
simba-git, line 115: Hardcode everything except q into search.js

dashboard/src/api/resources/Resources.js
Sami1309, line 113: make this a static class inside resources class

simba-git (changes requested):

simba-git (approved):

PR #182, opened 2022-03-14
Fix dashboard metrics logic and add tester to metrics with full coverage

Sami1309:


dashboard/src/components/entitypage/elements/PrometheusGraph.js
```diff
@@ -64,10 +68,10 @@ const PrometheusGraph = ({
...
       const endTimestamp = end.getTime() / 1000;
-      const url = `http://localhost:9090/api/v1/query_range?query=${query}${add_labels_string}start=${startTimestamp}&end=${endTimestamp}&step=${step}s`;
-      if (!stub) {
-        return Promise.resolve(JSON.parse(sample_query_data));
-      }
+      const url = `http://localhost:9090/api/v1/query_range?query=${query}${add_labels_string}&start=${startTimestamp}&end=${endTimestamp}&step=${step}s`;
+      // if (!stub) {
+      //   return Promise.resolve(JSON.parse(sample_query_data));
+      // }
       return fetch(url)
```
sdreyer, line 75: A more graceful exception for this would be good. Currently just crashes app

sdreyer (approved):
Should make a future task to add more graceful exceptions when requests fail but otherwise good. Dashboard currently crashes if prometheus not running

dashboard/src/components/entitypage/elements/PrometheusGraph.js
sdreyer: Moved this to an overall comment. Can be fixed in separate PR

PR #184, opened 2022-03-16
Feaure/go coverage/sterling

sdreyer:
Just basic coverage with codecov. Will add formatting and local coverage to separate branch


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,48 @@
...
+      run:
+        working-directory: ./ff/serving
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
```
Sami1309, line 17: some packages use 1.17

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: pip install grpcio-tools
+
+      - name: Install Protobuf
+        run: sudo snap install protobuf --classic
+
+      - name: Setup Proto
+        run: ./gen_grpc.sh
+
+      - name: Build
+        run: go build
```
Sami1309, line 29: some builds are failing in the github actions, should it be sure to terminate the whole job if this happens?

Sami1309 (dismissed):
looks good just some questions

.github/workflows/coverage.yml
sdreyer: Will update

.github/workflows/coverage.yml
sdreyer: Confirmed that any step that fails will cancel entire job. The exception right now is Go tests do to upcoming rework

PR #185, opened 2022-03-16
Feature/local go coverage/sterling

sdreyer:
Runs coverage testing locally and opens in browser


.github/workflows/link-check.yml
```diff
@@ -3,7 +3,7 @@ name: Check All Public Links
 on:
   push:
     branches:
-      - main
+      - client
```
sdreyer, line 6: Mistake, will fix

ff/serving/check_coverage.sh
```diff
@@ -0,0 +1,15 @@
...
+go test -coverpkg=./... -coverprofile cover.out ./...
+
+# Convert Coverage To HTML
+go tool cover -html=cover.out -o coverage.html
+
+# Cleanup Files
+rm cover.out
+
+# Open In Browser
+open coverage.html
```
simba-git, line 15: This may be mac specific

ff/serving/check_coverage.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
simba-git, line 1: Unrelated to this file, but add .idea to .gitignore

ff/serving/check_coverage.sh
sdreyer: Should I submit this first so saadhvi can use it then add support for windows and linux? Or should that all go in one go?

sdreyer: Changed To Be Mac Specific

ff/serving/check_coverage.sh
sdreyer: Added

PR #187, opened 2022-03-18
Create coverage tests for Metadata.go methods 

saadhvi27:


ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	if containsId == true {
+		t.Fatalf("Failed to return false in Has function for the empty case")
+	}
+	listAll, err := lookup.List()
+	if err != nil {
+		t.Fatalf("Failed to list all resources prior to setting a resource %s", err)
+	}
+	if len(listAll) > 0 {
+		t.Fatalf("Failed to get the correct number of results = 0")
```
Sami1309, line 40: add the actual num results in here if error is triggered

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	var resourceType ResourceType = "CSV"
+	listByType, errListForType := lookup.ListForType(resourceType)
+	if errListForType != nil {
+		t.Fatalf("Failed to list resources (by type) prior to setting a resource %s", errListForType)
+	}
+	if len(listByType) > 0 {
+		t.Fatalf("Failed to get the correct number of results (0)")
```
Sami1309, line 48: same here

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,233 @@
+package metadata
+
+import (
+	"github.com/featureform/serving/metadata/search"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/proto"
+	"testing"
+)
+
+// type ResourceLookup interface {
```
simba-git, line 10: Remove and don't commit comments like this if possible

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,233 @@
...
+		t.Fatalf("Failed to set up typesense %s", err)
+	}
+	for _, resource := range testIds {
+		if err := metaTypeServer.lookup.Set(resource, &mockResource{id: resource}); err != nil {
+			t.Fatalf("Failed to Set %s", err)
+		}
+	}
+}
+
+func TestMethodsTypesenseError(t *testing.T) {
```
simba-git, line 219: This test requires Typesense to be running, we'll need tests like this eventually but not yet. I'd remove for now and remove any test that requires typesense to be running

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,233 @@
...
+			Host:   "localhost",
+			Port:   "8108",
+			ApiKey: "xyz",
+		},
+	}
+	testIds := []ResourceID{
+		{
+			Name:    "Fraud",
+			Variant: "default",
+			Type:    ResourceType("CSV"),
```
simba-git, line 197: there is a list of ResourceType s toward the top of metadata.go that you should use. It's in a const ( ) block

simba-git (changes requested):

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1105 @@
...
+	}
+	return response, nil
+}
+
+func TestCreateMetadataserverMethods(t *testing.T) {
+	sugar := zap.NewExample().Sugar()
+	server := Config{
+		Logger: sugar,
+	}
+	metadataServer, err := NewMetadataServer(&server)
```
simba-git, line 761: Can you break all these up into separate tests with cleaner names?

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1105 @@
...
+			t.Fatalf("Failed to get users%s", err)
+		}
+	}()
+	go func() {
+		var v pb.Empty
+		if err := metadataServer.ListUsers(&v, stream); err != nil {
+			t.Fatalf("Failed to list users%s", err)
+		}
+	}()
+	_, err2 := stream.RecvToClient()
```
simba-git, line 783: you should never have numbers like this in your code, this can be inlined with the line below this

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1105 @@
+package metadata
```
simba-git, line 1: So some of the resources like FeatureResource, keep track of the trainingsets they are in. This is updated via the Notify call. Can you test that when a new training set is created, if I get the FeatureResource that it has the TrainingSet in it?

simba-git (changes requested):

ff/serving/metadata/metadata_test.go
sdreyer: @saadhvi27 checkout https://github.com/featureform/embeddinghub/blob/main/ff/serving/provider/offline_test.go as an example.
You can make subtests for each individual test and pass the metadataServer to each of them so you can use dependencies from previous tests.

Sami1309 (commented):
Looks very good. A few things not hit that could easily bump coverage even higher:

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1191 @@
...
+		Label:       &pb.NameVariant{Name: "is_fraud", Variant: "default"},
+		Features:    []*pb.NameVariant{{Name: "user_credit_score", Variant: "default"}},
+	}); err != nil {
+		return nil, err
+	}
+	return metadataServer, nil
+}
+
+func TestGetListUserMethods(t *testing.T) {
+	metadataServer, err := initializeCreate()
```
sdreyer, line 838: @saadhvi27 do you know if you can silence this call: "{"level":"debug","msg":"Creating new metadata server"}" so it doesnt show in the test output constantly?

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1191 @@
...
+func (m *SourceStreamMock) RecvToClient() (*pb.Source, error) {
+	response, more := <-m.sentFromServer
+	if !more {
+		return nil, errors.New("no more to recieve")
+	}
+	return response, nil
+}
+
+func initializeCreate() (*MetadataServer, error) {
+	sugar := zap.NewExample().Sugar()
```
sdreyer, line 757: When you have a chance, replace this with `zap.NewNop().Sugar()` .  As well as for any other zap logger initializations in this file

simba-git: I think we should probably use: https://pkg.go.dev/go.uber.org/zap/zaptest#NewLogger

Take a look at it @sdreyer

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1191 @@
...
+	metadataServer, err := initializeCreate()
+	if err != nil {
+		t.Fatalf("Failed to set up metadata server and create  %s", err)
+	}
+	stream := makeStreamMock()
+	if err := stream.SendFromClient(&pb.Name{Name: "Simba Khadder"}); err != nil {
+		t.Fatalf("Failed to send from client user%s", err)
+	}
+	go func() {
+		if err := metadataServer.GetUsers(stream); err != nil {
```
simba-git, line 847: You don't verify that we actually received the correct user.

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1191 @@
...
+			t.Fatalf("Failed to get users%s", err)
+		}
+	}()
+	go func() {
+		var v pb.Empty
+		if err := metadataServer.ListUsers(&v, stream); err != nil {
```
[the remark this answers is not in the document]
simba-git: You should also verify that it lists what you would expect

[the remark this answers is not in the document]
simba-git: You should check that it lists correctly when there are no resources as well as when there ar emore than one

ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1191 @@
...
+	_, errRecieveTrainingSetVar := streamTrainingSetVar.RecvToClient()
+	if errRecieveTrainingSetVar != nil {
+		t.Fatalf("failed to recieve")
+	}
+	featureVariant, errRecieveFeatureVariant := streamFeatureVariant.RecvToClient()
+	if errRecieveFeatureVariant != nil {
+		t.Fatalf("failed to recieve feature var")
+	}
+	var containsTrainingset bool = false
+	for _, feature := range featureVariant.Trainingsets {
```
simba-git, line 1115: this variable should be called ts, trainingSet, etc. was really confused

simba-git: This is only testing training set, there are a lot of other resources that get things updated

simba-git (changes requested):
The vast majority of these tests just techincally make a call and verify there's no error. They should actually read the values and verify that the values are what you would expect.

PR #189, opened 2022-03-21
Feature/go formatting/sterling

sdreyer:
Includes go linting tests


Sami1309 (dismissed):
Looks good, but go linter fails check. Maybe go linter should only check changed files, but that would guarantee less coverage in the beginning.

.github/workflows/coverage.yml
```diff
@@ -45,4 +45,51 @@ jobs:
...
+          go-version: 1.17
+
+      - name: Install grpc_tools
+        run: pip install grpcio-tools
+
+      - name: Install Protobuf
+        run: sudo snap install protobuf --classic
+
+      - name: Setup Proto
+        run: ./gen_grpc.sh
```
simba-git, line 71: Why?

sdreyer: Needed for the linter. Has import errors without it

simba-git: Ah I see, could you just add that in a comment just so its clear

simba-git (approved):
Just add one comment then good to go

.github/workflows/coverage.yml
sdreyer: Done

PR #192, opened 2022-03-24
Job metadata extension

Sami1309:


ff/serving/metadata/job.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, fmt.Errorf("payload is not job type")
+	}
+	if err := json.Unmarshal(res.Message, jobType); err != nil {
+		return nil, err
+	}
+	return jobType
+	
+}
+
+func (lookup etcdResourceLookup) findJobType(t JobType) (Job, error) {
```
sdreyer, line 31: Might want to change etcdResourceLookup to a different type. This currently implements the ResourceLookup interface which probably isnt necessary for job management. 

Could create something like 
`type etcdJobLookup struct {
	connection EtcdConfig
}`

which would function the same way but doesnt implement ResourceLookup

PR #200, opened 2022-03-29
Add MemoryOfflineStore

simba-git:


ff/serving/provider/offline.go
```diff
@@ -0,0 +1,116 @@
...
+
+type ResourceRecord struct {
+	Entity string
+	Value  interface{}
+	TS     time.Time
+}
+
+func (rec ResourceRecord) Check() error {
+	if rec.Entity == "" {
+		return errors.New("ResourceRecord must have Entity set.")
```
Sami1309, line 48: should it also have a timestamp?

Sami1309: Also should every new table assert that the Value of of each ResourceRecord it contains be the same type?

ff/serving/provider/offline_test.go
```diff
@@ -0,0 +1,102 @@
...
+	invalidIds := []ResourceID{
+		{Type: Feature},
+		{Name: uuid.NewString()},
+	}
+	for _, id := range invalidIds {
+		if _, err := store.CreateResourceTable(id); err == nil {
+			t.Fatalf("Succeeded in creating invalid ResourceID: %v", id)
+		}
+	}
+}
```
Sami1309, line 102: no tests for write or resourcerecord check

simba-git: Coming in another PR, currently you can only write to an OfflineStore, you can't read

ff/serving/provider/offline.go
simba-git: It should have a TS, good catch, the value asserts should come in with Schema which hasn't been implemented yet

PR #202, opened 2022-03-29
Add basic job scheduler and rudimentary testing functions

Sami1309:
I'm splitting my changes to add the job scheduler into separate PRs. #201 is my draft code repo that I'm redoing starting with this one. Steps are as follows

1. (this one) add schedule function to runner interface and a schedule function to the kubernetes client
2. Add a mock kubernetes client to actually test the kubernete functions
3. Abstract the JobClient, Job, CronJob interfaces to they aren't kubernetes specific


ff/serving/runner/kubernetes.go
```diff
@@ -121,7 +127,14 @@ func (k KubernetesRunner) Run() (CompletionWatcher, error) {
 	return KubernetesCompletionWatcher{jobClient: k.jobClient}, nil
 }
 
-func NewKubernetesRunner(config KubernetesRunnerConfig) (Runner, error) {
+func (k KubernetesRunner) Schedule(schedule string) error {
```
simba-git, line 130: Lets make the string a struct with similar fields to the string so its easy to implement for other things

ff/serving/runner/kubernetes.go
```diff
@@ -153,6 +170,14 @@ func (k KubernetesJobClient) Create(jobSpec *batchv1.JobSpec) (*batchv1.Job, err
 	return k.Clientset.BatchV1().Jobs(k.Namespace).Create(context.TODO(), job, metav1.CreateOptions{})
 }
 
+func (k KubernetesJobClient) Schedule(schedule string, jobSpec *batchv1.JobSpec) error {
+	cronJob := &batchv1.CronJob{ObjectMeta: metav1.ObjectMeta{Name: k.JobName, Namespace: k.Namespace}, Spec: batchv1.CronJobSpec{Schedule: schedule, JobTemplate: batchv1.JobTemplateSpec{Spec: *jobSpec}}}
```
sdreyer, line 174: Could wrap these lines for readability

sdreyer (commented):
Seems fine, the long lines of code could be wrapped and indented so you dont have to scroll to see them though

ff/serving/runner/kubernetes.go
```diff
@@ -14,6 +14,15 @@ import (
 
 var namespace string = "default"
 
+type Schedule struct {
+	spec string
```
simba-git, line 18: I meant create the string, which is some random archaic string via fields like Days, Hours, etc.

Otherwise every other CronRunner is gonna have to parse the string

simba-git (changes requested):

ff/serving/runner/kubernetes.go
Sami1309: Changed to what I believe this is addressing in the latest commit

simba-git (approved):

ff/serving/runner/kubernetes.go
```diff
@@ -18,6 +19,54 @@ import (
...
+
+func makeCronSchedule(schedule string) (*CronSchedule, error) {
+	if _, err := cronexpr.Parse(schedule); err != nil {
+		return nil, fmt.Errorf("invalid cron expression: %v", err)
+	}
+	cronSchedule := CronSchedule(schedule)
+	return &cronSchedule, nil
+}
+
+func MonthlySchedule(day, minute, hour int) (*CronSchedule, error) {
```
sdreyer, line 32: Is there an advantage to these over just having a single (day, minute, hour, month, year) function and the ones that can do the */n values?

Sami1309: It's a bit more confusing to do it more generally as the user has to have an "any" parameter that necessarily excludes numbers. This way we can enforce integer inputs.

sdreyer (approved):

PR #203, opened 2022-03-30
Implement Materialization API

simba-git:


sdreyer (approved):
lgtm

ff/serving/provider/offline.go
```diff
@@ -89,13 +111,81 @@ func (store *memoryOfflineStore) CreateResourceTable(id ResourceID) (OfflineTabl
...
 }
 
+// Used to implement sort.Interface for sorting.
+type materializedRecords []ResourceRecord
+
+func (recs materializedRecords) Len() int {
+	return len(recs)
+}
+
+func (recs materializedRecords) Less(i, j int) bool {
```
Sami1309, line 132: I assume this is for sorting the records. Would it make sense to sort them by timestamp (given the null timestamp will always go first), since the entity identifier isn't significant with respect to ordering?

ff/serving/provider/offline.go
```diff
@@ -89,13 +111,81 @@ func (store *memoryOfflineStore) CreateResourceTable(id ResourceID) (OfflineTabl
...
+
+type MaterializationNotFound struct {
+	id MaterializationID
+}
+
+func (err *MaterializationNotFound) Error() string {
+	return fmt.Sprintf("Materialization %s not found", err.id)
+}
+
+func (store *memoryOfflineStore) GetMaterialization(id MaterializationID) (Materialization, error) {
```
Sami1309, line 171: Should the materialization id somehow be tied to the resource id of the table that generates it, so a store's tables and materializations can be connected via some built-in metadata definition?

ff/serving/provider/offline.go
```diff
@@ -124,3 +214,50 @@ func (table *memoryOfflineTable) Write(rec ResourceRecord) error {
...
+
+func newMemoryFeatureIterator(recs []ResourceRecord) FeatureIterator {
+	return &memoryFeatureIterator{
+		data: recs,
+		idx:  -1,
+	}
+}
+
+func (iter *memoryFeatureIterator) Next() bool {
+	isLastIdx := iter.idx == int64(len(iter.data)-1)
```
Sami1309, line 249: styling; put this in the if block

ff/serving/provider/offline.go
Sami1309: I realize this would not necessarily guarantee deterministic ordering, so perhaps in the case of identical timestamps the entity should be the tiebreaker

PR #204, opened 2022-03-30
Generate training set via Offline Store

simba-git:


sdreyer (dismissed):
Nothing major that stands out, a lot to go through though

ff/serving/provider/offline.go
```diff
@@ -106,8 +351,39 @@ func newMemoryOfflineTable() *memoryOfflineTable {
...
+			if i == 0 {
+				return nil
+			}
+			return sortedRecs[i-1].Value
+		} else if i == lastIdx {
+			// Every record happened before the label.
+			return rec.Value
+		}
+	}
+	panic("Unable to getLastValue before timestamp")
```
sdreyer, line 382: Do we want this to panic or should it just return an error?

ff/serving/provider/offline.go
```diff
@@ -89,13 +170,177 @@ func (store *memoryOfflineStore) CreateResourceTable(id ResourceID) (OfflineTabl
...
+	Label    interface{}
+}
+
+type trainingRowsIterator struct {
+	data trainingRows
+	idx  int
+}
+
+func newTrainingRowsIterator(data trainingRows) TrainingSetIterator {
+	return &trainingRowsIterator{
```
Sami1309, line 317: perhaps call this memoryTrainingSetIterator to be more specific on the implementation

ff/serving/provider/offline.go
```diff
@@ -89,13 +170,177 @@ func (store *memoryOfflineStore) CreateResourceTable(id ResourceID) (OfflineTabl
...
+			return err
+		}
+		features[i] = feature
+	}
+	labelRecs := label.records()
+	trainingData := make([]trainingRow, len(labelRecs))
+	for i, rec := range labelRecs {
+		featureVals := make([]interface{}, len(features))
+		for i, feature := range features {
+			featureVals[i] = feature.getLastValueBefore(rec.Entity, rec.TS)
```
Sami1309, line 269: getlastvaluebefore returns nil if no entitymap to the given entity is set, but panics if it is but there are no resources. It should return nil in both cases I think

ff/serving/provider/offline.go
Sami1309: or return nil

ff/serving/provider/offline_test.go
```diff
@@ -44,18 +58,24 @@ func TestOfflineStores(t *testing.T) {
...
 }
 
-func randomResourceID() ResourceID {
-	possibleTypes := []OfflineResourceType{Label, Feature}
-	randType := possibleTypes[rand.Intn(2)]
+func randomID(types ...OfflineResourceType) ResourceID {
+	var t OfflineResourceType
+	if len(types) == 0 {
+		t = NoType
+	} else if len(types) == 1 {
```
Sami1309, line 65: else if and else can be combined since they perform the same logic, t = types[rand.Intn(len(types))]

Sami1309 (approved):
very minor changes, still fine and lgtm tho

ff/serving/provider/offline.go
simba-git: We never get to this line, but I can't return nothing so it is a panic

ff/serving/provider/offline.go
simba-git: Ill clarify the panic in a comment

ff/serving/provider/offline.go
simba-git: good call

ff/serving/provider/offline_test.go
simba-git: true, but I'd argue it a little clearer this way

PR #209, opened 2022-03-31
Implement Materialization runner and copy job factory to run materialization via kubernetes

Sami1309:


ff/serving/runner/worker/worker.go
```diff
@@ -15,11 +16,29 @@ func CreateAndRun() error {
 	if !ok {
 		return errors.New("NAME not set")
 	}
-	runner, err := runner.Create(name, []byte(config))
+	jobRunner, err := runner.Create(name, []byte(config))
 	if err != nil {
 		return err
 	}
-	watcher, err := runner.Run()
+	indexString, hasIndexEnv := os.LookupEnv("JOB_COMPLETION_INDEX")
```
simba-git, line 23: Why isn't this part of config? Why jump through hoops to do the whole SetIndex thing

Sami1309: It's based on one way of dealing with how Kubernetes delegates work to pods. Currently I'm using indexed completion which seperates the job into seperate pods and assigns each one an index in the environment variable job_completion_index. The alternative is to have a single pod with multiple containers, each one with different environment variables corresponding to their indices.

Doing it the single pod way would do away with this awkward convention since the index doesn't have to be separated from the config, but it requires either making multiple jobs or having all the runners  in one pod.

ff/serving/provider/provider.go
```diff
@@ -28,10 +28,7 @@ type RedisConfig struct {
 }
 
 func (r RedisConfig) Serialized() SerializedConfig {
-	config, err := json.Marshal(r)
-	if err != nil {
-		panic(err)
-	}
+	config, _ := json.Marshal(r)
```
simba-git, line 31: Why remove the panic

ff/serving/provider/offline.go
```diff
@@ -235,6 +241,14 @@ func (store *memoryOfflineStore) GetMaterialization(id MaterializationID) (Mater
 	return mat, nil
 }
 
+func (store *memoryOfflineStore) DeleteMaterialization(id MaterializationID) error {
+	if _, has := store.materializations[id]; !has {
```
simba-git, line 245: Can you add a test for this, easy coverage points

ff/serving/runner/copy.go
```diff
@@ -148,3 +143,71 @@ func (m *CopyCompletionWatcher) String() string {
...
+	OnlineConfig   provider.SerializedConfig
+	OfflineConfig  provider.SerializedConfig
+	MaterializedID provider.MaterializationID
+	ResourceID     provider.ResourceID
+	ChunkSize      int64
+	ChunkIdx       int64
+}
+
+func (m *MaterializedChunkRunnerConfig) Serialize() (Config, error) {
+	config, _ := json.Marshal(m)
```
simba-git, line 159: don't drop these errors even if you think they won't throw an error. Someone might add something to config one day and cause marshal to fail and it'll be much harder to catch. Since you don't expect an error you can make it panic instead of returning an error.

ff/serving/runner/copy.go
```diff
@@ -148,3 +143,71 @@ func (m *CopyCompletionWatcher) String() string {
...
+	runnerConfig := &MaterializedChunkRunnerConfig{}
+	if err := runnerConfig.Deserialize(config); err != nil {
+		return nil, fmt.Errorf("failed to deserialize materialize chunk runner config: %v", err)
+	}
+	onlineProvider, err := provider.Get(runnerConfig.OnlineType, runnerConfig.OnlineConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure online provider: %v", err)
+	}
+	offlineProvider, err := provider.Get(runnerConfig.OfflineType, runnerConfig.OfflineConfig)
+	if err != nil {
```
simba-git, line 181: Write tests for the error cases too

ff/serving/runner/materialize.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		kubernetesRunner, err := NewKubernetesRunner(kubernetesConfig)
+		if err != nil {
+			return nil, err
+		}
+		cloudWatcher, err = kubernetesRunner.Run()
+		if err != nil {
+			return nil, err
+		}
+	case Local:
```
simba-git, line 114: I would have made LocalMaterializeRunner and KubernetesMaterializeRunner, this is definitely a case for different classes.

ff/serving/provider/offline.go
```diff
@@ -125,7 +127,7 @@ type ResourceRecord struct {
 
 func (rec ResourceRecord) check() error {
 	if rec.Entity == "" {
-		return errors.New("ResourceRecord must have Entity set.")
+		return errors.New("resourceRecord must have Entity set.")
```
sdreyer, line 130: This can stay uppercase since it's the name of an object. You're IDE will probably continue to give you warnings

ff/serving/provider/provider.go
```diff
@@ -46,9 +43,13 @@ func (r *RedisConfig) Deserialize(config SerializedConfig) error {
 type Provider interface {
 	AsOnlineStore() (OnlineStore, error)
 	AsOfflineStore() (OfflineStore, error)
+	Type() Type
```
sdreyer, line 46: run gofmt at some point

ff/serving/provider/provider.go
Sami1309: added back

ff/serving/provider/offline.go
Sami1309: added

ff/serving/runner/copy.go
Sami1309: redone

ff/serving/runner/copy.go
Sami1309: added tests

ff/serving/runner/materialize.go
Sami1309: changed names

ff/serving/provider/offline.go
Sami1309: fixed

ff/serving/provider/provider.go
Sami1309: fixed

PR #210, opened 2022-04-01
Featureform/python client sdk/shabnam

shabbyjoon:


ff/serving/python_client/sdk/transformations.py
```diff
@@ -0,0 +1,49 @@
...
+        self.ft_type = ft_type
+
+def transformation(*args, **kwargs):
+
+    def add_transformation(definition_func):
+
+        transformation = Transformations(kwargs['feature'], definition_func.__name__ ,
+                                            kwargs['variant'], kwargs['inputs'], 
+                                            kwargs['entity'], kwargs['fn_type'], kwargs['executor'],  kwargs['frequency'], kwargs['ft_type'])
+        definition_func()
```
simba-git, line 22: Why do you run it here?

ff/serving/python_client/sdk/transformations.py
```diff
@@ -0,0 +1,49 @@
...
+        self.entity = entity
+        self.fn_type = fn_type
+        self.frequency = frequency
+        self.ft_type = ft_type
+
+def transformation(*args, **kwargs):
+
+    def add_transformation(definition_func):
+
+        transformation = Transformations(kwargs['feature'], definition_func.__name__ ,
```
simba-git, line 19: Could we figure out how to actually save the python code from the function, in this case:
    count = 10000000000
    return count

ff/serving/python_client/sdk/transformations.py
shabbyjoon: I don't need to run it here, actually. I had been playing around with different implementations of decorators in python and forgot to remove this line. 

We'd run it inside the inner function if we call the transformation decorator like so:
transformation(user_transaction_count)() 

https://www.geeksforgeeks.org/decorators-with-parameters-in-python/

ff/serving/python_client/sdk/register.py
```diff
@@ -0,0 +1,115 @@
+
+files_to_register = []
```
Sami1309, line 2: Since these are so similar imo they should be in a dictionary of lists (unregisteredResources{}), which the key being which resource type (entity, feature, etc).

ff/serving/python_client/sdk/register.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, key = str):
+        self.key = key
+
+def register_entity(name):
+    entity = Entity(name)
+    entities_to_register.append(entity)
+    return entity
+
+class Feature:
+    def __init__(self, name, description, src, column, entity, executor, cast, frequency):
```
Sami1309, line 30: why no type hints on this one?

ff/serving/python_client/sdk/register.py
```diff
@@ -0,0 +1,115 @@
...
+
+class File:
+    def __init__(self, path = str, name = str, index = str, storage = str):
+        self.path = path
+        self.name = name
+        self.index = index
+        self.storage = storage
+
+def register_file(*args, **kwargs):
+    newFile = File(args, kwargs['name'], kwargs['index'], kwargs['storage'])
```
Sami1309, line 16: args is a tuple, so this saves the path as a tuple with a single value

ff/serving/python_client/sdk/register.py
```diff
@@ -0,0 +1,115 @@
...
+        self.name = name
+        self.description = description
+        self.src = src
+        self.column = column
+        self.entity = entity
+        self.executor = executor
+        self.cast = cast
+        self.frequency = frequency
+
+def register_feature_from_source(*args, **kwargs):
```
Sami1309, line 40: args not used here

Sami1309 (changes requested):
looks good, just a few minor issues

ff/serving/python_client/sdk/providers.py
```diff
@@ -0,0 +1,58 @@
...
+def register_deltalake(*args, **kwargs):
+    register_provider(kwargs['name'], 'DeltaLake', kwargs['host'], kwargs['db'], kwargs['ssl'], kwargs['client_certificate'], kwargs['bucket'])
+
+def register_s3(*args, **kwargs):
+    register_provider(kwargs['name'], 'S3', kwargs['host'], kwargs['db'], kwargs['ssl'], kwargs['client_certificate'],kwargs['bucket'])
+
+def register_spark(*args, **kwargs):
+    register_provider(kwargs['name'], 'Spark', kwargs['host'], kwargs['db'], kwargs['ssl'], kwargs['client_certificate'], kwargs['bucket'])
+
+def register_flink(*args, **kwargs):
```
simba-git, line 42: You only use args, kwargs when there is a dynamic set of args or key-word args, not just always.

Here you could do:
def register_flink(name=None, host=None, ...

ff/serving/python_client/sdk/providers.py
```diff
@@ -0,0 +1,58 @@
...
+        self.client_certificate = client_certificate
+        self.bucket = bucket
+      
+
+def register_provider(name, provider_type, host, db, ssl, client_certificate, bucket):
+    provider = Provider(name, provider_type, host, db, ssl, client_certificate, bucket)
+    providers_to_register.append(provider)
+
+def register_redis(*args, **kwargs):
+    register_provider(kwargs['name'], 'Redis', kwargs['host'], kwargs['db'], kwargs['ssl'], kwargs['client_certificate'], kwargs['bucket'])
```
simba-git, line 19: We have the redis config on the backend, @Sami1309 wrote it.

You should match these parameters with those.

ff/serving/python_client/sdk/providers.py
```diff
@@ -0,0 +1,58 @@
+providers_to_register = []
+
+class Provider:
+    def __init__(self, name = str, provider_type = str, host = str, db = int, ssl = bool, client_certificate= str, bucket = str):
+        self.name = name
+        self.provider_type = provider_type
+        self.host = host
+        self.db = db
+        self.ssl = ssl
+        self.client_certificate = client_certificate
```
simba-git, line 10: Every provider will have different parameters. For example, it doesn't make sense for anything except S3 to have a bucket parameter.

ff/serving/python_client/sdk/register.py
```diff
@@ -0,0 +1,115 @@
...
+training_set_to_register = []
+
+class File:
+    def __init__(self, path = str, name = str, index = str, storage = str):
+        self.path = path
+        self.name = name
+        self.index = index
+        self.storage = storage
+
+def register_file(*args, **kwargs):
```
simba-git, line 15: register_file(path, name = "", index = "", storage = "")

simba-git (commented):
kwargs and args should only be used when there really is a dynamic amount of parameters that can be of any name. For example a function that can sum any number of argument mgiht be:
sum(*args) but if it only sums two then it should be: sum2(a, b)

PR #213, opened 2022-04-04
New serving implementation

simba-git:


ff/serving/serving/serving.go
```diff
@@ -0,0 +1,166 @@
...
+		obs.SetError()
+		return nil, err
+	}
+	entity, has := entityMap[meta.Entity()]
+	if !has {
+		logger.Errorw("Entity not found", "Entity", meta.Entity())
+		obs.SetError()
+		return nil, fmt.Errorf("No value for entity %s", meta.Entity())
+	}
+	providerEntry, err := meta.FetchProvider(serv.Metadata, ctx)
```
Sami1309: maybe not necessary now but for the process of feature serving it would be good to do a simple in-memory cache of the providers/stores/tables if it's ever a use case to make really big feature requests with lots of entities/features

ff/serving/newserving/main/main.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	serv, err := newserving.NewFeatureServer(meta, promMetrics, logger)
+
+	grpcServer := grpc.NewServer()
+	if err != nil {
+		logger.Panicw("Failed to create training server", "Err", err)
+	}
+	pb.RegisterFeatureServer(grpcServer, serv)
+	logger.Infow("Serving metrics", "Port", metrics_port)
```
sdreyer, line 39: Whats the difference between Info and Infow?

simba-git: Infow, Debugw, etc. take varags, they are K-V pairs so this logs a message: Serving metrics followed by a JSON object of context including a key "Port" which is set to metrics_port's value

ff/serving/main.go
```diff
@@ -0,0 +1,4 @@
+package main
+
+func main() {
```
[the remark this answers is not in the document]
simba-git: Here's what is solves: https://github.com/featureform/embeddinghub/runs/6014000412?check_suite_focus=true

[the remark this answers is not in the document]
sdreyer: Gotcha

ff/serving/newserving/main/main.go
sdreyer: Got it

sdreyer (approved):
No Issues standing out. Very clean

PR #214, opened 2022-04-05
Postgres Offline Store

sdreyer:
Still cleaning up some convoluted/reused code and restructuring the file, but wanted to get this up for feedback. 

Would like to move this file into the postgres directory, but was unsure if that is currently being used. Can be moved during the repo reorg.


ff/serving/provider/online_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
```
sdreyer, line 4: This file was added by accident. These changes should already be merged

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+// Allows storage of any type and simpler table creation
+type postgresTableItem struct {
+	Value    interface{} `json:"value"`
+	ItemType string      `json:"type"`
+}
+
+type postgresOfflineStore struct {
+	conn             *pgxpool.Pool
+	ctx              context.Context
+	materializations map[MaterializationID]*memoryMaterialization //Are these used?
```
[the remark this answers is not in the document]
simba-git: They shouldn't be if they are

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+	trainingSets     map[ResourceID]trainingRows
```
simba-git, line 25: Same with these

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+
+	value, err := table.serialize(rec.Value)
+	if err != nil {
+		return err
+	}
+	if exists, err := table.resourceExists(rec); err != nil {
+		return err
+	} else if !exists {
+		query := fmt.Sprintf("INSERT INTO %s (entity, value, ts) VALUES ($1, $2, $3) RETURNING entity, value, ts", tb)
```
simba-git, line 384: UPSERT is a thing in postgres and snowflake now I think

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+		"( SELECT * FROM "+
+		"( SELECT *, row_number() over() FROM %s )t1 WHERE row_number>$1 AND row_number<=$2)t2", sanitize(mat.tableName))
+	rows, err := mat.conn.Query(context.Background(), query, start, end)
+	defer rows.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	data := make([]ResourceRecord, 0)
+	for rows.Next() {
```
simba-git, line 484: Let's wrap this in a FeatureIterator?

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+		return nil, err
+	}
+	return &postgresOfflineStore{
+		conn: conn,
+		ctx:  ctx,
+	}, nil
+}
+
+func (store *postgresOfflineStore) getResourceTableName(id ResourceID) string {
+	return fmt.Sprintf("resource_%s", id.Name)
```
simba-git, line 86: let's add featureform_ to all these table names for now

simba-git (commented):
Great work, looks like some brutal SQL work

.github/workflows/integration-tests.yml
```diff
@@ -1,6 +1,12 @@
 name: Integration Testing
```
sdreyer, line 1: Will remove these changes as well. Should already be included

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+	tableCreateQry := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s AS (SELECT entity, value, ts FROM %s WHERE 1=2)", sanitizedTableName, resTableName)
+
+	_, err = store.conn.Exec(
+		context.Background(), tableCreateQry) // Set correct types
+	if err != nil {
+		return nil, err
+	}
+
+	materializeQry := fmt.Sprintf(
+		"INSERT INTO %s SELECT entity, value, ts FROM "+
```
Sami1309, line 179: alternative approach for nested select without partition is 
SELECT DISTINCT ON (entity) entity, ts, value ORDER BY ts DESC

ff/serving/provider/postgres.go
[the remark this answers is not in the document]
sdreyer: Forgot about those

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+
+	return &postgresMaterialization{
+		id:        matID,
+		conn:      store.conn,
+		tableName: matTableName,
+	}, nil
+
+}
+func (store *postgresOfflineStore) GetMaterialization(id MaterializationID) (Materialization, error) {
+	getMatQry := fmt.Sprintf("SELECT DISTINCT (table_name) FROM information_schema.tables WHERE table_name LIKE 'materialization_%s%%'", string(id))
```
Sami1309, line 196: given that you used sanitize to create the materialization table name, should the id pushed through getMaterializationTableName and sanitize be used for this query instead of hardcoding "materialization_" and using LIKE? What if a materialization matches another for the first n characters?

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+	rows, err := store.conn.Query(context.Background(), getMatQry)
+	defer rows.Close()
+	if err != nil {
+		return nil, err
+	}
+	rowCount := 0
+	var tableName string
+	for rows.Next() {
```
Sami1309, line 204: why scan through every row here?

ff/serving/provider/postgres.go
sdreyer: done

Sami1309, line 86: Doesn't distinguish tables by variant or id them by type

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+		tableJoinAlias := fmt.Sprintf("t%d", i)
+		columns = append(columns, resourceTableName)
+		query = fmt.Sprintf("%s LEFT JOIN LATERAL (SELECT entity , value as %s, ts  FROM %s WHERE entity=l.entity and ts <= l.ts ORDER BY ts desc LIMIT 1) %s on %s.entity=l.entity ",
+			query, resourceTableName, resourceTableName, tableJoinAlias, tableJoinAlias)
+		if i == len(def.Features)-1 {
+			query = fmt.Sprintf("%s )", query)
+		}
+	}
+	columnStr := strings.Join(columns, ", ")
+	fullQuery := fmt.Sprintf("CREATE TABLE %s AS (SELECT %s, l.value  FROM %s ", sanitize(tableName), columnStr, query)
```
Sami1309, line 244: l.value should be aliased as the label resource table name

Sami1309 (changes requested):
looks very good, quite the PR

ff/serving/provider/postgres.go
sdreyer: Yep, INSERT ... ON CONFLICT

ff/serving/provider/postgres.go
sdreyer: You mean having it get the rows individually from the db instead of requesting them in bulk and placing into memory?

ff/serving/provider/postgres.go
sdreyer: Done

ff/serving/provider/postgres.go
```diff
@@ -0,0 +1,501 @@
...
+
+func (store *postgresOfflineStore) tableExists(id ResourceID) (bool, error) {
+	var n int64
+	var tableName string
+	if id.check(Feature, Label) == nil {
+		tableName = store.getResourceTableName(id)
+	} else if id.check(TrainingSet) == nil {
+		tableName = store.getTrainingSetName(id)
+	}
+	err := store.conn.QueryRow(context.Background(), "select 1 from information_schema.tables where table_name=$1", tableName).Scan(&n)
```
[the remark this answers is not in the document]
sdreyer: Is there any advantage to using that here? Comparisons after the query will essentially still look the same

ff/serving/provider/postgres.go
sdreyer: Will probably leave this for now since its working fine

ff/serving/provider/postgres.go
sdreyer: Appended type and variant to table names

ff/serving/provider/postgres.go
sdreyer: Leftover from testing. Replaced with a prepared statement

ff/serving/provider/postgres.go
sdreyer: Wrapped it up

ff/serving/provider/postgres.go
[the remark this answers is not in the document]
Sami1309: they are the same pretty much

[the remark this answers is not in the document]
sdreyer: Gotcha, will keep in mind

ff/serving/provider/offline.go
```diff
@@ -157,7 +173,7 @@ func (store *memoryOfflineStore) AsOfflineStore() (OfflineStore, error) {
 	return store, nil
 }
 
-func (store *memoryOfflineStore) CreateResourceTable(id ResourceID) (OfflineTable, error) {
+func (store *memoryOfflineStore) CreateResourceTable(id ResourceID, schema SerializedSchema) (OfflineTable, error) {
```
sdreyer, line 176: Can I get some feedback on if this is an okay way to deal with the schema? (Passing it to CreateResourceTable)

Sami1309: by schema you mean SQL schema?

Sami1309 (approved):

ff/serving/provider/offline.go
sdreyer: Table schema, will change name to clarify

PR #219, opened 2022-04-10
Add serve and stop functions to metadata server.

simba-git:


ff/serving/metadata/metadata.go
```diff
@@ -801,11 +806,37 @@ func NewMetadataServer(server *Config) (*MetadataServer, error) {
...
+	}
+	grpcServer := grpc.NewServer()
+	pb.RegisterMetadataServer(grpcServer, serv)
+	serv.grpcServer = grpcServer
+	serv.Logger.Infow("Server starting", "Address", serv.address)
+	return grpcServer.Serve(lis)
+}
+
+func (serv *MetadataServer) GracefulStop() {
+	serv.grpcServer.GracefulStop()
```
Sami1309, line 831: Does this throw an error if there is no grpc server currently running?

Sami1309 (commented):
lgtm

ff/serving/metadata/metadata.go
simba-git: No, but the internally grpcServer doesn't either. I just mimicked it's behavior

PR #220, opened 2022-04-10
Fix metadata client hanging

simba-git:
The GetResourceVariant calls would hang forever. They did not properly
close the send stream. In general, we should just call the
GetResourceVariants method anyway to avoid duplicating logic.


Sami1309 (approved):
lgtm

PR #222, opened 2022-04-12
Fix TrainingSetDef shorthand

simba-git:
A TrainingSetDef's Features and Labels may not have their types set (since they are implied). This makes it more natural to manually write TrainingSetDefs. Prior to this PR, it wasn't working. This fixes the bug and adds a test for it.


Sami1309 (approved):
lgtm

PR #223, opened 2022-04-12
Metrics get count functions to interface

Sami1309:


ff/serving/metrics/test/metrics_test.go
```diff
@@ -75,22 +75,41 @@ func TestMetrics(t *testing.T) {
 		fmt.Println("error", err)
 	}
 	assert.Equal(t, int(servingCounterValue), servingNum, "5 feature rows should be served")
-
+	servingCounterValueInt, err := servingObserver.GetObservedRowCount()
+	if err != nil {
+		fmt.Println("error", err)
```
sdreyer, line 80: If you change this to t.log() it'll show up above the test results in the end summary which will help with clarity

ff/serving/metrics/interface.go
```diff
@@ -126,6 +127,22 @@ func (p PromFeatureObserver) Finish() {
 	p.Count.WithLabelValues(p.Name, p.Feature, p.Key, "success").Inc()
 }
 
+func (p PromFeatureObserver) GetObservedRowCount() (int, error) {
+	var m = &dto.Metric{}
+	if err := p.Count.WithLabelValues(p.Name, p.Feature, p.Key, "row serving").Write(m); err != nil {
```
sdreyer, line 132: Should the parameter that "row serving" is passed to accept an enum instead of a plain string?

Sami1309: yes, in fact since it is queried from the front end with it as an explicit label value is should be defined once

simba-git (approved):

PR #224, opened 2022-04-12
Snowflake Offline Store

sdreyer:
Working support for snowflake as an offline store.
Still fixing minor things, but the bulk is here.


ff/serving/provider/snowflake.go
```diff
@@ -0,0 +1,626 @@
...
+		return 0, err
+	}
+	return n, nil
+}
+
+func (mat *snowflakeMaterialization) IterateSegment(start, end int64) (FeatureIterator, error) {
+	query := fmt.Sprintf(""+
+		"SELECT entity, value, ts FROM "+
+		"( SELECT * FROM "+
+		"( SELECT *, row_number() over(ORDER BY (SELECT NULL)) as row_number FROM %s )t1 WHERE row_number>? AND row_number<=?)t2", sanitize(mat.tableName))
```
Sami1309, line 193: this doesn't guarantee an order, so is it guaranteed that there will be no overlapping between segments given the order by isn't deterministic? If this is verified with testing ignore this comment

ff/serving/provider/offline_test.go
```diff
@@ -65,6 +86,47 @@ func TestOfflineStores(t *testing.T) {
...
+}
+
+func destroySnowFlakeDatabase(c SnowflakeConfig) error {
+	url := fmt.Sprintf("%s:%s@%s-%s", c.Username, c.Password, c.Organization, c.Account)
+	db, err := sql.Open("snowflake", url)
+	if err != nil {
+		return err
+	}
+	databaseQuery := fmt.Sprintf("DROP DATABASE IF EXISTS %s", sanitize(c.Database))
+	if _, err := db.Exec(databaseQuery); err != nil {
```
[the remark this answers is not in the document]
sdreyer: The "IF EXISTS" allows it to return successfully if the database doesnt exist

ff/serving/provider/offline_test.go
```diff
@@ -155,49 +217,49 @@ func testMaterializations(t *testing.T, store OfflineStore) {
 		},
 		"NoOverlap": {
 			WriteRecords: []ResourceRecord{
-				{Entity: "a", Value: 1},
-				{Entity: "b", Value: 2},
-				{Entity: "c", Value: 3},
+				{Entity: "a", Value: 1, TS: time.UnixMilli(0).UTC()},
```
simba-git, line 220: We should not have to do this, your code should add it.

ff/serving/provider/snowflake.go
```diff
@@ -0,0 +1,626 @@
...
+	"time"
+)
+
+// snowflakeColumnType is used to specify the column type of a resource value.
+type snowflakeColumnType string
+
+const (
+	SFInt    snowflakeColumnType = "integer"
+	SFNumber                     = "NUMBER"
+	SFFloat                      = "float8"
```
simba-git, line 20: 8 bytes or bits? Is this a double?

ff/serving/provider/snowflake.go
```diff
@@ -0,0 +1,626 @@
...
+	tableName string
+}
+
+func (mat *snowflakeMaterialization) ID() MaterializationID {
+	return mat.id
+}
+
+func (mat *snowflakeMaterialization) NumRows() (int64, error) {
+	n := int64(0)
+	query := fmt.Sprintf("SELECT COUNT(*) FROM %s", sanitize(mat.tableName))
```
simba-git, line 180: If you add row_number to materialization you could also make this a MAX isntead of a count which is faster

ff/serving/provider/offline_test.go
```diff
@@ -1,17 +1,45 @@
...
+	snowFlakeDatabase := strings.ToUpper(uuid.NewString())
+	var snowflakeConfig = SnowflakeConfig{
+		Username:     os.Getenv("SNOWFLAKE_USERNAME"),
+		Password:     os.Getenv("SNOWFLAKE_PASSWORD"),
+		Organization: os.Getenv("SNOWFLAKE_ORG"),
+		Account:      os.Getenv("SNOWFLAKE_ACCOUNT"),
+		Database:     snowFlakeDatabase,
+	}
+	serialSFConfig := snowflakeConfig.Serialize()
+	if err := createSnowFlakeDatabase(snowflakeConfig); err != nil {
```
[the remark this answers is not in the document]
sdreyer: Good eye

ff/serving/provider/snowflake.go
```diff
@@ -0,0 +1,626 @@
...
+	"strings"
+	"time"
+)
+
+// snowflakeColumnType is used to specify the column type of a resource value.
+type snowflakeColumnType string
+
+const (
+	SFInt    snowflakeColumnType = "integer"
+	SFNumber                     = "NUMBER"
```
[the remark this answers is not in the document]
sdreyer: Snowflake uses integer to create the  column, but when reading the column it returns the type as NUMBER.

ff/serving/provider/offline_test.go
```diff
@@ -165,9 +214,9 @@ func testMaterializations(t *testing.T, store OfflineStore) {
 			SegmentStart: 0,
 			SegmentEnd:   3,
 			ExpectedSegment: []ResourceRecord{
-				{Entity: "a", Value: 1},
-				{Entity: "b", Value: 2},
-				{Entity: "c", Value: 3},
+				{Entity: "a", Value: 1, TS: time.UnixMilli(0).UTC()},
```
simba-git, line 217: Can you remove all these from the tests

ff/serving/provider/snowflake.go
```diff
@@ -0,0 +1,631 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+// snowflakeColumnType is used to specify the column type of a resource value.
+type snowflakeColumnType string
+
+const (
+	SFInt    snowflakeColumnType = "integer"
```
simba-git, line 18: We really should have generic type schemas and they should map into these things. That'd also allow us to make this consts be private

simba-git: I know you're working on it, and you should merge this, just want to make sure it doesn't get forgotten

simba-git (approved):
LGTM, great work

ff/serving/provider/offline_test.go
sdreyer: Right now have to leave them in for the "Expected" results. When the resource is written, its set to start of epoch time if TS wasnt initialized. So when the test result is returned, uninitialized resource TS will be queried and contain the start of epoch time, which requires TS: time.UnixMilli(0).UTC() to match. Only way to avoid this is to not scan TS, which we can't do, or remove timestamps that equal start of epoch which seems like odd behavior.

ff/serving/provider/snowflake.go
sdreyer: Yeah I was going to rework a lot of the Type stuff when doing the generic tables/transformations. Will get it in the next PR

simba-git (approved):

PR #225, opened 2022-04-13
Adds ServeOnListener to Metadata Server

simba-git:
In some cases (like testing) it'd be nice to provide our own listener to serve metadata on.


Sami1309 (approved):
lgtm

PR #230, opened 2022-04-15
Remove Main files from coverage

sdreyer:


.github/workflows/testing.yml
```diff
@@ -240,7 +240,7 @@ jobs:
         if: always()
         working-directory: ./coverage
         run: |
-          cat cover.out.tmp | grep -v "proto" > cover.out
+          cat cover.out.tmp | grep -v "proto" | grep -v "main" | grep -v "metadata/client.go" > cover.out
```
simba-git, line 243: Why remove metadata.client?

simba-git (changes requested):

.github/workflows/testing.yml
sdreyer: Its a file sam was using to test the dashboard , which creates a server, then loads data in. Pretty irrelevant to coverage as is. Should probably be converted into an actual test file for integration testing or removed.

simba-git (approved):

PR #232, opened 2022-04-18
Metadata tests

simba-git:
Currently a draft, have about 80% coverage, gonna clean up the test code quite a bit and add more cases though

### DIFF
--- a/ff/serving/metadata/metadata_test.go
+++ b/ff/serving/metadata/metadata_test.go
@@ -1,0 +1,1510 @@
+package metadata
+
+import (
+	"context"
+	"net"
+	"reflect"
+	"testing"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+)
+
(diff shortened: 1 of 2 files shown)


ff/serving/metadata/metadata_test.go
```diff
@@ -0,0 +1,1446 @@
...
+	config := &Config{
+		Logger:          logger.Sugar(),
+		StorageProvider: LocalStorageProvider{},
+	}
+	serv, err := NewMetadataServer(config)
+	if err != nil {
+		panic(err)
+	}
+	// listen on a random port
+	lis, err := net.Listen("tcp", ":0")
```
Sami1309, line 272: is this not a random port?

Sami1309 (approved):
lgtm

ff/serving/metadata/metadata_test.go
simba-git: It is

PR #236, opened 2022-04-20
Add licenses to ff files

simba-git:


Sami1309 (approved):
Should do some integration testing with this PR

PR #239, opened 2022-04-21
Sandbox Terraform

sdreyer:


ff/serving/terraform-sandbox/main.tf
```diff
@@ -0,0 +1,5 @@
+
+module "sandbox"  {
+  source = "./sandbox"
```
sdreyer, line 3: @simba-git I looked to see if there was an alternative way to deploy, rather than having them hardcoded in the same file. Terraform does have "environments" that will maintain a separate state for a variation of a deployment, but need to look into it more since I'm unsure if will react with VCS or being run on a different computer the way we want.

PR #243, opened 2022-04-26
add coordinator basic jobs for transformation and register source

Sami1309:


ff/serving/coordinator/coordinator_test.go
```diff
@@ -67,14 +77,15 @@ func TestCoordinatorCalls(t *testing.T) {
 	if err := testCoordinatorTrainingSet(); err != nil {
 		t.Fatalf("coordinator could not create training set: %v", err)
 	}
-	//g.Go(testCoordinatorCreateTransformation)
+	if err := testRegisterPrimaryTableFromSource(); err != nil {
```
sdreyer, line 80: At some point, give change these to be subtests with individual names instead of having them output as a single test

ff/serving/coordinator/coordinator.go
```diff
@@ -3,8 +3,11 @@ package coordinator
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"go.uber.org/zap"
+	//"github.com/alexkappa/mustache"
```
simba-git, line 9: remove comment

ff/serving/coordinator/coordinator.go
```diff
@@ -13,6 +16,20 @@ import (
 	"go.etcd.io/etcd/client/v3/concurrency"
 )
 
+func templateReplace(template string, replacements map[string]string) string {
+	formattedString := ""
+	numEscapes := strings.Count(template, "{{")
+	for i := 0; i < numEscapes; i++ {
+		split := strings.SplitN(template, "{{", 2)
+		afterSplit := strings.SplitN(split[1], "}}", 2)
+		keyFob := strings.TrimSpace(afterSplit[0])
```
simba-git, line 25: any reason to not name this key

ff/serving/coordinator/coordinator.go
```diff
@@ -90,19 +107,96 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+	providerResourceID := provider.ResourceID{Name: resID.Name, Variant: resID.Variant, Type: provider.Transformation}
+	transformationConfig := provider.TransformationConfig{TargetTableID: providerResourceID, Query: query}
+	if err := offlineStore.CreateTransformation(transformationConfig); err != nil {
+		return err
+	}
+	if err := c.Metadata.SetStatus(context.Background(), resID, metadata.READY); err != nil {
+		return err
+	}
 	return nil
+	//do we set this in metadata?
```
simba-git, line 153: ?

ff/serving/coordinator/coordinator.go
```diff
@@ -90,19 +107,96 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+
 }
 
-func (c *Coordinator) runTransformationJob(resID metadata.ResourceID) error {
-	//it's morphin time
+func (c *Coordinator) runPrimaryTableJob(transformSource *metadata.SourceVariant, resID metadata.ResourceID, offlineStore provider.OfflineStore) error {
+	providerResourceID := provider.ResourceID{Name: resID.Name, Variant: resID.Variant}
+	sourceName := transformSource.PrimaryDataSQLTableName()
+	if sourceName == "" {
+		return fmt.Errorf("creating blank source not implemented")
```
simba-git, line 162: Would we ever implement it? If not just say its an error and not that its unimplemented

ff/serving/coordinator/coordinator.go
```diff
@@ -90,19 +107,96 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+	}
+	if _, err := offlineStore.RegisterPrimaryFromSourceTable(providerResourceID, sourceName); err != nil {
+		fmt.Println(err)
```
simba-git, line 165: log or remove print

ff/serving/coordinator/coordinator.go
Sami1309: We will I imagine, creating an empty primary table with a predefined schema.

ff/serving/coordinator/coordinator.go
```diff
@@ -90,21 +112,120 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+		if source.Status() != metadata.READY {
+			return nil, fmt.Errorf("source in query not ready")
+		}
+		providerResourceID := provider.ResourceID{Name: source.Name(), Variant: source.Variant()}
+		if source.IsSQLTransformation() {
+			tableName = provider.GetTransformationName(providerResourceID)
+		} else if source.IsPrimaryDataSQLTable() {
+			tableName = provider.GetPrimaryTableName(providerResourceID)
+		}
+		sourceMap[fmt.Sprintf("%s.%s", nameVariant.Name, nameVariant.Variant)] = tableName
```
simba-git, line 132: make a struct called name variant, and make the map between that type and a string. If you do it this way, you risk ambiguity with a dot in it.

ff/serving/coordinator/coordinator.go
```diff
@@ -90,21 +112,120 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+func sanitize(ident string) string {
+	return db.Identifier{ident}.Sanitize()
+}
+
+func (c *Coordinator) runSQLTransformationJob(transformSource *metadata.SourceVariant, resID metadata.ResourceID, offlineStore provider.OfflineStore) error {
+	c.Logger.Debug("Running SQL transformation job on resource: ", resID)
+	templateString := transformSource.SQLTransformationQuery()
+	sources := transformSource.SQLTransformationSources()
+	allReady := false
+	for sourceVariants, err := c.Metadata.GetSourceVariants(context.Background(), sources); !allReady; sourceVariants, err = c.Metadata.GetSourceVariants(context.Background(), sources) {
```
simba-git, line 146: this should be a while loop:
for !allReady:
   sourcesVariant, err = ...

ff/serving/coordinator/coordinator.go
```diff
@@ -90,21 +112,120 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+		if err != nil {
+			return err
+		}
+		total := len(sourceVariants)
+		totalReady := 0
+		for _, sourceVariant := range sourceVariants {
+			if sourceVariant.Status() == string(metadata.READY) {
```
simba-git, line 153: why not make status return the same type as metadata.READY?

ff/serving/coordinator/coordinator.go
```diff
@@ -90,21 +112,120 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+				totalReady += 1
+			}
+			if sourceVariant.Status() == string(metadata.FAILED) {
+				return err
+			}
+		}
+		if total == totalReady {
```
simba-git, line 160: allReady = total == totalReady

ff/serving/coordinator/coordinator.go
simba-git: You don't have to cast to string here at all I think

ff/serving/coordinator/coordinator.go
```diff
@@ -90,21 +112,120 @@ func (c *Coordinator) WatchForNewJobs() error {
...
+	c.Logger.Debug("Running primary table job on resource: ", resID)
+	providerResourceID := provider.ResourceID{Name: resID.Name, Variant: resID.Variant}
+	sourceName := transformSource.PrimaryDataSQLTableName()
+	if sourceName == "" {
+		return fmt.Errorf("no source name set")
+	}
+	if _, err := offlineStore.RegisterPrimaryFromSourceTable(providerResourceID, sourceName); err != nil {
+		return err
+	}
+	if err := c.Metadata.SetStatus(context.Background(), resID, string(metadata.READY)); err != nil {
```
simba-git, line 193: let's avoid this string thing, we should just use metadata.READY's type

ff/serving/coordinator/coordinator.go
```diff
@@ -114,7 +235,7 @@ func (c *Coordinator) runFeatureMaterializeJob(resID metadata.ResourceID) error
 	if status == metadata.READY || status == metadata.FAILED {
 		return fmt.Errorf("feature already set to %s", metadata.ResourceStatus(status))
 	}
-	if err := c.Metadata.SetStatus(context.Background(), resID, metadata.PENDING); err != nil {
+	if err := c.Metadata.SetStatus(context.Background(), resID, string(metadata.PENDING)); err != nil {
```
simba-git, line 238: remove all these casts

ff/serving/coordinator/coordinator.go
```diff
@@ -352,23 +475,23 @@ func (c *Coordinator) executeJob(jobKey string) error {
 	switch job.Resource.Type {
```
simba-git, line 475: very repititve, you could probably do something like:

fns = map[resourceType]function(resource) error = {
   metadata.TRAINING_SET_VARIANT: c.runTrainingSetJob,
   ...
}
if err := fns[job.Resource.Type](job.Resource); err != nil {
			statusErr := c.Metadata.SetStatus(context.Background(), job.Resource, fmt.Sprintf("%v", err))
			return fmt.Errorf("training set job failed: %v", fmt.Sprintf("%v: %v", err, statusErr))
}

ff/serving/coordinator/coordinator.go
```diff
@@ -352,23 +475,23 @@ func (c *Coordinator) executeJob(jobKey string) error {
...
 	case metadata.TRAINING_SET_VARIANT:
 		if err := c.runTrainingSetJob(job.Resource); err != nil {
-			return fmt.Errorf("training set job failed: %v", err)
+			statusErr := c.Metadata.SetStatus(context.Background(), job.Resource, fmt.Sprintf("%v", err))
+			return fmt.Errorf("training set job failed: %v", fmt.Sprintf("%v: %v", err, statusErr))
```
simba-git, line 479: Errorf is already an Sprintf:
			return fmt.Errorf("training set job failed: %v: %v", err, statusErr)

ff/serving/metadata/client.go
```diff
@@ -74,9 +74,9 @@ type ResourceDef interface {
 	ResourceType() ResourceType
 }
 
-func (client *Client) SetStatus(ctx context.Context, resID ResourceID, status ResourceStatus) error {
+func (client *Client) SetStatus(ctx context.Context, resID ResourceID, status string) error {
```
simba-git, line 77: why is this a string

ff/serving/metadata/client.go
```diff
@@ -584,15 +584,15 @@ type SourceType interface {
 func (t TransformationSource) isSourceType() {}
 func (t PrimaryDataSource) isSourceType()    {}
 
-func (t SQLTransformationType) isTransformationType() {}
+func (t SQLTransformationType) IsTransformationType() {}
```
simba-git, line 587: should this return something

ff/serving/metadata/metadata.go
```diff
@@ -140,7 +140,7 @@ type Resource interface {
 	ID() ResourceID
 	Dependencies(ResourceLookup) (ResourceLookup, error)
 	Proto() proto.Message
-	UpdateStatus(ResourceStatus) error
```
simba-git, line 143: can we not make this a string, and could we make the types also exist in the proto files. Let's avoid using strings when we mean Enums.

simba-git (changes requested):

ff/serving/coordinator/coordinator.go
```diff
@@ -91,34 +114,30 @@ func NewCoordinator(meta *metadata.Client, logger *zap.SugaredLogger, cli *clien
...
 			for _, ev := range wresp.Events {
-				fmt.Println("FOUND:", ev)
 				if ev.Type == 0 {
-					fmt.Println("EXECUTING:", ev.Kv.Key)
-					go func() {
+					go func(ev *clientv3.Event) {
 						err := c.executeJob(string(ev.Kv.Key))
 						if err != nil {
-							fmt.Println(err)
+							c.Logger.Errorf("Error executing job: %v", err)
```
sdreyer, line 138: At whenever youre going through all the logging stuff, should come back to this and change it to be more structured. 
c.Logger.Debugw("Error executing job", "error", err)

sdreyer: Makes it easier to search for errors if we add the logs into a indexer

ff/serving/coordinator/coordinator.go
```diff
@@ -197,17 +204,17 @@ func (c *Coordinator) runSQLTransformationJob(transformSource *metadata.SourceVa
 	}
 	providerResourceID := provider.ResourceID{Name: resID.Name, Variant: resID.Variant, Type: provider.Transformation}
 	transformationConfig := provider.TransformationConfig{TargetTableID: providerResourceID, Query: query}
-	c.Logger.Debug(fmt.Sprintf("%v", transformationConfig))
 	if err := offlineStore.CreateTransformation(transformationConfig); err != nil {
 		return err
 	}
-	if err := c.Metadata.SetStatus(context.Background(), resID, string(metadata.READY)); err != nil {
+	if err := c.Metadata.SetStatus(context.Background(), resID, pb.ResourceStatus{Status: pb.ResourceStatus_READY}); err != nil {
```
sdreyer, line 210: Does it make more sense to do:
SetStatus(ctx, redID, pb.ResourceStatus_READY)
and have SetStatus nest the message? Feel like it would look cleaner and be easier to interact with

ff/serving/coordinator/coordinator.go
```diff
@@ -254,56 +258,46 @@ func (c *Coordinator) runRegisterSourceJob(resID metadata.ResourceID) error {
...
-		return fmt.Errorf("feature already set to %s", metadata.ResourceStatus(status))
+		return fmt.Errorf("feature already set to %s", status.String())
 	}
-	if err := c.Metadata.SetStatus(context.Background(), resID, string(metadata.PENDING)); err != nil {
+	if err := c.Metadata.SetStatus(context.Background(), resID, pb.ResourceStatus{Status: pb.ResourceStatus_PENDING}); err != nil {
 		return err
 	}
 	sourceNameVariant := feature.Source()
-	source, err := c.Metadata.GetSourceVariant(context.Background(), sourceNameVariant)
+	source, err := c.AwaitPendingSource(sourceNameVariant, 30) //second argument is seconds to wait for source to become ready
```
sdreyer, line 275: Could add this timeout value to the coordinator config so it can be changed further up

Sami1309: sure, I was just unsure where to put it as there are multiple timeout levels that may not be obvious from a top down perspective what each one corresponds to. Though a global timeout value with optional overrides would make sense

ff/serving/coordinator/coordinator.go
Sami1309: yes it would, I'll change it

ff/serving/coordinator/coordinator.go
Sami1309: yes, I'll change it and for each level in the debugging heiarchy

ff/serving/coordinator/coordinator.go
sdreyer: Yeah that would work

ff/serving/metadata/proto/metadata.proto
```diff
@@ -54,10 +54,37 @@ message Name {
...
+        CREATED = 1;
+        PENDING = 2;
+        READY = 3;
+        FAILED = 4;
+      }
+    Status status = 1;
+    string error_message = 2;
+}
+
+enum MetadataResourceType {
```
simba-git, line 69: Why isn't this just ResourceType

ff/serving/coordinator/coordinator.go
```diff
@@ -43,6 +44,28 @@ type Coordinator struct {
 	Spawner    JobSpawner
 }
 
+func (c *Coordinator) AwaitPendingSource(sourceNameVariant metadata.NameVariant, timeout int32) (*metadata.SourceVariant, error) {
```
simba-git, line 47: why not make the second variable a time.Duration so you don't have to add a comment making that clear?

ff/serving/coordinator/coordinator.go
```diff
@@ -91,34 +114,30 @@ func NewCoordinator(meta *metadata.Client, logger *zap.SugaredLogger, cli *clien
 const MAX_ATTEMPTS = 1
 
 func (c *Coordinator) WatchForNewJobs() error {
+	c.Logger.Debug("Watching for new jobs")
 	getResp, err := (*c.KVClient).Get(context.Background(), "JOB_", clientv3.WithPrefix())
 	if err != nil {
 		return err
 	}
 	for _, kv := range getResp.Kvs {
-		fmt.Println("FOUND:", kv)
```
simba-git, line 99: You may want to keep something like this as a debug log

ff/serving/metadata/proto/metadata.proto
Sami1309: why come

ff/serving/coordinator/coordinator.go
```diff
@@ -85,40 +108,37 @@ func NewCoordinator(meta *metadata.Client, logger *zap.SugaredLogger, cli *clien
...
 			for _, ev := range wresp.Events {
-				fmt.Println("FOUND:", ev)
 				if ev.Type == 0 {
-					fmt.Println("EXECUTING:", ev.Kv.Key)
-					go func() {
+					go func(ev *clientv3.Event) {
 						err := c.executeJob(string(ev.Kv.Key))
 						if err != nil {
-							fmt.Println(err)
+							c.Logger.Debugw("Error executing job", "error", err)
```
sdreyer, line 139: Change these so we can tell them apart if an error occurs. 
Can be something like "Error executing job: Initial Search"/ "Error executing job: Polling Search"

ff/serving/coordinator/coordinator.go
```diff
@@ -163,12 +174,12 @@ func sanitize(ident string) string {
 }
 
 func (c *Coordinator) runSQLTransformationJob(transformSource *metadata.SourceVariant, resID metadata.ResourceID, offlineStore provider.OfflineStore) error {
+	c.Logger.Debug("Running SQL transformation job on resource: ", resID)
```
sdreyer, line 177: I think this could be Info

ff/serving/coordinator/coordinator.go
```diff
@@ -197,17 +205,17 @@ func (c *Coordinator) runSQLTransformationJob(transformSource *metadata.SourceVa
...
 	}
-	if err := c.Metadata.SetStatus(context.Background(), resID, string(metadata.READY)); err != nil {
+	if err := c.Metadata.SetStatus(context.Background(), resID, metadata.READY, ""); err != nil {
 		return err
 	}
 	return nil
 }
 
 func (c *Coordinator) runPrimaryTableJob(transformSource *metadata.SourceVariant, resID metadata.ResourceID, offlineStore provider.OfflineStore) error {
+	c.Logger.Debug("Running primary table job on resource: ", resID)
```
sdreyer, line 218: Set as Info

ff/serving/coordinator/coordinator.go
```diff
@@ -216,31 +224,28 @@ func (c *Coordinator) runPrimaryTableJob(transformSource *metadata.SourceVariant
...
 	}
-	if err := c.Metadata.SetStatus(context.Background(), resID, string(metadata.READY)); err != nil {
+	if err := c.Metadata.SetStatus(context.Background(), resID, metadata.READY, ""); err != nil {
 		return err
 	}
 	return nil
 }
 
 func (c *Coordinator) runRegisterSourceJob(resID metadata.ResourceID) error {
+	c.Logger.Debug("Running register source job on resource: ", resID)
```
sdreyer, line 234: Set as info, same as all the other job starts/ends
Essentially all the jobs initiations that start running should be info since they are initiated by either user action or a schedule. If we set this to production mode and log level is info it would be nice to see when jobs start and end

sdreyer (approved):
I think I you made all the previous changes I had asked. 

Main thing here is that we should have a bit more verbosity for the INFO logging level. I dont want to have to set the threshold to Debug to see if things like jobs were run

simba-git (approved):

PR #246, opened 2022-04-27
Dashboard valuable fixes

Sami1309:


ff/serving/metadata/etcd.go
```diff
@@ -129,6 +126,9 @@ func (s EtcdStorage) Get(key string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	if len(resp.Kvs) == 0 {
```
simba-git, line 129: Why did we change this?

Sami1309: It's the cause of the 500 error in the case of no data on the backend. Given that the etcd client returns an empty struct and not an error in the event of a get request with no data, it seemed ok to me to return an empty return rather than an error if the get returns nothing.

sdreyer (approved):

PR #247, opened 2022-05-11
Feature/helm charts/sterling

sdreyer:
This is a working version of the sandbox. There are still some things to do though:
- Removed hardcoded endpoints. Currently some services have the endpoints they connect to hardcoded instead of using an env var. Especially in the dashboard
- Optimize Docker images
- Change tls from self-signed to an issuer
- Check that prometheus and serving are working properly
- Removed extraneous commets/commented code


ff/serving/client/metadata_pb2_grpc.py
```diff
@@ -0,0 +1,1249 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
sdreyer, line 1: Remove this file

ff/serving/client/metadata_pb2.py
```diff
@@ -0,0 +1,241 @@
+# -*- coding: utf-8 -*-
```
sdreyer, line 1: Remove this file

ff/serving/coordinator/e2e_test.go
```diff
@@ -0,0 +1,810 @@
...
+	}
+	fmt.Printf("%#v\n", defs)
+	if err := client.CreateAll(context.Background(), defs); err != nil {
+		fmt.Println("Could not create all", err)
+		return err
+	}
+	return nil
+}
+
+//
```
sdreyer, line 135: Remove commented code and maybe entire file

PR #249, opened 2022-05-17
Working SQL Base Class

sdreyer:
TODO: 
- Give functions better names
- Would like to separate the functions into different interfaces for clarity


ff/serving/provider/sql.go
```diff
@@ -0,0 +1,1225 @@
...
+}
+
+func (store *sqlOfflineStore) getResourceTableName(id ResourceID) string {
+	var idType string
+	if id.Type == Feature {
+		idType = "feature"
+	} else {
+		idType = "label"
+	}
+	return fmt.Sprintf("featureform_resource_%s_%s_%s", idType, id.Name, id.Variant)
```
simba-git, line 92: what happens if name or variant has an underscore then:
name = abc_def
variant = ghi

is the same as
name = abc
variant = def_ghi

ff/serving/provider/sql.go
```diff
@@ -0,0 +1,1225 @@
...
+	table, err := store.newsqlOfflineTable(store.db, tableName, valueType)
+	if err != nil {
+		return nil, err
+	}
+	return table, nil
+}
+
+// getValueIndex returns the index of the value column in the schema.
+// Returns -1 if an entity column is not found
+func (store *sqlOfflineStore) getValueIndex(columns []TableColumn) int {
```
simba-git, line 345: nit: it's more "go-like" for functions like this to return an (int, error). Probably not worth changing here.

ff/serving/provider/sql.go
```diff
@@ -0,0 +1,1225 @@
...
+	}
+}
+
+func (store *sqlOfflineStore) newsqlOfflineTable(db *sql.DB, name string, valueType ValueType) (*sqlOfflineTable, error) {
+	columnType, err := determineColumnType(valueType)
+	if err != nil {
+		return nil, err
+	}
+	tableCreateQry := store.query.newSQLOfflineTable(name, columnType)
+	//tableCreateQry := fmt.Sprintf("CREATE TABLE %s (entity VARCHAR, value %s, ts TIMESTAMP_NTZ, UNIQUE (entity, ts))", sanitize(name), columnType)
```
simba-git, line 835: leftover coment

ff/serving/provider/sql.go
```diff
@@ -0,0 +1,1225 @@
...
+	}
+	return true, nil
+}
+
+func (store *sqlOfflineStore) CreateTransformation(config TransformationConfig) error {
+	name, err := store.createTransformationName(config.TargetTableID)
+	if err != nil {
+		return err
+	}
+	// Only allow transformations to be run with SELECT queries
```
simba-git, line 898: leftover?

ff/serving/provider/sql.go
```diff
@@ -0,0 +1,1225 @@
...
+	if _, err := store.db.Exec(query); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (store *sqlOfflineStore) createTransformationName(id ResourceID) (string, error) {
+	switch id.Type {
+	case Label, Feature:
```
simba-git, line 910: should this be possible?

ff/serving/provider/sql.go
```diff
@@ -0,0 +1,1225 @@
...
+func (it *sqlGenericTableIterator) Err() error {
+	return it.err
+}
+
+type defaultOfflineSQLQueries struct {
+	BindingStyle variableBindingStyle
+}
+
+func (q *defaultOfflineSQLQueries) newVariableBindingIterator() VariableBindingIterator {
+	//fmt.Println(q.BindingStyle)
```
simba-git, line 998: leftover

ff/serving/provider/sql.go
sdreyer: Yeah

ff/serving/provider/sql.go
sdreyer: Good point

simba-git (approved):

PR #253, opened 2022-05-20
Add serving to python package

sdreyer:
Python client/api server with serving included. 

Currently just includes a test to check if the python client can successfully connect to the api server, where it will fail connecting to metadata and serving. Will build out a better test suite in the E2E task


ff/serving/api/api.go
```diff
@@ -109,23 +149,29 @@ func (serv *ApiServer) Serve() error {
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 	}
-	conn, err := grpc.Dial(serv.metaAddr, opts...)
+	metaConn, err := grpc.Dial(serv.metadata.address, opts...)
```
Sami1309, line 152: could these be done in paralell?

ff/serving/api/api.go
```diff
@@ -196,7 +242,7 @@ func main() {
 			panic(fmt.Sprintf("health check HTTP server failed: %+v", err))
 		}
 	}()
-	serv, err := NewApiServer(logger, "0.0.0.0:7878", "sandbox-metadata-server:8080")
+	serv, err := NewApiServer(logger, "0.0.0.0:7878", "sandbox-metadata-server:8080", "sandbox-serving-server:8080")
```
Sami1309, line 245: should port be derived from environmental variable in helm chart, or more globally to be seen higher? Just because 8080 is typed in plaintext in seperate places

ff/serving/client/examples/cli.py
```diff
@@ -0,0 +1,55 @@
...
+snowflake = ff.register_snowflake(
+    name="snowflake",
+    username="username",
+    password="password",
+    account="account",
+    organization="organization",
+    database="database",
+    schema="schema",
+    description="Snowflake",
+    team="Featureform Success Team",
```
Sami1309, line 16: no team, organization, atm.

ff/serving/api/api.go
sdreyer: Does it need to be? Its just opening a channel right?

ff/serving/api/api.go
sdreyer: Was leaving this as part of the helm task to remove hardcoded addresses so I can do them all at once. There are other things in the chart that all need to change together

ff/serving/client/examples/cli.py
sdreyer: Created an asana task to add these into the snowflake config

ff/serving/api/api.go
Sami1309: ok

ff/serving/client/examples/cli.py
Sami1309: ok, but should they also be in metadata for the providers?

ff/serving/api/api.go
Sami1309: it doesn't just saves time i imagine, nbd tho

Sami1309 (dismissed):
lgtm

ff/serving/client/examples/cli.py
sdreyer: Actually org is already being used. Its mainly just needed for auth. We can add Team in later. 

Org is just a generated string, nothing we really need for anything else

ff/serving/api/api.go
sdreyer: Gonna leave it for now since it only runs at startup and dont want to mess with it too much atm

PR #255, opened 2022-05-25
Enable Updates For Offline Store

sdreyer:


Sami1309 (commented):
very good, just minor things but probably because I don't understand the full way the whole goes together

ff/serving/provider/postgres.go
```diff
@@ -103,11 +103,20 @@ func (q postgresSQLQueries) primaryTableRegister(tableName string, sourceName st
...
-			"AS rn FROM %s) t WHERE rn=1)", sanitize(tableName), sanitize(resultName))
+			"AS rn FROM %s) t WHERE rn=1);  CREATE UNIQUE INDEX ON %s (entity);", sanitize(tableName), sanitize(sourceName), sanitize(tableName))
+}
+
+func (q postgresSQLQueries) materializationUpdate(db *sql.DB, tableName string, sourceName string) error {
+	_, err := db.Exec(fmt.Sprintf("REFRESH MATERIALIZED VIEW CONCURRENTLY %s", sanitize(tableName)))
+	return err
+}
+
+func (q postgresSQLQueries) materializationExists() string {
```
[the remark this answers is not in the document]
sdreyer: All of the queries are inside functions because it forces the Offline SQL interface to be implemented and provides default values as well

ff/serving/provider/sql.go
```diff
@@ -35,10 +37,13 @@ type OfflineTableQueries interface {
 	registerResources(db *sql.DB, tableName string, schema ResourceSchema, timestamp bool) error
 	primaryTableRegister(tableName string, sourceName string) string
 	primaryTableCreate(name string, columnString string) string
-	getColumnNames() string
+	getColumnNames(db *sql.DB, tableName string) (*sql.Rows, error)
```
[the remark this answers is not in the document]
sdreyer: Its because postgres and snowflake execute this query differently so the database pointer is passed through instead of the function just returning a string

ff/serving/provider/postgres.go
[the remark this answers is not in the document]
sdreyer: It also allows a binding function to be added to create more generic queries as seen later on with bind.Next()

ff/serving/provider/sql.go
```diff
@@ -322,13 +329,20 @@ func (store *sqlOfflineStore) GetTransformationTable(id ResourceID) (Transformat
...
+	rows, err := store.db.Query(existsQuery, name)
+	defer rows.Close()
+	if err != nil {
 		return nil, err
-	} else if !exists {
-		return nil, &TableNotFound{id.Name, id.Variant}
 	}
-	columnQuery := store.query.getColumnNames()
-	rows, err := store.db.Query(columnQuery, name)
+	rowCount := 0
```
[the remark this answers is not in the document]
sdreyer: Will change this to an If

ff/serving/provider/sql.go
```diff
@@ -542,14 +556,47 @@ func (store *sqlOfflineStore) GetMaterialization(id MaterializationID) (Material
...
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := store.db.Query(getMatQry, tableName)
+	defer rows.Close()
+	if err != nil {
+		return nil, err
+	}
+	rowCount := 0
```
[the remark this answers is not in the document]
sdreyer: Changed to if !rows.next() instead

ff/serving/provider/sql.go
```diff
@@ -1108,25 +1182,57 @@ func (q defaultOfflineSQLQueries) registerResources(db *sql.DB, tableName string
 func (q defaultOfflineSQLQueries) primaryTableRegister(tableName string, sourceName string) string {
 	return fmt.Sprintf("CREATE TABLE %s AS SELECT * FROM TABLE('%s')", sanitize(tableName), sanitize(sourceName))
 }
-func (q defaultOfflineSQLQueries) getColumnNames() string {
+func (q defaultOfflineSQLQueries) getColumnNames(db *sql.DB, name string) (*sql.Rows, error) {
```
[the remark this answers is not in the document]
sdreyer: Yeah the query returns the information table for column names of a table which returns them as rows

ff/serving/provider/postgres.go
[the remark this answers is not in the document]
Sami1309: ok 👍

ff/serving/provider/sql.go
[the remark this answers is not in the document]
Sami1309: Function name is confusing. Does it return an iterator on rows that represent column names?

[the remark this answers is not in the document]
Sami1309: Saw your other comment and I understand. Does sql.Rows have a length function?

ff/serving/provider/postgres.go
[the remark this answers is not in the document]
simba-git: You can just return a string though, no need for the Sprintf, you're not formatting anything

ff/serving/provider/postgres.go
```diff
@@ -142,6 +151,14 @@ func (q postgresSQLQueries) createValuePlaceholderString(columns []TableColumn)
...
 
 func (q postgresSQLQueries) trainingSetCreate(store *sqlOfflineStore, def TrainingSetDef, tableName string, labelName string) error {
+	return q.trainingSetGeneral(store, def, tableName, labelName, false)
+}
+
+func (q postgresSQLQueries) trainingSetUpdate(store *sqlOfflineStore, def TrainingSetDef, tableName string, labelName string) error {
+	return q.trainingSetGeneral(store, def, tableName, labelName, true)
+}
+
+func (q postgresSQLQueries) trainingSetGeneral(store *sqlOfflineStore, def TrainingSetDef, tableName string, labelName string, isUpdate bool) error {
```
simba-git, line 161: Can we get a better name for this, trainingSetQuery?

ff/serving/provider/postgres.go
```diff
@@ -212,5 +245,28 @@ func (q postgresSQLQueries) numRows(n interface{}) (int64, error) {
...
 func (q postgresSQLQueries) transformationCreate(name string, query string) string {
-	return fmt.Sprintf("CREATE TABLE %s AS %s ", sanitize(name), query)
+	return fmt.Sprintf("CREATE TABLE  %s AS %s", sanitize(name), query)
+}
+
+func (q postgresSQLQueries) transformationUpdate(db *sql.DB, tableName string, query string) error {
+	santizedName := sanitize(tableName)
+	tempName := sanitize(fmt.Sprintf("tmp_%s", tableName))
+	oldName := sanitize(fmt.Sprintf("old_%s", tableName))
+	transaction := fmt.Sprintf("BEGIN;"+
```
simba-git, line 255: This looks really similar to the transaction above, can we generalize?

ff/serving/provider/postgres.go
```diff
@@ -212,5 +245,28 @@ func (q postgresSQLQueries) numRows(n interface{}) (int64, error) {
...
+		"ALTER TABLE %s RENAME TO %s;"+
+		"ALTER TABLE %s RENAME TO %s;"+
+		"DROP TABLE %s;"+
+		"COMMIT;", tempName, query, santizedName, oldName, tempName, santizedName, oldName)
+	_, err := db.Exec(transaction)
+	return err
+}
+
+func (q postgresSQLQueries) transformationExists() string {
+	return fmt.Sprintf("SELECT * FROM pg_matviews WHERE matviewname = $1")
```
simba-git, line 266: no need for the Sprintf

ff/serving/provider/sql.go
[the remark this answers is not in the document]
simba-git: can't you return an []string with all the queries instead? I still don't undersatnd why you have to pass in the db

simba-git (dismissed):
Few minor things and a couple questions

ff/serving/provider/postgres.go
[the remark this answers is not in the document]
sdreyer: Right, missed that

ff/serving/provider/postgres.go
sdreyer: Changed

ff/serving/provider/postgres.go
sdreyer: Changed

ff/serving/provider/sql.go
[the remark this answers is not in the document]
sdreyer: Will have it return strings instead. Its because snowflake needs to bind the table name and postgres doesnt so they need specific implementations

ff/serving/provider/postgres.go
sdreyer: you got it

sdreyer: Done

ff/serving/provider/offline.go
```diff
@@ -117,14 +117,17 @@ type OfflineStore interface {
...
 	CreateTransformation(config TransformationConfig) error
 	GetTransformationTable(id ResourceID) (TransformationTable, error)
+	UpdateTransformation(config TransformationConfig) error
 	CreatePrimaryTable(id ResourceID, schema TableSchema) (PrimaryTable, error)
 	GetPrimaryTable(id ResourceID) (PrimaryTable, error)
 	CreateResourceTable(id ResourceID, schema TableSchema) (OfflineTable, error)
 	GetResourceTable(id ResourceID) (OfflineTable, error)
 	CreateMaterialization(id ResourceID) (Materialization, error)
 	GetMaterialization(id MaterializationID) (Materialization, error)
+	UpdateMaterialization(id ResourceID) (Materialization, error)
```
Sami1309, line 127: Why update materializations? Materializations are meant to be created ad-hoc and then deleted throughout the online store process.

Sami1309: materializations change, no deletion now

Sami1309: materializations change, no deletion now

ff/serving/provider/postgres.go
```diff
@@ -212,5 +238,46 @@ func (q postgresSQLQueries) numRows(n interface{}) (int64, error) {
...
+func (q postgresSQLQueries) atomicUpdate(db *sql.DB, tableName string, tempName string, query string) error {
+	santizedName := sanitize(tableName)
+	oldName := sanitize(fmt.Sprintf("old_%s", tableName))
+	transaction := fmt.Sprintf("BEGIN;"+
+		"%s;"+
+		"ALTER TABLE %s RENAME TO %s;"+
+		"ALTER TABLE %s RENAME TO %s;"+
+		"DROP TABLE %s;"+
+		"COMMIT;", query, santizedName, oldName, tempName, santizedName, oldName)
+	_, err := db.Exec(transaction)
```
Sami1309, line 259: If this fails, does the temp table persist in the database? Is there a way to delete and restart before the transaction?

ff/serving/provider/snowflake.go
```diff
@@ -68,3 +68,7 @@ func snowflakeOfflineStoreFactory(config SerializedConfig) (Provider, error) {
 	}
 	return store, nil
 }
+
+func (q snowflakeSQLQueries) materializationDrop(tableName string) string {
```
Sami1309, line 72: is this syntax unique to snowflake?

Sami1309, line 72: is this syntax unique to snowflake?

ff/serving/provider/postgres.go
sdreyer: All actions are rollback automatically if one step fails, so it will just leave it will the original table and no temp table

ff/serving/provider/snowflake.go
sdreyer: I've left the default as dropping a materialized view. Snowflake just uses a table in this case so I made it non-default

sdreyer: dupe?

Sami1309 (dismissed):

PR #258, opened 2022-05-27
Table copy switched to view

sdreyer:


simba-git (approved):
lgtm

ff/serving/provider/sql.go
```diff
@@ -145,12 +144,19 @@ func (store *sqlOfflineStore) tableExists(id ResourceID) (bool, error) {
 	}
 	query := store.query.tableExists()
 	err = store.db.QueryRow(query, tableName).Scan(&n)
-	if n == 0 {
-		return false, nil
+	if n > 0 && err == nil {
+		return true, nil
 	} else if err != nil {
```
Sami1309, line 149: no need for an else if since the previous is a return

ff/serving/provider/sql.go
```diff
@@ -145,12 +144,19 @@ func (store *sqlOfflineStore) tableExists(id ResourceID) (bool, error) {
...
 		return false, err
 	}
-	return true, nil
+	query = store.query.viewExists()
+	err = store.db.QueryRow(query, tableName).Scan(&n)
+	if n > 0 && err == nil {
+		return true, nil
+	} else if err != nil {
+		return false, err
```
Sami1309, line 157: Is there no distinction programatically between a table existing but no view and both not existing? Could this prompt the creation of a view if the first is not an error but no rows return and the next is false?

sdreyer: This checks to see if there is an existing table with the name. If no results are returned and there is no error, it then checks to see if there is an existing view with that name. If no results are returned and theres no error, then it returns false. 

Tables and views can be interchageable by the rest of the offline store so only one needs to exist

ff/serving/provider/sql.go
sdreyer: yeah could change that to two IFs, but felt like theres more organization with if/elseif

ff/serving/provider/sql.go
Sami1309: Ok sounds good.

ff/serving/provider/sql.go
Sami1309: ok just a style thing no biggie

Sami1309 (approved):
lgtm

PR #259, opened 2022-05-28
Add name and variant checking to resources in metadata

Sami1309:


ff/serving/metadata/metadata.go
```diff
@@ -112,6 +114,38 @@ func (id ResourceID) Parent() (ResourceID, bool) {
...
+func resourceNamedSafely(id ResourceID) error {
+	for _, substr := range bannedStrings {
+		if len(id.Name) >= len(substr) && strings.Contains(id.Name, substr) {
+			return fmt.Errorf("resource name %s contains banned string %s", id.Name, substr)
+		}
+		if len(id.Variant) >= len(substr) && strings.Contains(id.Variant, substr) {
+			return fmt.Errorf("resource variant %s contains banned string %s", id.Name, substr)
+		}
+	}
+	for _, substr := range bannedPrefixes {
```
sdreyer, line 130: I believe there a strings.suffix and strings.prefix function

Sami1309: good call

sdreyer (approved):

PR #267, opened 2022-06-03
Helm chart fixes

sdreyer:


charts/featureform/charts/api-server/values.yaml
```diff
@@ -43,7 +43,7 @@ service:
   port: 7878
 
 apiserver:
-  port: 7878
+  port: "7878"
```
Sami1309, line 46: why r there quotes here and not around the other 7878

sdreyer: leftover from how I was doing it before. Will fix

sdreyer: Fixed

Sami1309 (approved):
lgtm

PR #270, opened 2022-06-03
Feature/dynamodb support/riddhi

RiddhiBagadiaa:


ff/serving/provider/dynamodb.go
```diff
@@ -0,0 +1,249 @@
...
+		ProviderConfig: options.Serialized(),
+	},
+	}
+}
+
+func (store *dynamodbOnlineStore) AsOnlineStore() (OnlineStore, error) {
+	return store, nil
+}
+
+func CreateMetadataTable(dynamodbClient *dynamodb.DynamoDB) {
```
sdreyer, line 75: Whats this metadatatable for? Is it storing the names of the feature tables?

ff/serving/provider/provider.go
```diff
@@ -51,6 +58,23 @@ func (r *RedisConfig) Deserialize(config SerializedConfig) error {
 	return nil
 }
 
+// added by Riddhi
```
sdreyer, line 61: Remove

ff/serving/provider/online_test.go
```diff
@@ -56,6 +61,8 @@ func TestOnlineStores(t *testing.T) {
 		{LocalOnline, []byte{}, false},
 		{RedisOnline, redisMockConfig.Serialized(), false},
 		{RedisOnline, redisLiveConfig.Serialized(), true},
+		{DynamoDBOnline, dynamoConfig.Serialized(), false},
```
sdreyer, line 64: Only need true or false. Redis has both because we have a mock redis cluster and a live redis cluster we test on

ff/serving/provider/dynamodb.go
RiddhiBagadiaa: The names of the feature tables and the value type associated with them. Primarily for the GetTable() function

ff/serving/provider/dynamodb.go
```diff
@@ -0,0 +1,249 @@
+package provider
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
```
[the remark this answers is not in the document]
RiddhiBagadiaa: Pushed the go.mod and go.sum files

.github/workflows/testing.yml
```diff
@@ -113,6 +113,13 @@ jobs:
         ports:
           - 6379:6379
 
+      dynamodb:
+        image: dynamodb
```
sdreyer, line 117: Change image to this: amazon/dynamodb-local 

dynamodb is invalid

ff/serving/provider/online_test.go
```diff
@@ -48,6 +48,11 @@ func TestOnlineStores(t *testing.T) {
 	redisLiveConfig := &RedisConfig{
 		Addr: liveAddr,
 	}
+
+	dynamoConfig := &DynamodbConfig{
+		Addr:   "http://localhost:4566",
```
sdreyer, line 53: Make this environment variable

.github/workflows/testing.yml
```diff
@@ -5,6 +5,7 @@ env:
   TYPESENSE_PORT: 8108
   TYPESENSE_API_KEY: "xyz"
   REDIS_PORT: 6379
+  DYNAMO_PORT: 8000
```
sdreyer, line 8: Delete this, unused

.github/workflows/testing.yml
```diff
@@ -125,6 +126,12 @@ jobs:
         # supported for use outside of 'steps'
         ports:
           - 9042:9042
+      dynamodb:
+        image: amazon/dynamodb-local
```
sdreyer, line 130: This is container is unused

client/src/featureform/register_test.py
```diff
@@ -78,14 +78,43 @@ def test_register_redis(registrar, args):
...
 
+@pytest.mark.parametrize("args", [
+    {
+        "name": "dynamodb"
+    },
+    {
+        "name": "dynamodb",
+        "description": "test",
+        "team": "featureform",
+        "host": "localhost",
```
sdreyer, line 96: remove host/port

client/src/featureform/resources_test.py
```diff
@@ -38,6 +38,14 @@ def redis_config():
         db=3,
     )
 
+@pytest.fixture
+def dynamodb_config():
+    return DynamodbConfig(
+        host="localhost",
```
sdreyer, line 44: remove host/port, add region

sdreyer (approved):

client/src/featureform/register_test.py
```diff
@@ -78,14 +78,43 @@ def test_register_redis(registrar, args):
...
+    },
+])
+def test_register_dynamodb(registrar, args):
+    registrar.register_dynamodb(**args)
+
+
+minimal_user_args = {
+    "name": "user",
+    "access_key": "abc",
+    "secret_key": "abc"
```
sdreyer, line 109: add region field

provider/dynamodb.go
```diff
@@ -0,0 +1,295 @@
...
+	config := &aws.Config{
+		Region:      aws.String(options.Region),
+		Credentials: credentials.NewStaticCredentials(options.AccessKey, options.SecretKey, ""),
+	}
+	sess := session.Must(session.NewSession(config))
+	dynamodbClient := dynamodb.New(sess)
+	if err := CreateMetadataTable(dynamodbClient); err != nil {
+		return nil, err
+	}
+	return &dynamodbOnlineStore{dynamodbClient, options.Prefix, BaseProvider{
```
sdreyer, line 69: Add a field called timeout that we can set here. Use it in the table creation wait loop so it returns an error if it waits for too long

.github/workflows/testing.yml
steffitan23: should i delete or should i change it to something?

sdreyer: just remove this block. Were testing it on aws instead of the container rn

provider/dynamodb.go
steffitan23: do we want users to be able to set this?

sdreyer: We can just do it internally for now. Something like 5 minutes should be more than enough. Just dont want it retrying forever

provider/dynamodb.go
```diff
@@ -0,0 +1,301 @@
...
+		ProviderType:   DynamoDBOnline,
+		ProviderConfig: options.Serialized(),
+	}, 10,
```
sdreyer, line 73: This needs to be increased. Its only 10 seconds right now. Should be more like 2 minutes

sdreyer (approved):

PR #276, opened 2022-06-07
Seperate offline and online feature registration

Sami1309:
If a feature is registered without a provider, it defaults to registering it in the offline store. With an online provider (only online is valid and makes sense) it will register offline if not registered so far, and then registers it online.

Also for some reason the pb file was commited, then i deleted it, so it interpets it as a big delete when it's generated with testing anyway

### DIFF
--- a/client/src/featureform/proto/metadata.proto
+++ b/client/src/featureform/proto/metadata.proto
@@ -86,6 +86,7 @@ message ResourceStatus {
         PENDING = 2;
         READY = 3;
         FAILED = 4;
+        READY_ONLINE = 5;
       }
     Status status = 1;
     string error_message = 2;
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -467,6 +467,83 @@ func (c *Coordinator) runLabelRegisterJob(resID metadata.ResourceID, schedule st
 	return nil
 }
 
+func (c *Coordinator) runFeatureRegisterOffline(resID metadata.ResourceID) error {
+	feature, err := c.Metadata.GetFeatureVariant(context.Background(), metadata.NameVariant{resID.Name, resID.Variant})
+	if err != nil {
+		return fmt.Errorf("get feature variant from metadata: %w", err)
+	}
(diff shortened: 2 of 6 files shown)


client/src/featureform/proto/metadata.proto
```diff
@@ -69,6 +69,7 @@ message ResourceStatus {
         PENDING = 2;
         READY = 3;
         FAILED = 4;
+        READY_ONLINE = 5;
```
sdreyer, line 72: Do we need another status for if offline is ready?

Currently if offline store succeeds, but online store fails, then it counts the whole job as failed. However the training set will still run.

coordinator/coordinator.go
```diff
@@ -356,7 +356,7 @@ func (c *Coordinator) runSQLTransformationJob(transformSource *metadata.SourceVa
 	return nil
 }
 
-func (c *Coordinator) runPrimaryTableJob(transformSource *metadata.SourceVariant, resID metadata.ResourceID, offlineStore provider.OfflineStore, schedule string) error {
```
sdreyer, line 359: Is this schedule string field supposed to be removed in all these?

client/src/featureform/proto/metadata.proto
sdreyer: Maybe have a FAILED_ONLINE status so we know that the offline table was created so it doesnt linger

sdreyer:
Its working. I think we just need a clearer outcome if the offline table succeeds and the online table fails

client/src/featureform/proto/metadata.proto
Sami1309: honestly it makes sense to have status for online and offline separately since the logic is so bifurcated

coordinator/coordinator.go
Sami1309: yes, since primary table jobs never need a schedule since they're relational

sdreyer:
Will rebase and test this later

PR #277, opened 2022-06-07
Add env host pull and error if not set

Sami1309:


client/src/featureform/cli.py
```diff
@@ -78,6 +79,11 @@ def cli():
 def apply(host, cert, insecure, files):
     """apply changes to featureform
     """
+    if host == "":
+        envHost = os.getenv('HOST')
```
sdreyer, line 83: Think FEATUREFORM_HOST would be better to avoid any conflicts

client/src/featureform/cli.py
```diff
@@ -78,6 +79,11 @@ def cli():
...
+        if envHost == None:
+            raise ValueError("Host value must be set")
```
sdreyer, line 85: Be more specific. say to either use --host or set env variable

sdreyer (approved):

PR #281, opened 2022-06-08
Quick start

sdreyer:


client/examples/quickstart.py
```diff
@@ -0,0 +1,77 @@
+# Run in command line with:
+# python3 -m featureform examples/cli.py -host <hostname:port>
```
Sami1309, line 2: should be python3 -m featureform apply examples/quickstart.py --host <hostname:port>

Sami1309 (dismissed):
good just change the script comment

PR #287, opened 2022-06-21
Other/e2e testing/sterling

sdreyer:
Supports basic E2E testing using the quickstart materials. 
Setups up a minikube instance, installs featureform and quickstart helm charts, then runs a python test to verify the correct resources were created


client/tests/e2e.py
```diff
@@ -0,0 +1,181 @@
...
+        if all_ready:
+            break
+        req = requests.get("http://localhost:7000/data/training-sets", json=True)
+        json_ret = req.json()
+        if retries > 20:
+            assert (json_ret == training_sets)
+        all_ready = True
+        for res in json_ret:
+            for v in res['variants']:
+                del res['variants'][v]['created']
```
ahmadnazeri, line 174: Is there a reason for this delete?

sdreyer: Yeah everytime a resource is created it gets a timestamp. Theres not a great way to test the exact time the resource was created, since it depends on a lot of external sources, so I remove it instead

client/tests/e2e.py
```diff
@@ -0,0 +1,181 @@
...
+                if res['variants'][v]['status'] != 'READY':
+                    time.sleep(5)
+                    retries += 1
+                    all_ready = False
+                    break
```
Sami1309, line 179: all these test loops are identical so they could just be a single function with parameters, (test_resource(endpoint, expected))

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,298 @@
...
+            fi
+          done
+          echo "Containers Ready"
+          kubectl get pods
+        shell: bash
+
+      - name: Enable Minikube Ingress
+        run: |
+          kubectl port-forward svc/featureform-ingress-nginx-controller 8000:443 7000:80 &
+          sleep 2
```
Sami1309, line 290: is this always enough time to guarantee port forwarding is up and running. alternative would be to run a curl do while loop until it returns true to guarantee next command doesn't run until it works. Very minor though

do curl:7000
timer=0
while curl is false and timer <10
  timer++
  curl:7000
  sleep 1

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,298 @@
...
+          helm repo add jetstack https://charts.jetstack.io
+          helm repo update
+          helm install certmgr jetstack/cert-manager --set installCRDs=true --version v1.8.0 --namespace cert-manager --create-namespace
+          helm install featureform ./charts/featureform --set global.repo=local --set global.pullPolicy=Never
+          helm install quickstart ./charts/quickstart
+
+      - name: Check Pods
+        run: |
+          seconds=0
+          while [ $(kubectl get pods | grep "featureform" | grep -wv featureform-quickstart | awk {'print $3'} | column -t | grep -wv Running -c ) != 0 ]
```
Sami1309, line 271: you can query pods based on a label set in the config, or based on their service name if specified in the charts. Set by the metadata labels in the yaml files

kubectl get pods -l app=featureform

avoids needing to pattern match indirectly

Also kubectl has a wait command which pauses until a queried container is ready

kubectl wait --for-condition=ready pod -l app=featureform

Since this works though it's fine

Sami1309 (commented):
very nice 10/10

client/tests/e2e.py
sdreyer: True, will do

.github/workflows/e2e.yml
sdreyer: Changed this

.github/workflows/e2e.yml
sdreyer: Changed this as well

Sami1309 (approved):

PR #289, opened 2022-06-22
Adding support for Redshift

ahmadnazeri:
Adding the support for AWS Redshift.

### DIFF
--- a/provider/offline_test.go
+++ b/provider/offline_test.go
@@ -50,6 +50,15 @@ func TestOfflineStores(t *testing.T) {
 	}
 	defer destroySnowflakeDatabase(snowflakeConfig)
 
+	var redshiftConfig = RedshiftConfig{
+		Endpoint: os.Getenv("REDSHIFT_ENDPOINT"),
+		Port:     os.Getenv("REDSHIFT_PORT"),
+		Database: os.Getenv("REDSHIFT_DATABASE"),
+		Username: os.Getenv("REDSHIFT_USERNAME"),
+		Password: os.Getenv("REDSHIFT_PASSWORD"),
+	}
+	serialRSConfig := redshiftConfig.Serialize()
+
 	testFns := map[string]func(*testing.T, OfflineStore){
 		"CreateGetTable":          testCreateGetOfflineTable,
 		"TableAlreadyExists":      testOfflineTableAlreadyExists,
@@ -86,6 +95,7 @@ func TestOfflineStores(t *testing.T) {
 		{MemoryOffline, []byte{}, false},
 		{PostgresOffline, serialPGConfig, true},
 		{SnowflakeOffline, serialSFConfig, true},
+		{RedshiftOffline, serialRSConfig, true},
 	}
 	for _, testItem := range testList {
 		if testing.Short() && testItem.integrationTest {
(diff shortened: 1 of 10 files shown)


.github/workflows/testing.yml
```diff
@@ -10,6 +10,9 @@ env:
   POSTGRES_PASSWORD: "password"
   ETCD_HOST: "localhost"
   ETCD_PORT: 2379
+  REDSHIFT_ENDPOINT: redshift-testing.ci0jtp5exqlr.us-east-1.redshift.amazonaws.com
```
sdreyer, line 13: Add this into github as a secret

provider/offline_test.go
```diff
@@ -375,21 +385,36 @@ func testMaterializations(t *testing.T, store OfflineStore) {
...
+		expectedRows := test.ExpectedSegment
 		for seg.Next() {
 			actual := seg.Value()
-			expected := test.ExpectedSegment[i]
-			if !reflect.DeepEqual(actual, expected) {
-				t.Fatalf("Value not equal in materialization: %v %v\n"+
-					"%T:%T %T:%T %T:%T (received:expected)\n", actual, expected, actual.Entity, expected.Entity, actual.Value, expected.Value, actual.TS, expected.TS)
+
+			found := false
+			for i, expRow := range expectedRows {
```
sdreyer, line 395: So if things are ordering properly now, can we revert this back to how it was before?

provider/offline_test.go
```diff
@@ -2310,12 +2366,20 @@ func testChainTransform(t *testing.T, store OfflineStore) {
 			t.Fatalf("Could not write value: %v: %v", err, value)
 		}
 	}
+
+	configQuery := ""
+	if store.Type() == "REDSHIFT_OFFLINE" {
+		configQuery = fmt.Sprintf("SELECT entity, \"int\", flt, str FROM %s", sanitize(table.GetName()))
```
sdreyer, line 2372: whats this needed for?

provider/redshift.go
```diff
@@ -0,0 +1,282 @@
...
+	"context"
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	_ "github.com/lib/pq"
+	sf "github.com/snowflakedb/gosnowflake"
```
sdreyer, line 13: Dont think we need snowflake here

sdreyer (commented):
Looking pretty good. Mainly minor things

provider/offline_test.go
sdreyer: Thats bad planning on my part. You can just modify the column name so its valid

provider/offline_test.go
ahmadnazeri: the materialization isn't necessarily ordered in the expected order.

provider/offline_test.go
ahmadnazeri: will make that change with the BigQuery PR change if you don't mind?

provider/offline_test.go
ahmadnazeri: yes!

sdreyer: Alright thats fine

provider/offline_test.go
ahmadnazeri: Done! I thought it would be more complex change and wanted to close the PR.

PR #293, opened 2022-06-28
Python and API server scheduler extension

Sami1309:
Allows user to add an optional schedule param to resources, as well as implementing safe client protocol for the "update schedule" function, which updates the schedule of an existing resource.


client/src/featureform/resources.py
```diff
@@ -93,33 +93,7 @@ def serialize(self) -> bytes:
         return bytes(json.dumps(config), "utf-8")
 
 
-@typechecked
-@dataclass
-class RedshiftConfig:
```
sdreyer, line 98: why remove this?

Sami1309: my client version must have not been pulled with the latest redshift changes

sdreyer (dismissed):

PR #298, opened 2022-07-01
Feature/python get functions/steffi

steffitan23:
Add list functions, split code up into multiple files, add format functions, clean up for get functions.


api/api.go
```diff
@@ -397,16 +397,158 @@ func (serv *MetadataServer) GetModels(stream pb.Api_GetModelsServer) error {
...
+		}
+	}
+}
+func (serv *MetadataServer) ListProviders(in *pb.Empty, stream pb.Api_ListProvidersServer) error {
+	proxyStream, err := serv.meta.ListProviders(stream.Context(), in)
+	if err != nil {
+		return err
+	}
+	for {
+		res, err := proxyStream.Recv()
```
Sami1309, line 545: this snippet of code is pasted so frequently it could be its own function if possible that returns either a nil, err or send err

client/src/featureform/cli.py
```diff
@@ -52,134 +53,131 @@ def cli():
...
-    elif resource_type == "model":
+            raise ValueError("Variant not needed")
+        rc.get_entity(name)
+    elif resource_type == "provider":
         if variant:
-            print("Variant not needed.")
-            return
-        else:
-            GetModel(stub, name)
+            raise ValueError("Variant not needed")
```
Sami1309, line 118: stylistically I don't like long if/else chains. I feel some sort of dictionary/function of functions/errors would be better suited for this section

client/src/featureform/cli.py
```diff
@@ -52,134 +53,131 @@ def cli():
...
+        rc.list_training_sets()
+    elif resource_type == "users":
+        rc.list_users()
+    elif resource_type == "models":
+        rc.list_models()
+    elif resource_type == "entities":
+        rc.list_entities()
+    elif resource_type == "providers":
+        rc.list_providers()
+    else:
```
Sami1309, line 178: same here

client/src/featureform/format.py
```diff
@@ -0,0 +1,26 @@
...
+divider = "-----------------------------------------------"
+
+def formatRows(formatObj, formatObj2=None, formatObj3=None, formatObj4=None):
+    if not formatObj2 and not formatObj3:
+        for s in formatObj:
+            if len(s) == 2:
+                formatRows(s[0], s[1])
+            elif len(s) == 3:
+                formatRows(s[0], s[1], s[2])
+            elif len(s) == 3:
```
Sami1309, line 13: two elifs in sequence with the same condition

Sami1309 (changes requested):
just stylistic things for now but one potentially important comment

api/api.go
steffitan23: can't do this because the args are different (streams are different types) for each function

api/api.go
```diff
@@ -397,16 +397,158 @@ func (serv *MetadataServer) GetModels(stream pb.Api_GetModelsServer) error {
 	}
 }
 
+func (serv *MetadataServer) ListUsers(in *pb.Empty, stream pb.Api_ListUsersServer) error {
+	proxyStream, err := serv.meta.ListUsers(stream.Context(), in)
+	if err != nil {
+		return err
+	}
+	for {
```
sdreyer, line 405: I believe sam commented on this, but you can put this whole for-loop into a function an reuse it for most of these

client/src/featureform/cli.py
```diff
@@ -52,134 +53,47 @@ def cli():
 @click.option("--insecure",
               is_flag=True,
               help="Disables TLS verification")
+@click.option("--local",
```
sdreyer, line 56: Does local work for this?

client/src/featureform/cli.py
```diff
@@ -215,7 +131,54 @@ def apply(host, cert, insecure, local, files):
...
+              "cert",
+              required=False,
+              help="Path to self-signed TLS certificate")
+@click.option("--insecure",
+              is_flag=True,
+              help="Disables TLS verification")
+@click.option("--local",
+              is_flag=True,
+              help="Enable local mode")
+def apply(host, cert, insecure, local, files):
```
sdreyer, line 172: was this removed from somewhere? whys it being readded?

client/src/featureform/resources.py
```diff
@@ -679,7 +734,10 @@ def create_all_local(self) -> None:
     def create_all(self, stub) -> None:
         for resource in self.__create_list:
             try:
-                print("Creating", resource.type(), resource.name)
+                if resource.type()[:3] == "get":
```
sdreyer, line 737: Whats this doing?

api/api.go
steffitan23: i tried doing that before, but couldn't since the streams are all considered different types (api user stream is a different type than api features stream). i tried making one with a generic type, but it got weirdly complicated

client/src/featureform/resources.py
steffitan23: checks if the resource is a ProviderReference or EntityReference, etc. (the _create function for those objects are actually getting the object), so the message is different

client/src/featureform/cli.py
sdreyer: What happens when someone runs GET with the --local flag right now?

client/src/featureform/cli.py
```diff
@@ -48,133 +53,52 @@ def cli():
...
-            GetEntity(stub, name)
-    elif resource_type == "model":
-        if variant:
-            print("Variant not needed.")
-            return
-        else:
-            GetModel(stub, name)
+        rc = ResourceClient(host, True, cert)
+
+    funcDictWithVariant = {
```
sdreyer, line 80: should be snake case

client/src/featureform/cli.py
```diff
@@ -48,133 +53,52 @@ def cli():
...
+        "feature": rc.get_feature,
+        "label": rc.get_label,
+        "source": rc.get_source,
+        "trainingset": rc.get_training_set,
+        "training-set": rc.get_training_set
+    }
+
+    funcDictNoVariant = {
```
sdreyer, line 88: also snake case

client/src/featureform/cli.py
sdreyer: Also rename to something more descriptive

client/src/featureform/cli.py
```diff
@@ -183,6 +107,54 @@ def get(host, cert, insecure, resource_type, name, variant):
...
+        host = os.getenv('FEATUREFORM_HOST')
+        if host == None:
+            raise ValueError(
+                "Host value must be set with --host flag or in env as FEATUREFORM_HOST")
+    if insecure:
+        rc = ResourceClient(host, False)
+    else:
+        rc = ResourceClient(host, True, cert)
+
+    funcDict = {
```
sdreyer, line 141: Same as above. snake case with a more descriptive name

client/src/featureform/client_test.py
```diff
@@ -0,0 +1,70 @@
+import featureform as ff
```
sdreyer, line 1: Instead of including this, if you add the GET functions to client/tests/e2e.py then they will be tested end to end each time

client/src/featureform/format.py
```diff
@@ -0,0 +1,26 @@
+twoRowSpacing = "{:<30} {:<25}"
+threeRowSpacing = "{:<30} {:<30} {:<30}"
+fourRowSpacing = "{:<30} {:<30} {:<30} {:<30}"
+divider = "-----------------------------------------------"
+
+def formatRows(formatObj, formatObj2=None, formatObj3=None, formatObj4=None):
```
sdreyer, line 6: snake_case on everything here

client/src/featureform/get.py
```diff
@@ -1,252 +1,211 @@
...
-                    s.name, s.variant, "source"))
+                formatRows(
+                    s.name, s.variant, "source")
+            formatNewPara("")
+            return user
     except grpc._channel._MultiThreadedRendezvous:
         print("User not found.")
 
-def GetFeature(stub, name):
+def GetEntity(stub, name):
```
sdreyer, line 30: snake case

client/src/featureform/resource_client_test.py
```diff
@@ -0,0 +1,76 @@
+import featureform as ff
```
sdreyer, line 1: Is this being used?

client/src/featureform/resources.py
```diff
@@ -533,6 +561,73 @@ def _create_label_resource(self, db) -> None:
...
+@typechecked
+@dataclass
+class SourceReference:
+    name: str
+    variant: str
+    obj: Union[Source, None]
+
+    @staticmethod
+    def operation_type() -> str:
+        return "get"
```
sdreyer, line 618: these can be enums I believe

client/src/featureform/format.py
```diff
@@ -0,0 +1,26 @@
...
+            else:
+                return "Tuple length not formattable."
+    elif format_obj_2 and not format_obj_3:
+        print(two_row_spacing.format(format_obj, format_obj_2))
+    elif format_obj_2 and format_obj_3 and not format_obj_4:
+        print(three_row_spacing.format(format_obj, format_obj_2, format_obj_3))
+    else:
+        print(four_row_spacing.format(format_obj, format_obj_2, format_obj_3, format_obj_4))
+
+def format_new_para(s):
```
ahmadnazeri, line 24: you can make the parameter `s=""` and when you calling it, you don't have to pass the empty string.

client/src/featureform/get.py
```diff
@@ -1,252 +1,211 @@
...
-            print("{:<20} {:<15}".format("NAME: ", x.name))
-            print("{:<20} {:<15}".format("STATUS: ", x.status.Status._enum_type.values[x.status.status].name))
-            print("VARIANTS:")
-            print("{:<20} {:<25}".format(x.default_variant, 'default'))
-            for v in x.variants:
-                if v != x.default_variant:
-                    print("{:<20} {:<35}".format(v, ''))
+                format_rows(t.name, t.variant)
+            format_new_para("")
+            return x
```
ahmadnazeri, line 209: Is there a reason why this returns after one iteration?

client/src/featureform/list.py
```diff
@@ -0,0 +1,77 @@
+from featureform.proto import metadata_pb2
+from .format import *
+
+def list_name_status(stub, resource_type):
+    stub_list_functions = {
+        "entity": stub.ListEntities,
+        "user": stub.ListUsers
+    }
+    format_rows("NAME", "STATUS")
+    res = []
```
ahmadnazeri, line 10: does `res` need to be sorted?

client/src/featureform/serving.py
```diff
@@ -9,10 +9,11 @@
 
 import grpc
 import numpy as np
+from featureform.proto import serving_pb2
+from featureform.proto import serving_pb2_grpc
+import random
```
ahmadnazeri, line 14: importing `random` and `os` twice.

client/src/featureform/resource_client_test.py
```diff
@@ -0,0 +1,80 @@
...
```
ahmadnazeri, line 1: how is this file being used? is it a test file?

client/src/featureform/list.py
steffitan23: yea, for the tests, because otherwise, sometimes the resources get printed in a different order

client/src/featureform/resource_client_test.py
steffitan23: it's a test file that's not being used, shld i delete it?

client/src/featureform/list.py
ahmadnazeri: You can sort it short hand and even pull it out as separate function. 
```
res = sorted([f for f in stub_list_functions[resource_type](metadata_pb2.Empty())], key=lambda x:x.name)
for f in res:
      format_rows(f.name, f.status.Status._enum_type.values[f.status.status].name)
```

This will remove the need for two lists. Makes it a little cleaner

client/src/featureform/resource_client_test.py
ahmadnazeri: If it is not being used and not needed then I would delete it.

client/src/featureform/get.py
steffitan23: it just returns the first in the stream

ahmadnazeri (approved):
LGTM

PR #307, opened 2022-07-13
Local Mode E2E Testing

sdreyer:


client/setup.cfg
```diff
@@ -28,6 +28,7 @@ install_requires =
     numpy==1.21.6
     pandas==1.3.5
     typing_extensions==4.3.0
+    dataclasses==0.6
```
simba-git, line 31: didn't we remove this for a reason?

sdreyer: We removed something else I believe. This is still used and was downloaded as a dependency by a different package, but the 3.7 compatible version doesn't download it as a dependency so I had to add it to ours as a direct dependency

sdreyer: We removed a type of Dict i believe

PR #311, opened 2022-07-15
Localmode/testing/riddhi

RiddhiBagadiaa:


client/src/featureform/serving_test.py
```diff
@@ -6,6 +6,15 @@
 print(fpf)
 # # fpf = client.features([("SepalLength", "centimeters")], {"CustomerID": "1"})
 # # Run features through model
-training_set = client.training_set("join", "v1")
+training_set = client.training_set("jjoin", "v1")
```
Sami1309, line 9: why jjoin

RiddhiBagadiaa: Was testing error handling. Ignore this branch for now

client/src/featureform/local_test.py
```diff
@@ -199,3 +199,19 @@ def transform3(df):
     features=[("SepalLength", "join"), ("SepalWidth", "join"), ("PetalLength", "join"),
               ("PetalWidth", "join")],
 )
+
+# Should error
+ff.register_training_set(
```
sdreyer, line 204: Is there something that checks if this is supposed to error?

client/src/featureform/resources.py
```diff
@@ -578,20 +578,24 @@ def _create(self, stub) -> None:
         stub.CreateTrainingSetVariant(serialized)
 
     def _create_local(self, db) -> None:
+        self._insert_training_set_features(db)
```
sdreyer, line 581: Whats this for?

client/src/featureform/serving_test.py
sdreyer: Does this need to be changed back?

client/src/featureform/serving_test.py
```diff
@@ -6,6 +6,15 @@
...
 for r in training_set:
     print(r)
+
+# ff.register_user("featureformer").make_default_owner()
```
sdreyer, line 13: remove?

client/src/featureform/resources.py
RiddhiBagadiaa: This stores the feature name and the training associated with it in a separate table so that it is easy to access the corresponding features

client/src/featureform/local_test.py
```diff
@@ -199,3 +195,25 @@ def transform3(df):
     features=[("SepalLength", "join"), ("SepalWidth", "join"), ("PetalLength", "join"),
               ("PetalWidth", "join")],
 )
+
+def add_invalid_label():
```
sdreyer, line 199: If you want to use this as a pytest file, you should remove the other register functions from above. Seems like a weird hybrid definitions + test file instead of one or the other right now. Or just make everything a test function

client/src/featureform/resources.py
```diff
@@ -578,20 +578,24 @@ def _create(self, stub) -> None:
...
+        try:
+            db.getNameVariant("labels_variant", "labelName", self.label[0], "variantName", self.label[1])
```
sdreyer, line 583: Is this done for features too or only labels?

client/src/featureform/resources.py
```diff
@@ -578,20 +578,24 @@ def _create(self, stub) -> None:
...
+        except ValueError:
+            raise ValueError(self.label[0] + " does not exist. Failed to register training set")
```
sdreyer, line 585: Typically cleaner to do something like:
"Failed to register training set. Label does not exist: {}".format(label[0])

client/src/featureform/resources.py
```diff
@@ -603,6 +607,10 @@ def _create_training_set_resource(self, db) -> None:
 
     def _insert_training_set_features(self, db) -> None:
         for feature in self.features:
+            try:
```
sdreyer, line 610: I see the features are here. Can you move the label call inside of this functions instead and change the name to insert_training_set_resources ? 

If it makes sense

client/src/featureform/sqlite_metadata.py
```diff
@@ -194,13 +193,19 @@ def getVariantResource(self, type, column, resource):
...
         return variant_data.fetchall()
 
     def getNameVariant(self, type, column1, resource1, column2, resource2):
         variant_table_query = "SELECT * FROM " + type + " WHERE " + column1 + "='" + resource1 + "' AND " + column2 + "='" + resource2 + "';"
         variant_data = self.__conn.execute(variant_table_query)
         self.__conn.commit()
-        return variant_data.fetchall()
+        variant_data_list = variant_data.fetchall()
+        if len(variant_data_list) == 0:
+          raise ValueError(type + " with " + column1 + " : " + resource1 + " and " + column2 + " : " + resource2 + " not found")
```
sdreyer, line 207: Should use "  ".format() for these. Makes it a lot easier to read

client/src/featureform/sqlite_metadata.py
```diff
@@ -194,13 +193,19 @@ def getVariantResource(self, type, column, resource):
         variant_table_query = "SELECT * FROM " + type + " WHERE " + column + "='" + resource + "';"
         variant_data = self.__conn.execute(variant_table_query)
         self.__conn.commit()
+        variant_data_list = variant_data.fetchall()
+        if len(variant_data_list) == 0:
+          raise ValueError(type + " with " + column + " : " + resource + " not found")
         return variant_data.fetchall()
 
     def getNameVariant(self, type, column1, resource1, column2, resource2):
```
sdreyer, line 201: These need to be changed to snake case, but I think thats part of the refactor task

client/src/featureform/resources.py
sdreyer: Yeah its just a little odd that the label one uses the database directly, but the training set one has its own function that takes the datbase as an arg. Feel like both of them should just have a function or be in the same function

RiddhiBagadiaa: Okayy, Ill modify the function and put them together

client/src/featureform/register.py
```diff
@@ -813,24 +813,46 @@ def register_training_set(self,
...
+            channel = self.tls_check(host, cert_path, insecure)
+            self._stub = ff_grpc.ApiStub(channel)
+
+        self.apply()
+
+    def tls_check(self, host, cert, insecure):
+        if insecure:
+            channel = grpc.insecure_channel(
+                host, options=(('grpc.enable_http_proxy', 0),))
+        elif cert != None or os.getenv('FEATUREFORM_CERT') != None:
```
ahmadnazeri, line 839: Can we do this instead before line 836? 
```
cert = cert if cert else os.getenv('FEATUREFORM_CERT')
```
then this elif condition can be just 
```
elif cert != None
```
and we can remove the nested if condition (lines 840 -841)

Let me know if I am understanding the code correctly and if this makes sense.

client/src/featureform/sqlite_metadata.py
```diff
@@ -194,13 +193,19 @@ def getVariantResource(self, type, column, resource):
...
         variant_table_query = "SELECT * FROM " + type + " WHERE " + column1 + "='" + resource1 + "' AND " + column2 + "='" + resource2 + "';"
         variant_data = self.__conn.execute(variant_table_query)
         self.__conn.commit()
-        return variant_data.fetchall()
+        variant_data_list = variant_data.fetchall()
+        if len(variant_data_list) == 0:
+          raise ValueError("{} with {}: {} and {}: {} not found".format(type, column1, resource1, column2, resource2))
```
ahmadnazeri, line 207: For future reference, leveraging f-strings makes it a lot easier to understand.
Here is how this one would look:
```
f"{type} with {column1}: {resource1} and {column2}: {resource2} not found"
```

ahmadnazeri (approved):
LGTM, just two small comments.

PR #315, opened 2022-07-18
Combined client and localclient in serving

RiddhiBagadiaa:


client/src/featureform/serving.py
```diff
@@ -15,90 +15,37 @@
...
 
-    def processLabelCSV(self, source_path, entity_name, labelColumnName, timestamp_column):
-        df = pd.read_csv(source_path)
-        if timestamp_column != "":
-            df = df[[entity_name, labelColumnName, timestamp_column]]
-        else:
-            df = df[[entity_name, labelColumnName]]
-        df.set_index(entity_name)
-        return df
+    def dataset(self, name, version):
```
sdreyer, line 41: I think this should probably be the same as training_set right? Since we want to APIs to look the same. @simba-git

PR #318, opened 2022-07-20
Other/localmode feature test/sterling

sdreyer:


client/examples/local_serving.py
```diff
@@ -1,17 +1,15 @@
...
-#     print(r)
-#     i+=1
-#     if i > 10:
 import featureform as ff
 
+client = ff.ServingLocalClient()
+
+training_set = client.training_set("fraud_training", "quickstart")
+i=0
+for r in training_set:
```
ahmadnazeri, line 7: this might be a little cleaner if we do 
`for i, r in enumerate(training_set):
       print(r)
       if i > 10:
           break
`

client/src/featureform/serving.py
```diff
@@ -99,7 +99,7 @@ def _local_training_set(self, trainingSetName, trainingSetVariant):
                 df.set_index(feature_row[9])
 
                 df.rename(columns={feature_row[11]: name_variant}, inplace=True)
-                feature_df = df#.sort_values(by=feature_row[10], ascending=False)
+                feature_df = df
```
ahmadnazeri, line 102: Is there still a reason for this?

client/src/featureform/serving.py
```diff
@@ -99,7 +99,7 @@ def _local_training_set(self, trainingSetName, trainingSetVariant):
                 df.set_index(feature_row[9])
```
ahmadnazeri, line 99: Just out of curiosity, what do the numbers represent? the `9`, `10`, `11` inside the `feature_row`.

client/src/featureform/serving.py
```diff
@@ -168,54 +166,70 @@ def process_transformation(self, name, variant):
...
-                feature_dataframes.add(featureVariantTuple[0])
-                dataframe_mapping[featureVariantTuple[0]] = df
+                dataframe_mapping.append(df)
             else:
-                feature_dataframes, dataframe_mapping = self.process_feature_csv(source_row[10], entity_tuple[0],
-                                                                                 feature_dataframes,
-                                                                                 feature_column_name,
-                                                                                 dataframe_mapping,
-                                                                                 featureVariantTuple[0], "")
+                name_variant = "{}.{}".format(featureVariantTuple[0], featureVariantTuple[1])
```
ahmadnazeri, line 189: f-string is a cleaner way of doing formats as long as we only support Python >3.6. 
```
name_variant = "{featureVariantTuple[0]}.{featureVariantTuple[1]}"
```

client/tests/test_serving.py
```diff
@@ -0,0 +1,181 @@
...
+import shutil
+import time
+from tempfile import NamedTemporaryFile
+from unittest import TestCase
+import os, stat
+import pandas as pd
+import pytest
+from featureform import ServingClient, ResourceClient
+
+from localmode_cases import features_no_ts, features_with_ts, feature_invalid_entity, feature_invalid_value, \
```
ahmadnazeri, line 11: [conftest](https://www.tutorialspoint.com/pytest/pytest_conftest_py.htm) might be a better approach getting the features.

ahmadnazeri (dismissed):
Just a few small comments that could be addressed later on if needed.

client/src/featureform/serving.py
sdreyer: 9 is entity name, believe 11 is the value column name and 11 is the timestamp column name. Part of ksshirajas task this week is to refactor this so were not using magic numbers anymore

client/src/featureform/serving.py
sdreyer: Just for clarity mainly, but unnecessary

client/src/featureform/serving.py
sdreyer: Done. I'm still in my Python2.7 days

client/tests/test_serving.py
sdreyer: Is there a better approach than adding them all into a single conftest file? The localmode file is getting pretty long and this is just the first set of cases

ahmadnazeri: the conftest would allow the fixtures to be used across different test files without importing it. If they are only being used in a particular file, we can add them as fixtures to that file.

ahmadnazeri (approved):
LGTM

PR #330, opened 2022-07-22
Added localmode quickstart tests to single file

sdreyer:
Tests the quickstart using the resource client. 
Creates a general quickstart testing class that can be used for both the CLI and Resource Client


ahmadnazeri (dismissed):
LGTM

PR #332, opened 2022-07-25
made default variant

RiddhiBagadiaa:


sdreyer:
Should remove .featureform/SQLiteDB/metadata.db and add it to .gitignore

.gitignore
```diff
@@ -17,4 +17,5 @@ rm.**
 venv/
 terraform.tfstate
 .venv
-
+client/src/.featureform
```
sdreyer, line 20: You should make this **/.featureform instead so it isnt neccesarily in that path

sdreyer: Also make sure that .featureform directory is removed. I think the directory still exists on this branch

client/src/featureform/resources.py
```diff
@@ -366,12 +366,12 @@ def kwargs(self):
 @dataclass
 class Source:
     name: str
-    variant: str
     definition: SourceDefinition
     owner: str
     provider: str
     description: str
     schedule: str = ""
+    variant: str = "default"
```
sdreyer, line 374: Move this above schedule

.gitignore
RiddhiBagadiaa: Will change it. I dont see the directory anymore?

client/src/featureform/cli.py
```diff
@@ -24,7 +24,8 @@
 ]
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
-
+env_host = os.getenv('HOST')
```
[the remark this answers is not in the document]
RiddhiBagadiaa: I dont think so? I just added this so that while testing you dont need to write the the long host and cert name repeatedly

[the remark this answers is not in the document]
sdreyer: It got moved to register.py. All that logic for the env vars should already be in there

sdreyer (approved):

PR #333, opened 2022-07-25
Spark Offline Store Minimal Factory Register Functions

Sami1309:
Outline of spark store that functional implementations depend on.


ahmadnazeri (dismissed):
LGTM

PR #334, opened 2022-07-26
Make hosted/local mode serving have the same API.

simba-git:


client/src/featureform/serving.py
```diff
@@ -160,9 +160,12 @@ def process_transformation(self, name, variant):
         new_data = func(*dataframes)
         return new_data
 
-    def _local_features(self, feature_variant_list, entity_tuple):
+    def _local_features(self, feature_variant_list, entity):
         if len(feature_variant_list) == 0:
             raise Exception("No features provided")
+        # This code was originally written to take a tuple, this is a quick fix to turn a dict with a single entry into that tuple.
+        # This should all be refactored later.
+        entity_tuple = list(entity.items())[0]
```
RiddhiBagadiaa, line 168: entity.items() returns a list of tuples. So I dont think we need the list() function?

simba-git: It returns a dict_items which is not subscriptable. Have to convert it to a list first.

sdreyer (approved):

ahmadnazeri (approved):

PR #335, opened 2022-07-26
Change batch serving to return a BatchRow

simba-git:


RiddhiBagadiaa (dismissed):

sdreyer (dismissed):

client/src/featureform/serving.py
```diff
@@ -445,6 +445,33 @@ def to_numpy(self):
...
+    def features(self):
+        return self._features
+
+    def labels(self):
+        return self._labels
+
+    def to_list(self):
+        return self._rows
+
+    def len(self):
```
ahmadnazeri, line 473: would recommend using 
```
def __len__(self):
      return len(self._rows)
```
because the user would expect to use `len(rows)`. 
If you agree and we change this, then line 364 can be reverted back to `len(rows)`.

ahmadnazeri (approved):
LGTM, just one comment about `BatchRow.len()`

ahmadnazeri (approved):

PR #347, opened 2022-07-27
Localmode/duplicate testing/riddhi

RiddhiBagadiaa:


RiddhiBagadiaa:
The purpose of this PR is to check if the exact same resource is being registered again. If it is then instead of throwing an error, something is printed and the file continues running

client/src/featureform/resources.py
```diff
@@ -277,6 +276,22 @@ def _create_local(self, db) -> None:
...
+            return False
+        if self.team != other.team:
+            return False
+        if str(self.config.serialize(), 'utf-8') != str(other.config.serialize(), 'utf-8'):
+            return False
+        if self.config.type() != other.config.type():
+            return False
+        if self.config.software() != other.config.software():
+            return False
+        return True
```
Sami1309, line 292: there are ways of cycling though all the attributes of a class iteratively rather than hardcoding every comparison
https://stackoverflow.com/questions/11637293/iterate-over-object-attributes-in-python
stuff like this would be cleaner, and would also mean these methods wouldn't need to be updated every time an attribute is added to a class

client/src/featureform/resources.py
```diff
@@ -373,6 +385,8 @@ class Source:
     variant: str = "default"
     schedule: str = ""
     schedule_obj: Schedule = None
+    is_transformation = 0
```
ahmadnazeri, line 388: Is there a reason why we are using `0` and `1` versus `False` and `True`?

client/src/featureform/resources.py
```diff
@@ -771,19 +807,25 @@ def _check_insert_training_set_resources(self, db) -> None:
...
         except ValueError:
             raise ValueError("{} does not exist. Failed to register training set".format(self.label[0]))
-        for feature in self.features:
+        for feature_name, feature_variant in self.features:
             try:
-                db.getNameVariant("feature_variant", "name", feature[0], "variant", feature[1])
+                db.getNameVariant("feature_variant", "name", feature_name, "variant", feature_variant)
             except ValueError:
-                raise ValueError("{} does not exist. Failed to register training set".format(feature[0]))
+                raise ValueError("{} does not exist. Failed to register training set".format(feature_name))
```
ahmadnazeri, line 814: Would recommend using f-strings. 
```
f"{feature_name} does not exist. Failed to register training set"
``` 
It makes it easier to read when looking at the code.

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
     label = get_label(df, entity, label)
     training_set_df = get_training_set(label, feature, entity)
     return training_set_df.values.tolist()
+
+class TestLocalMode:
+    def test_duplicate_resources(self):
+        ff.register_user("featureformer").make_default_owner()
+
+        local = ff.register_local()
```
ahmadnazeri, line 118: Is there a reason for the duplicated code?

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
...
+        local = ff.register_local()
+
+        transactions = local.register_file(
```
ahmadnazeri, line 121: duplicated code

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
...
+        )
+
+        transactions = local.register_file(
+            name="transactions",
+            variant="quickstart",
+            description="A dataset of fraudulent transactions",
+            path="transactions.csv"
+        )
+
+        user = ff.register_entity("user")
```
ahmadnazeri, line 135: duplicated code

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
...
+        # Register a column from our transformation as a feature
+        average_user_transaction.register_resources(
+            entity=user,
+            entity_column="CustomerID",
+            inference_store=local,
+            features=[
+                {"name": "avg_transactions", "variant": "quickstart", "column": "TransactionAmount", "type": "float32"},
+            ],
+        )
+        average_user_transaction.register_resources(
```
ahmadnazeri, line 153: duplicated code

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
...
+
+        # Register label from our base Transactions table
+        transactions.register_resources(
+            entity=user,
+            entity_column="CustomerID",
+            labels=[
+                {"name": "fraudulent", "variant": "quickstart", "column": "IsFraud", "type": "bool"},
+            ],
+        )
+        transactions.register_resources(
```
ahmadnazeri, line 170: duplicated code

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
...
+            ],
+        )
+
+        ff.register_training_set(
+            "fraud_training", "quickstart",
+            label=("fraudulent", "quickstart"),
+            features=[("avg_transactions", "quickstart")],
+        )
+
+        ff.register_training_set(
```
ahmadnazeri, line 184: duplicated code

client/tests/localmode.py
```diff
@@ -108,3 +110,159 @@ def get_training_set_from_file(file, entity, feature_col, label, name_variant):
...
+        # Register a column from our transformation as a feature
+        average_user_transaction.register_resources(
+            entity=user,
+            entity_column="CustomerID",
+            inference_store=local,
+            features=[
+                {"name": "avg_transactions", "variant": "quickstart", "column": "TransactionAmount", "type": "float32"},
+            ],
+        )
+        average_user_transaction.register_resources(
```
ahmadnazeri, line 225: duplicated code

ahmadnazeri (commented):
I have a few questions but overall, it looks good.

client/tests/localmode.py
RiddhiBagadiaa: Yes we want to make sure that the code doesn't break/error when the same resource is registered again. Thats why we need duplicated code

client/src/featureform/resources.py
RiddhiBagadiaa: Yes, some versions of SLQite3 dont support true and false

PR #355, opened 2022-07-29
Added makefile target for pytests

sdreyer:


Makefile
```diff
@@ -47,6 +58,7 @@ gen_grpc:						## Generates GRPC Dependencies
 
 update_python: create_venv gen_grpc 			## Updates the python package locally
 	cd test_venv/bin && source activate
+	cd ../../
```
ahmadnazeri, line 61: if we are change back directories then I would recommend `source test_venv/bin/activate` on the previous line

Makefile
```diff
@@ -22,6 +22,17 @@
 help:     						## Show this help.
 	@sed -ne '/@sed/!s/## //p' $(MAKEFILE_LIST)
 
+pytest:
+	-rm -r .featureform
```
ahmadnazeri, line 26: just curious, what's the difference between `-rm` and `rm`?

ahmadnazeri (approved):
LGTM, just a minor suggestion.

Makefile
sdreyer: -rm allows it to fail, so if that directory doesn't exist, then it skips it

Makefile
sdreyer: Actually gonna remove this. its not the proper way to access the virtual environment with the makefile anyways

ahmadnazeri (approved):
LGTM

PR #358, opened 2022-08-01
Adding support for GCP BigQuery

ahmadnazeri:
Adding support for GCP BigQuery.


provider/bigquery.go
```diff
@@ -0,0 +1,1377 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+	return newBigQueryGenericTableIterator(it, pt.query), nil
+}
+
+func (pt *bqPrimaryTable) NumRows() (int64, error) {
+	var n []bigquery.Value
+	tableName := fmt.Sprintf("%s.%s", pt.query.getTablePrefix(), pt.name)
```
sdreyer, line 131: Is this logic used elsewhere? Should it be its own function?

provider/bigquery.go
```diff
@@ -0,0 +1,1377 @@
...
+
+	bqQ := pt.client.Query(upsertQuery)
+	bqQ.Parameters = recordsParameter
+
+	_, err := bqQ.Read(pt.query.getContext())
+
+	return err
+}
+
+func (pt *bqPrimaryTable) getNonNullRecords(rec GenericRecord) ([]bigquery.QueryParameter, []TableColumn, string) {
```
sdreyer, line 164: Whats this function do?

ahmadnazeri: Basically, I couldn't add `<nil>` values so I made it that I add the non-null values and the columns with nulls will be automatically set to null.

sdreyer: Is it the driver not accepting nil values as an acceptable value or something?

sdreyer (approved):

provider/bigquery.go
```diff
@@ -0,0 +1,1382 @@
...
+		currentValue: nil,
+		err:          nil,
+		query:        query,
+	}
+}
+
+func (q defaultBQQueries) registerResources(client *bigquery.Client, tableName string, schema ResourceSchema, timestamp bool) error {
+	var query string
+	if timestamp {
+		query = fmt.Sprintf("CREATE VIEW `%s` AS SELECT %s as entity, %s as value, %s as ts FROM `%s`", q.getTableName(tableName),
```
simba-git, line 235: Is this safe? Do we need to sanitize it in any way?

provider/bigquery.go
```diff
@@ -0,0 +1,1382 @@
...
+			schema.Entity, schema.Value, schema.TS, q.getTableName(schema.SourceTable))
+	} else {
+		query = fmt.Sprintf("CREATE VIEW `%s` AS SELECT %s as entity, %s as value, to_timestamp('%s', 'YYYY-DD-MM HH24:MI:SS +0000 UTC')::TIMESTAMP as ts FROM `%s`", q.getTableName(tableName),
```
simba-git, line 238: Is this assuming the resource doesn't have a timestamp column itself?

simba-git: I'm specifically confused about teh time.UnixMilli(0).UTC() parameter

simba-git (approved):

provider/bigquery.go
ahmadnazeri: the driver doesn't accept null values in the current query approach. i.e. ```INSERT INTO table (col1) VALUES (nil)```

provider/bigquery.go
sdreyer: This is something I put in the sql offlinestore. I believe it was to make the resource tables schemas consistent to make creating the training set simpler by always having entity, value, time. If the source table doesnt have its own timestamp, it replaces it with time.UnixMilli(0).UTC() because any values earlier than that breaks some databases like snowflake.

provider/bigquery.go
ahmadnazeri: I don't see any issues. What are you thinking?

PR #359, opened 2022-08-01
Refactored strings

ksshiraja:
Refactored f strings and ensured tls_verify and insecure were consistent with each other.

### DIFF
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -59,12 +59,10 @@ class Client:
         if not self.local:
             raise ValueError("Only supported in localmode. Please try using dataset()")
         training_set_row = \
-            self.sqldb.getNameVariant("training_set_variant", "name", training_set_name, "variant",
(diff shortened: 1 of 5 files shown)


client/src/featureform/serving.py
```diff
@@ -17,7 +17,7 @@
 
 class Client:
 
-    def __init__(self, host=None, local=False, tls_verify=True, cert_path=None):
+    def __init__(self, host=None, local=False, insecure=False, cert_path=None):
```
sdreyer, line 20: make sure that serving client and resource client are consistent with this

ksshiraja: Yeah i checked this and they are consistent. All the tls_Verify variables have been changed to insecure variables for consistency.

ahmadnazeri (commented):
great work! only thing is changing the `query` variables into f-strings. Also, resolve the merge conflicts.

PR #360, opened 2022-08-01
Fix for invalid redis auth

sdreyer:


.github/workflows/testing.yml
```diff
@@ -118,13 +120,20 @@ jobs:
...
         # supported for use outside of 'steps'
         ports:
           - 6379:6379
 
+      redis-secure:
+        image: redis
+          # Hard coded port because environment variables not currently
+          # supported for use outside of 'steps'
+        ports:
+          - 6378:6379
```
Sami1309, line 135: is this mapping correct?

Sami1309 (commented):
besides comment looks good to me, merge if tests pass

.github/workflows/testing.yml
sdreyer: Yeah. For testing Ive setup a secure and insecure redis instance and test them both. Have to make one port different so they dont interfere

ahmadnazeri (dismissed):
LGTM

PR #364, opened 2022-08-02
Fixed logic for tls verification and self-signed cert

sdreyer:


client/src/featureform/register.py
```diff
@@ -1276,34 +1277,22 @@ def __init__(self, host=None, local=False, insecure=True, cert_path=None):
         """
         super().__init__()
         self.local = local
-        if self.local:
+        if local:
```
ahmadnazeri, line 1280: can we combined the if conditions? 
```
if local and host: 
     raise ValueError("Cannot run locally with a host")
```

client/src/featureform/register.py
```diff
@@ -1276,34 +1277,22 @@ def __init__(self, host=None, local=False, insecure=True, cert_path=None):
...
                 raise ValueError("Cannot be local and have a host")
         else:
-            if host == None:
-                host = os.getenv('FEATUREFORM_HOST')
-                if host == None:
-                    raise ValueError(
-                    "Host value must be set or in env as FEATUREFORM_HOST")
-
-            channel = self.tls_check(host, cert_path, insecure)
+            host = host or os.getenv('FEATUREFORM_HOST')
```
ahmadnazeri, line 1284: if host and `FEATUREFORM_HOST` are both set, would we use `host` instead of the env?

client/src/featureform/tls.py
```diff
@@ -0,0 +1,15 @@
+import grpc
+import os
+
+def insecure_channel(host):
+    return grpc.insecure_channel(host, options=(('grpc.enable_http_proxy', 0),))
+
+def secure_channel(host, cert_path):
+    cert_path = cert_path or os.getenv('FEATUREFORM_CERT')
```
ahmadnazeri, line 8: if both are set, do we want to use `cert_path` or `FEATUREFORM_CERT`?

client/src/featureform/tls.py
```diff
@@ -0,0 +1,15 @@
...
+    if cert_path != None:
```
ahmadnazeri, line 9: you can just do `if cert_path:`

ahmadnazeri (approved):
LGTM; just some questions

client/src/featureform/tls.py
sdreyer: We want cert_path because it allows an override over the env var

client/src/featureform/tls.py
sdreyer: Done

PR #368, opened 2022-08-03
Gave environment variables default values

sdreyer:
Created a helper package that creates a function for default environment variables

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202764133312840


ahmadnazeri (approved):
LGTM

PR #370, opened 2022-08-04
Change transformation decorator to input only transformation name

ksshiraja:


client/src/featureform/register.py
```diff
@@ -1086,10 +1087,13 @@ def df_transformation(self,
         Returns:
             decorator (DFTransformationDecorator): decorator
         """
+
         if not isinstance(owner, str):
             owner = owner.name()
         if owner == "":
             owner = self.must_get_default_owner()
+        if variant == "":
```
ahmadnazeri, line 1095: why can't we default it to `"default"` instead of empty string? Is it to be consistent?

ahmadnazeri (approved):
LGTM; just one question

PR #375, opened 2022-08-08
Maintain consistency between Row and BatchRow

RiddhiBagadiaa:
Changed the label function name and altered the return type such that features always returns a 2D array and label always returns a 1D array to make the training set value consistent always. This also allows sklearn's partial fit function to train on individual rows if required"

### DIFF
--- a/client/tests/serving_test.py
+++ b/client/tests/serving_test.py
@@ -325,10 +325,6 @@ class TestTransformation(TestCase):
         except:
             print("File Already Removed")
 
-
-
-
-
 class TestTrainingSet(TestCase):
     def _register_feature(self, feature, local, case, index, name):
         file = create_temp_file(feature)
@@ -338,7 +334,6 @@ class TestTrainingSet(TestCase):
             description="",
             path=file
         )
-
         test_file.register_resources(
             entity=case['entity'],
             entity_column=case['entity_loc'],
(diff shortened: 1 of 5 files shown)


client/tests/serving_test.py
```diff
@@ -401,14 +400,21 @@ def test_all(self):
...
+                expected_len = len(case['expected_features'])
                 for i, r in enumerate(tset):
                     actual_len += 1
-                    actual = r.features() + [r.label()]
-                    if actual in case['expected']:
-                        case['expected'].remove(actual)
+                    actual_features = r.features()
+                    actual_label = r.label()
+                    # actual = r.features() + [r.label()]
+                    if actual_features in case['expected_features']:
```
sdreyer, line 409: Need a way to check that the label and feature in each iteration are in the same index. otherwise there could be an incorrect pair that passes the test. For example
Expected: 
[
[feature='a', label='True']
[feature='b', label='False']
]
Actual: 
[
[feature='a', label='False']
[feature='b', label='True']
]
Would still pass

client/tests/serving_test.py
```diff
@@ -404,11 +403,18 @@ def test_all(self):
                 expected_len = len(case['expected'])
                 for i, r in enumerate(tset):
                     actual_len += 1
-                    actual = r.features() + [r.label()]
+                    # actual_features = r.features()
```
ahmadnazeri, line 406: remove the commented code if not needed.

ahmadnazeri (dismissed):
LGTM; just remove the commented code that's not needed.

ahmadnazeri (approved):

PR #380, opened 2022-08-12
Refactor register resource and register training set

ksshiraja:
Users can now assign their resources to variables which can be used in registering a training set

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -1366,7 +1366,6 @@ class Registrar:
         Returns:
             resource (ResourceRegistrar): resource
         """
-        print(features)
         if not isinstance(owner, str):
             owner = owner.name()
         if owner == "":
@@ -1381,7 +1380,6 @@ class Registrar:
             resources = []
         for resource in resources:
             features += resource.features()
-            print(features)
             resource_label = resource.label()
             if label == ():
                 label = resource_label
@@ -1393,8 +1391,7 @@ class Registrar:
             raise ValueError("Label must be set")
         if features == []:
             raise ValueError("A training-set must have atleast one feature")
-        print(name, features)
-
+        
         resource = TrainingSet(
             name=name,
             variant=variant,
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -1341,8 +1341,9 @@ class Registrar:
     def register_training_set(self,
                               name: str,
                               variant: str,
-                              label: NameVariant,
-                              features: List[NameVariant],
+                              label: NameVariant = None,
+                              features: List[NameVariant] = None,
+                              resources: Union[str, ResourceRegistrar] = "",
                               owner: Union[str, UserRegistrar] = "",
                               description: str = "",
                               schedule: str = ""):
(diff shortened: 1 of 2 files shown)


client/src/featureform/register.py
```diff
@@ -411,8 +411,8 @@ class ResourceRegistrar():
 
     def __init__(self, registrar, features, labels):
         self.__registrar = registrar
-        self.__features = features
-        self.__labels = labels
+        self.features = features
```
sdreyer, line 414: You should leave these as private and create a features() function instead

client/src/featureform/register.py
```diff
@@ -1342,6 +1354,14 @@ def register_training_set(self,
             owner = owner.name()
         if owner == "":
             owner = self.must_get_default_owner()
+        if isinstance(features,tuple):
+            raise ValueError("Features must be entered as a list")
+        if isinstance(label, list):
+            if len(label) != 1:
+                raise ValueError("Only one label can be used to register a training set")
+            label = (label[0]["name"], label[0]["variant"])
```
sdreyer, line 1362: Put this logic in a label() function instead. They should be able to register mutliple labels, but only call a single one

client/src/featureform/register.py
```diff
@@ -1316,6 +1322,26 @@ def register_column_resources(
...
+                feature_nv_list.append(feature_nv)
+            elif isinstance(feature, list):
+                feature_nv_list.extend(self.__get_feature_nv(feature))
+            else:
+                feature_nv_list.append(feature)
+        return feature_nv_list
+    
+    def __get_label_nv(self, label):
+        if isinstance(label, list):
+            if len(label) != 1:
```
sdreyer, line 1339: We want this logic inside the label() function, otherwise it won't allow us to register more than one label from a resource.

client/src/featureform/register.py
```diff
@@ -455,6 +455,12 @@ def create_training_set(self,
             schedule=schedule,
             description=description,
         )
+    
+    def features(self):
+        return self.__features
+    
+    def labels(self):
```
sdreyer, line 462: Change this to label() instead of labels()

client/src/featureform/register.py
```diff
@@ -455,6 +455,16 @@ def create_training_set(self,
...
         )
+    
+    def features(self):
+        return self.__features
+    
+    def label(self):
+        if isinstance(self.__labels, list):
+            if len(self.__labels) != 1:
+                raise ValueError("This resource has multiple labels. Please call only one label to register a training set")
+            self.__labels = (self.__labels[0]["name"], self.__labels[0]["variant"])
```
sdreyer, line 466: Need to check if labels is empty

client/src/featureform/register.py
```diff
@@ -1342,6 +1365,15 @@ def register_training_set(self,
             owner = owner.name()
         if owner == "":
             owner = self.must_get_default_owner()
+        if resources and (features or label):
+            raise ValueError("Either a resource or features/label must be entered")
+        if isinstance(features,tuple):
+            raise ValueError("Features must be entered as a list")
+        if resources != "":
+            features = resources.features()
+            label = resources.label()
```
sdreyer, line 1374: Should check if label() has a label and if the label parameter has a label. Throw an error if both or neither have a label

sdreyer: This should still check if they manually entered a label. 
Should be 
if resources.label() and label: error
if not resources.label() and not label: error

client/src/featureform/register.py
```diff
@@ -1338,10 +1366,35 @@ def register_training_set(self,
...
+            features = []
+        if resources == None:
+            resources = []
+        for resource in resources:
+            features += resource.features()
+            print(features)
+            resource_label = resource.label()
+            if label == ():
+                label = resource_label
+            elif resource_label != ():
```
sdreyer, line 1388: This will error if the label is added from resource.label()

sdreyer (approved):

PR #386, opened 2022-08-15
Spark submit script and materialization boilerplate

Sami1309:
# Description

Adds spark script to run offline provider functions in spark, along with implementation of the creatematerialization function along with tests.
Necessary step in implementing offline store in spark.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/provider/scripts/offline_store_spark_runner.py
+++ b/provider/scripts/offline_store_spark_runner.py
@@ -11,36 +11,33 @@ from pyspark.sql import SparkSession
 # source_list: [s3://featureform/{type}/{name}/{variant}, s3://featureform/{type}/{name}/{variant}...] (implicitly mapped to source_1, source_2, etc)
 
 def execute_sql_query(job_type, output_uri, sql_query, source_list):
-    with SparkSession.builder.appName("Execute SQL Query").getOrCreate() as spark:
+    try:
+        with SparkSession.builder.appName("Execute SQL Query").getOrCreate() as spark:
+            if job_type == "Transformation" or job_type == "Materialization":
+                for i, source in enumerate(source_list):
+                    source_df = spark.read.option("header","true").option("recursiveFileLookup","true").parquet(source)
+                    source_df.createOrReplaceTempView(f'source_{i}')
 
-        if job_type == "Transformation":
-            for i, source in enumerate(source_list):
-                source_df = spark.read.option("header","true").option("recursiveFileLookup","true").parquet(source)
-                source_df.createOrReplaceTempView(f'source_{i}')
+            output_dataframe = spark.sql(sql_query)
 
-        output_dataframe = spark.sql(sql_query)
+            dt = datetime.now()
+            output_uri_with_timestamp = f'{output_uri}{dt}'
 
-        dt = datetime.now()
-        output_uri_with_timestamp = f'{output_uri}/{dt}'
+            output_dataframe.write.option("header", "true").mode("overwrite").parquet(output_uri_with_timestamp)
+    except Exception as ident:
+        print(ident)
+    
 
-        output_dataframe.write.option("header", "true").mode("overwrite").parquet(output_uri_with_timestamp)
-
-
-def parse_script_args(args=None):
+if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "--job_type", choices=["Transformation", "Materialization", "Training Set"], help="type of job being run on spark") 
+        "--job_type", help="type of job being run on spark") 
     parser.add_argument(
         '--output_uri', help="output S3 file location")
     parser.add_argument(
         '--sql_query', help="The SQL query you would like to run on the data source")
     parser.add_argument(
         "--source_list", nargs="+", help="list of sources in the transformation string")
-    return parser.parse_args(args)
-
-if __name__ == "__main__":
-    parser = argparse.ArgumentParser()
-    
-    args = parser.parse_script_args()
+    args = parser.parse_args()
 
     execute_sql_query(args.job_type, args.data_source, args.output_uri, args.sql_query, args.source_list)
--- a/provider/scripts/offline_store_spark_runner.py
+++ b/provider/scripts/offline_store_spark_runner.py
@@ -13,11 +13,10 @@ from pyspark.sql import SparkSession
 def execute_sql_query(job_type, output_uri, sql_query, source_list):
     try:
         with SparkSession.builder.appName("Execute SQL Query").getOrCreate() as spark:
-            # if job_type == "Transformation" or job_type == "Materialization":
-            #     for _, source in enumerate(source_list):
-                    
-            source_df = spark.read.option("header","true").parquet(source_list[0])  
-            source_df.createOrReplaceTempView("source_0")
+            if job_type == "Transformation" or job_type == "Materialization":
+                for i, source in enumerate(source_list):          
+                    source_df = spark.read.option("header","true").parquet(source)  
+                    source_df.createOrReplaceTempView(f'source_{i}')
             output_dataframe = spark.sql(sql_query)
 
             dt = datetime.now()
--- a/provider/scripts/offline_store_spark_runner.py
+++ b/provider/scripts/offline_store_spark_runner.py
@@ -1,0 +1,42 @@
+import argparse
+from datetime import datetime
+
+from pyspark.sql import SparkSession
+
+# Expected format
+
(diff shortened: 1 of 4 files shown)


provider/scripts/offline_store_spark_runner.py
```diff
@@ -0,0 +1,42 @@
...
+
+        dt = datetime.now()
+        output_uri_with_timestamp = f'{output_uri}/{dt}'
+
+        output_dataframe.write.option("header", "true").mode("overwrite").parquet(output_uri_with_timestamp)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
```
ahmadnazeri, line 31: Seems like this argument is expecting `Transformation`, `Materialization`, or `Training Set`. I would recommend a choices parameter in the function. 
```
parser.add_argument("--job_type", choices=["Transformation", "Materialization", "Training Set"], help="type of job being run on spark") 
```

provider/scripts/offline_store_spark_runner.py
```diff
@@ -0,0 +1,42 @@
...
+        output_dataframe = spark.sql(sql_query)
+
+        dt = datetime.now()
+        output_uri_with_timestamp = f'{output_uri}/{dt}'
+
+        output_dataframe.write.option("header", "true").mode("overwrite").parquet(output_uri_with_timestamp)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
ahmadnazeri, line 30: In general, it is better to add the parser lines into it's own function so you can test the args. 

For example: 
```
def parse_args(args=None):
         parser = argparse.ArgumentParser()
         ...
         return parser.parse_args(args)
```
Then when testing the args, you can pass in arguments direction in the function.

provider/spark.go
```diff
@@ -701,8 +708,54 @@ func (spark *SparkOfflineStore) GetResourceTable(id ResourceID) (OfflineTable, e
...
+func (s *S3FeatureIterator) Err() error {
+	return nil
+}
+
 func (spark *SparkOfflineStore) CreateMaterialization(id ResourceID) (Materialization, error) {
-	return nil, nil
+	if _, err := spark.GetResourceTable(id); err != nil {
+		return nil, fmt.Errorf("Resource not registered: %v", err)
+	}
+	materializationID := ResourceID{Name: id.Name, Variant: id.Variant, Type: FeatureMaterialization}
```
sdreyer, line 745: Whats the difference between Feature and FeatureMaterialization?

provider/spark.go
```diff
@@ -701,8 +708,54 @@ func (spark *SparkOfflineStore) GetResourceTable(id ResourceID) (OfflineTable, e
...
+	destinationPath := spark.Store.ResourcePath(materializationID)
+	exists, _ := spark.Store.FileExists(destinationPath)
+	if exists {
+		return nil, fmt.Errorf("materialization already exists")
+	}
+	materializationQuery := spark.query.materializationCreate("", "source_0")
```
sdreyer, line 751: Make source_0 a const

sdreyer: And "" so its more clear what the empty string represents

provider/spark_test.go
```diff
@@ -99,7 +111,10 @@ func testRegisterResource(store *SparkOfflineStore) error {
 	if !reflect.DeepEqual(fetchedTable, table) {
 		return fmt.Errorf("Did not properly register table")
 	}
-	return nil
+
```
sdreyer, line 114: Should make this function an actual Go test

provider/spark.go
Sami1309: featurematerialization is just a materialization, i just named it that since Materialization is already a struct type in offline.go

sdreyer: Why don't the other offline stores use this?

provider/spark.go
```diff
@@ -701,8 +712,60 @@ func (spark *SparkOfflineStore) GetResourceTable(id ResourceID) (OfflineTable, e
...
+
 func (spark *SparkOfflineStore) CreateMaterialization(id ResourceID) (Materialization, error) {
-	return nil, nil
+	resourceTable, err := spark.GetResourceTable(id)
+	if err != nil {
+		return nil, fmt.Errorf("resource not registered: %v", err)
+	}
+	sparkResourceTable, ok := resourceTable.(*S3OfflineTable)
+	if !ok {
+		return nil, fmt.Errorf("could not convert offline table to sparkResourceTable")
```
sdreyer, line 753: Add the resource ID to the errorr

provider/spark.go
```diff
@@ -701,8 +712,60 @@ func (spark *SparkOfflineStore) GetResourceTable(id ResourceID) (OfflineTable, e
...
+	}
+	materializationID := ResourceID{Name: id.Name, Variant: id.Variant, Type: FeatureMaterialization}
+	destinationPath := spark.Store.ResourcePath(materializationID)
+	exists, _ := spark.Store.FileExists(destinationPath)
+	if exists {
+		return nil, fmt.Errorf("materialization already exists")
```
sdreyer, line 759: Add resource ID to the error

provider/spark.go
```diff
@@ -701,8 +712,60 @@ func (spark *SparkOfflineStore) GetResourceTable(id ResourceID) (OfflineTable, e
...
+	}
+	materializationQuery := sparkMaterializationQuery(sparkResourceTable.schema)
+	sourcePath := spark.Store.ResourcePath(id)
+	sparkArgs := spark.Store.SparkSubmitArgs(destinationPath, materializationQuery, []string{sourcePath})
+	if err := spark.Executor.RunSparkJob(sparkArgs); err != nil {
+		return nil, err
```
sdreyer, line 765: prepend the error with something descriptive

provider/spark.go
Sami1309: they store materializations under the separate category materialized view. For S3 paths to distinguish materializations we need a new type string.

provider/spark_test.go
Sami1309: it's functional rn to inherit the initialization

provider/spark_test.go
```diff
@@ -16,6 +16,58 @@ import (
...
+	for i := 0; i < 5; i++ {
+		exampleStructArray[i] = exampleStruct{
+			Name:       fmt.Sprintf("John Smith_%d", i),
+			Age:        30 + i,
+			Score:      100.4 + float32(i),
+			Winner:     false,
+			Registered: time.UnixMilli(int64(i)),
+		}
+	}
+	for i := 0; i < 5; i++ {
```
ahmadnazeri, line 30: We can either combine the two for loops into one. Or make the second for-loop start from 5 and go up to 10. `for i := 5; i < 10; i++`

ahmadnazeri (approved):
LGTM; just two small comments.

provider/spark.go
sdreyer: Whats the separate category? Aren't they all just tables?

provider/spark.go
```diff
@@ -120,6 +130,18 @@ type sparkSQLQueries struct {
 	defaultOfflineSQLQueries
 }
 
+func sparkMaterializationQuery(schema ResourceSchema) string {
```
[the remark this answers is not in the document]
Sami1309: spark needs different parameters for materialization create than defined by the type OfflineTableQueries, so it needs to be seperate from sparkSQLQueries

provider/spark_test.go
Sami1309: changed

provider/spark.go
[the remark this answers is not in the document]
ahmadnazeri: Understandable! But can we just create sparkSQLQueries that isn't based off OfflineTableQueries?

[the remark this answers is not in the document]
Sami1309: we could, I guess it makes since since there will be other spark implementations in the future. Added

provider/spark.go
Sami1309: yes, I was looking at the postgres implementation and there it's a materialized view rather than a table. The way I do paths for each resource is {type}/{name}/{variant}. Any materialization of a feature would have the same name and variant so I created a new type so they are stored in seperate pseudo S3 directories. It felt low-risk to use the existing OfflineResourceType convention as a path prefix for each type stored in S3.

sdreyer: Got it

sdreyer (approved):

sdreyer (approved):

ahmadnazeri (approved):

PR #388, opened 2022-08-16
Feature/transformation config update

ahmadnazeri:
# Description
Updating the TransformationConfig to accept `Type` and `SourceMapping`. 

```
type TransformationConfig struct {
	Type          TransformationType
	TargetTableID ResourceID
	Query         string
	SourceMapping []SourceMapping
}
```
The goal is that with the modified version of the Config, we can run Spark SQL and Dataframe jobs. 

n/a

### Select type(s) of change

- [ ] Documentation update

# Checklist:

- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -50,7 +50,7 @@ func getSourceMapping(template string, replacements map[string]string) ([]provid
 		if !has {
 			return nil, fmt.Errorf("no key set")
 		}
-		sourceMap = append(sourceMap, provider.SourceMapping{template: sanitize(replacement), source: replacement})
+		sourceMap = append(sourceMap, provider.SourceMapping{provider.SourceMapping.template: sanitize(replacement), provider.SourceMapping.source: replacement})
 		template = afterSplit[1]
 	}
 	return sourceMap, nil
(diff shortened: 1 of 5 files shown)


coordinator/coordinator.go
```diff
@@ -39,6 +39,23 @@ func templateReplace(template string, replacements map[string]string) (string, e
 	return formattedString, nil
 }
 
+func getSourceMapping(template string, replacements map[string]string) ([]provider.SourceMapping, error) {
```
Sami1309, line 42: So this is to reconcile the fact that the template replace will apply sanitized table names to the query but we still need to store the unsanitized names?

ahmadnazeri: Basically, this will be a temporary fixed that we can leverage for Spark SQL & Dataframes. Once we have time to change all the providers, this will be modified to create the list of `{"{{ template }}" : "table" }` and we can use this instead of the template replace.

Sami1309 (approved):

coordinator/coordinator.go
```diff
@@ -283,13 +300,18 @@ func (c *Coordinator) runSQLTransformationJob(transformSource *metadata.SourceVa
 	if err != nil {
 		return fmt.Errorf("map name: %w sources: %v", err, sources)
 	}
+	sourceMapping, err := getSourceMapping(templateString, sourceMap)
```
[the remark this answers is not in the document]
ahmadnazeri: yes! Then we can slowly move the other providers to using this as well.

sdreyer (approved):

PR #404, opened 2022-08-23
Updated Python SDK Docs

sdreyer:
# Description

Updated and cleaned up Python SDK docs so users have a reference for the Python API. 

Made SDK reference available from the main docs

No

### Select type(s) of change

- [x] Documentation update

# Checklist:

- [-] I have performed a self-review of my code
- [-] I have commented my code, particularly in hard-to-understand areas
- [-] I have added tests that prove my fix is effective or that my feature works
- [-] New and existing unit tests pass locally with my changes


client/src/featureform/register.py
```diff
@@ -125,6 +140,27 @@ def name(self) -> str:
...
+            description="A dataset of fraudulent transactions",
+            path="transactions.csv"
+        )
+        ```
+        Args:
+            name (str): Name for how to reference the file later
+            description (str): Description of the file
+            path (str): Path to the file
+            variant (str): File variant
+            owner (str): Owner of the file
```
Sami1309, line 159: specify which ones if any are optional here

client/src/featureform/register.py
```diff
@@ -153,6 +189,31 @@ def df_transformation(self,
                           name: str = "",
                           description: str = "",
                           inputs: list = []):
+        """
+        Register a Dataframe transformation source. The local.df_transformation decorator takes the contents
+        of the following function and executes the code it contains at serving time.
+
+        The name of the function is the name of the resulting source.
```
Sami1309, line 196: somewhat ambiguous, may imply reverse causality. Switch to mayb "the name of the function is passed to the resulting source as its own name" or something

client/src/featureform/register.py
```diff
@@ -153,6 +189,31 @@ def df_transformation(self,
...
+
+        The specified inputs are loaded into dataframes that can be accessed using the function parameters.
+
+        **Examples**:
+        ``` py
+        @local.df_transformation(inputs=[("source", "one"), ("source", "two")]) # Sources are added as inputs
+        def average_user_transaction(source_one, source_two):                   # Sources can be manipulated by adding them as params
```
Sami1309, line 203: this convention may imply that the name is generated from the snake case of the function, may be misleading

client/src/featureform/register.py
```diff
@@ -234,6 +331,33 @@ def register_resources(
...
+            ],
+        )
+        ```
+
+        Args:
+            entity (Union[str, EntityRegistrar]): The name to reference the entity by when serving features
+            entity_column (str): The name of the column in the source to be used as the entity
+            owner (Union[str, UserRegistrar]): The owner of the resource(s)
+            inference_store (Union[str, OnlineProvider]): Where to store the materialized feature for serving. (Use the local provider in Localmode)
+            features (List[ColumnMapping]): A list of column mappings to define the features
```
Sami1309, line 354: is it obvious what column mapping means generally?

client/src/featureform/register.py
```diff
@@ -1437,10 +1588,10 @@ def __init__(self, host=None, local=False, insecure=False, cert_path=None):
...
 
         Args:
-            host (str): Host path
-            local (bool): If localmode is being used
-            insecure (bool): If true, do not do TLS verification
-            cert_path (str): Path to certificate
+            host (str): The hostname of the Featureform instance. Exclude if using Localmode.
+            local (bool): True if using Localmode.
+            insecure (bool): True if connecting to an insecure Featureform endpoint. False if using a self-signed or public TLS certificate
+            cert_path (bool): The path to a public certificate if using a self-signed certificate.
```
Sami1309, line 1594: why is this a bool?

client/src/featureform/serving.py
```diff
@@ -18,23 +18,81 @@
...
+    ```
+    """
 
     def __init__(self, host=None, local=False, insecure=False, cert_path=None):
+        """
+        Args:
+            host (str): The hostname of the Featureform instance. Exclude if using Localmode.
+            local (bool): True if using Localmode.
+            insecure (bool): True if connecting to an insecure Featureform endpoint. False if using a self-signed or public TLS certificate
+            cert_path (bool): The path to a public certificate if using a self-signed certificate.
```
Sami1309, line 48: should be a str i think

Sami1309 (commented):
just minor things

client/src/featureform/register.py
sdreyer: Clearer?

client/src/featureform/register.py
sdreyer: this clearer?

client/src/featureform/register.py
sdreyer: Added a comment to the example above

client/src/featureform/register.py
sdreyer: fixed

client/src/featureform/serving.py
sdreyer: fixed

Sami1309 (approved):
lgtm

Issue #423, opened 2022-08-31
[Bug]: Inability to do time-based join, and potentially to name time column different than 'ts'


samuell:
### Expected Behavior

Showing some numerical values from features and labels in the console.

(Documented in detail in this repo, which contains a full bug reproduction setup: https://github.com/samuell/bugs/tree/main/20220831-ff-join-bug#readme)

### Actual Behavior

Got this stack trace:

```bash
$ python train.py 
Traceback (most recent call last):
  File "/home/sal/proj/sav/2022/bug-reproductions/20220831-ff-join-bug/train.py", line 5, in <module>
    train_data = client.training_set("traindata", "default")
  File "/home/sal/.cache/pypoetry/virtualenvs/20220831-ff-join-bug-JAwT4QYU-py3.9/lib/python3.9/site-packages/featureform/serving.py", line 49, in training_set
    return self._local_training_set(name, version)
  File "/home/sal/.cache/pypoetry/virtualenvs/20220831-ff-join-bug-JAwT4QYU-py3.9/lib/python3.9/site-packages/featureform/serving.py", line 112, in _local_training_set
    trainingset_df = pd.merge_asof(trainingset_df, df.sort_values(['ts']), direction='backward',
  File "/home/sal/.cache/pypoetry/virtualenvs/20220831-ff-join-bug-JAwT4QYU-py3.9/lib/python3.9/site-packages/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "/home/sal/.cache/pypoetry/virtualenvs/20220831-ff-join-bug-JAwT4QYU-py3.9/lib/python3.9/site-packages/pandas/core/frame.py", line 6259, in sort_values
    k = self._get_label_or_level_values(by, axis=axis)
  File "/home/sal/.cache/pypoetry/virtualenvs/20220831-ff-join-bug-JAwT4QYU-py3.9/lib/python3.9/site-packages/pandas/core/generic.py", line 1779, in _get_label_or_level_values
    raise KeyError(key)
KeyError: 'ts'
```

After trying to rename all the occurances of "ts" to "time" in `data.csv` and `defs.py`, and running
```
rm -rf .featureform && sleep 1 && featureform apply --local defs.py
python train.py
```
.. I instead got this:  

```bash
$ python train.py 
Traceback (most recent call last):
  File "/home/sal/proj/sav/2022/bug-reproductions/20220831-ff-join-bug/train.py", line 5, in <module>
[...]

PR #427, opened 2022-08-31
423 bug inability to do time based join and potentially to name time column different than ts

sdreyer:
# Description

Fixes an issue where the timestamp column 'ts' was hardcoded in the serving API. This has been swapped out with the correct timestamp variable

https://github.com/featureform/featureform/issues/423

### Select type(s) of change

- [ ] Documentation update

# Checklist:

- [-] I have made corresponding changes to the documentation

### DIFF
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -262,6 +262,7 @@ class LocalClientImpl:
                                            right_on=feature_row['source_timestamp'], left_by=label_row['source_entity'],
                                            right_by=feature_row['source_entity'])
             df.drop_duplicates(subset=[feature_row['source_entity']], keep="last", inplace=True)
+            df.drop(columns=feature_row['source_timestamp'], inplace=True)
         else:
             df.drop_duplicates(subset=[feature_row['source_entity']], keep="last", inplace=True)
             trainingset_df.reset_index(inplace=True)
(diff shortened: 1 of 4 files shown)


client/src/featureform/serving.py
```diff
@@ -257,11 +257,12 @@ def feature_df_from_csv(self, feature, filename):
 
     def merge_feature_into_ts(self, feature_row, label_row, df, trainingset_df):
         if feature_row['source_timestamp'] != "":
-            trainingset_df = pd.merge_asof(trainingset_df, df.sort_values(['ts']), direction='backward',
+            trainingset_df = pd.merge_asof(trainingset_df, df.sort_values(feature_row['source_timestamp']), direction='backward',
                                            left_on=label_row['source_timestamp'],
                                            right_on=feature_row['source_timestamp'], left_by=label_row['source_entity'],
                                            right_by=feature_row['source_entity'])
-            df.drop_duplicates(subset=[feature_row['source_entity']], keep="last", inplace=True)
+            if feature_row['source_timestamp'] != label_row['source_timestamp']:
```
Sami1309, line 264: so this drops a feature's timestamp column if they do not match the label's timestamp? Is this because only the label timestamp is relevant?

sdreyer: It iterates through all the features and does point in time matching with each one. Sometimes they have the same timestamp column name, so we don't want to drop since it'll drop the labels timestamp. But if it has a different timestamp we drop it because its not useful anymore

Sami1309: ok makes sense 👍

Sami1309 (approved):
lgtm

PR #433, opened 2022-09-02
Fix user error

ksshiraja:
# Description

The changes now doesn't require a user to have registered feature variants.

The dashboard user page was return a 500 error because a registered user did not have feature resources. And the sqlite code required the retrieved list length to be greater than 0

## Type of change
Bug Fix

### Select type(s) of change

- [ ] Documentation update

# Checklist:

- [x ] I have performed a self-review of my code
- [ x] I have made corresponding changes to the documentation
- [ x] My changes generate no new warnings
- [ x] New and existing unit tests pass locally with my changes
- [ x] I have fixed any merge conflicts

### DIFF
--- a/client/src/featureform/sqlite_metadata.py
+++ b/client/src/featureform/sqlite_metadata.py
@@ -193,7 +193,7 @@ class SQLiteMetadata:
         variant_data = self.__conn.execute(query)
         self.__conn.commit()
         variant_data_list = variant_data.fetchall()
-        if len(variant_data_list) == 0:
+        if len(variant_data_list) == 0 and column != "owner":
           raise ValueError(f"{type} with {column}: {resource} not found")
         return variant_data_list
   


client/src/featureform/sqlite_metadata.py
```diff
@@ -193,7 +193,7 @@ def query_resource(self, type, column, resource):
         variant_data = self.__conn.execute(query)
         self.__conn.commit()
         variant_data_list = variant_data.fetchall()
-        if len(variant_data_list) == 0:
+        if len(variant_data_list) == 0 and column != "owner":
```
sdreyer, line 196: Shouldnt there be the default user though?

ksshiraja: This checks the owner column in resource metadata tables. For example if a user is registered but doesn't create any resources the query_resource will return an empty list. So in the feature_variant table there are chances there is no feature owned by ____.

sdreyer (approved):

PR #436, opened 2022-09-02
add source mapping condition for spark transformation

Sami1309:
# Description

Currently the transformation definition is passed to the coordinator like so:

`SELECT * FROM {{name.variant}}`

before spark, the coordinator would look up the SQL table name for each name.variant in the transformation and replace them, in the coordinator, with their SQL table name. So the above becomes, in the coordinator

`SELECT * FROM featureform_primary_name_variant`

Originally it was simple to convert the {{name.variant}} to `featureform_primary_name_variant` since this mapping was the same for every provider. There was a minor issue in this case in Bigquery where the table names needed an extra prefix that Ahmad was able to solve.

However for Spark, the table names are stored as S3 file paths. While the prefix of the file path is deterministic based on the resourceID, there is also a random file name that the ResourceKey method in spark.go resolves.

spark.go can't use just a formatted SQL query, it needs the query and a separate source list. However, it is still being passed this:

`SELECT * FROM featureform_primary_name_variant`

And since spark stores primary tables and transformations in a different way, if passed this, the spark store has no idea what to do at this point, we need to distinguish primary tables and transformation tables and let the replacement function reverse the templating in the coordinator

`SELECT * FROM featureform_primary_name_variant`

Source mapping: featureform_primary_name_variant: featureform_primary_name_variant

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/resources.py
+++ b/client/src/featureform/resources.py
@@ -273,7 +273,7 @@ class SparkAWSConfig:
         return "spark"
 
     def type(self) -> str:
-        return "SPARK_AWS_OFFLINE"
+        return "SPARK_OFFLINE"
 
     def serialize(self) -> bytes:
         config = {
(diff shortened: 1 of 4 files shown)


ahmadnazeri (approved):
LGTM

PR #454, opened 2022-09-13
Unimpliment coalesce from spark script and make necessary changes to spark runner

Sami1309:
# Description

Removes the .coalesce option called when saving resources in the spark script. This saves the resources in multiple parts instead of conjoining them into a single part. Implementing this requires changing how iterators parse through files and the functional definition of a resource in the offline store (so now a resource is defined as the parent directory containing all the parts, not the single parquet file)

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/spark.go
```diff
@@ -783,7 +828,13 @@ func (s *S3PrimaryTable) GetName() string {
 }
 
 func (s *S3PrimaryTable) IterateSegment(n int64) (GenericTableIterator, error) {
-	columnStream, err := s.store.ResourceStreamConv(s.sourcePath, 0)
+	var columnStream chan interface{}
+	var err error
+	if s.isTransformation {
```
sdreyer, line 833: Why do we care if its a transformation or not?

provider/spark.go
```diff
@@ -250,8 +250,11 @@ type SparkExecutor interface {
 type SparkStore interface {
 	UploadSparkScript() error //initialization function
 	ResourceKey(id ResourceID) (string, error)
+	ResourceKeys(id ResourceID) ([]string, error)
 	ResourceStreamConv(key string, begin int64) (chan interface{}, error)
-	ResourceRowCt(key string) (int, error)
+	ResourceMultiPartStream(id ResourceID) (chan interface{}, error)
+	ResourceRowCt(id ResourceID) (int, error)
```
sdreyer, line 256: Can you just rename these to Count instead of Ct?

provider/spark.go
```diff
@@ -250,8 +250,11 @@ type SparkExecutor interface {
 type SparkStore interface {
 	UploadSparkScript() error //initialization function
 	ResourceKey(id ResourceID) (string, error)
+	ResourceKeys(id ResourceID) ([]string, error)
 	ResourceStreamConv(key string, begin int64) (chan interface{}, error)
-	ResourceRowCt(key string) (int, error)
+	ResourceMultiPartStream(id ResourceID) (chan interface{}, error)
```
sdreyer, line 255: This interface is really confusing, but refactoring is out of scope for this. Will create a task to clean this up after so we can get this working

provider/spark.go
```diff
@@ -1119,18 +1177,21 @@ func (s *S3Materialization) ID() MaterializationID {
...
 	return int64(numRows), nil
 }
 
 func (s *S3Materialization) IterateSegment(begin, end int64) (FeatureIterator, error) {
-	stream, err := s.store.ResourceStreamConv(s.Key, begin)
+	stream, err := s.store.ResourceMultiPartStream(s.id)
 	if err != nil {
 		return nil, err
 	}
+	for i := int64(0); i < begin; i++ {
```
sdreyer, line 1192: Wasn't there a function that does this?

provider/spark.go
```diff
@@ -356,6 +359,35 @@ func ResourcePrefix(id ResourceID) string {
 	return fmt.Sprintf("featureform/%s/%s/%s/", id.Type, id.Name, id.Variant)
 }
 
+func (s *S3Store) ResourceKeys(id ResourceID) ([]string, error) {
```
sdreyer, line 362: What does this mean?

provider/spark.go
```diff
@@ -356,6 +359,35 @@ func ResourcePrefix(id ResourceID) string {
...
+		s.logger.Errorw("Failure retrieving objects from S3 Store", err)
+		return nil, err
+	}
+	if len(objects.Contents) == 0 {
+		s.logger.Errorw("no objects with key prefix exist in S3", latestFilePrefix)
+		return []string{}, nil
+	}
+	s.logger.Debugf("Found %d objects", len(objects.Contents))
+	returnList := make([]string, 0, len(objects.Contents)-1)
+	for _, object := range objects.Contents {
```
sdreyer, line 382: Unclear what this is doing

provider/spark.go
Sami1309: Added an implementation for multiple files

sdreyer:
@Sami1309 can you run this with the tests @ahmadnazeri wrote and verify those still work?

sdreyer (approved):

PR #456, opened 2022-09-14
Ability to use a URL in `featureform apply`

sdreyer:
# Description

Allows both URLs and files to be used in `featureform apply` to allow us to host definition files for faster demos.

No

### Select type(s) of change

- [x] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -1796,7 +1796,7 @@ class ResourceClient(Registrar):
     ```
     """
 
-    def __init__(self, host=None, local=False, insecure=False, cert_path=None):
+    def __init__(self, host=None, local=False, insecure=False, cert_path=None, dry_run=False):
         """Initialise a Resource Client object.
 
         Args:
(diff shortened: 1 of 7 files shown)


client/src/featureform/register.py
```diff
@@ -1796,7 +1796,7 @@ class ResourceClient(Registrar):
     ```
     """
 
-    def __init__(self, host=None, local=False, insecure=False, cert_path=None):
+    def __init__(self, host=None, local=False, insecure=False, cert_path=None, dry_run=False):
```
Sami1309, line 1799: What does dry run do? Print everything without applying? Is it just used for testing?

sdreyer: Yeah, it should be replaced in the future with a "plan" argument, but this is just so we can check the inputs without it trying to run anything. It can be used outside of testing too

Sami1309 (approved):
lgtm

PR #457, opened 2022-09-14
Featureform in a single container

sdreyer:
# Description

This pull request adds featureform into a single container using nginx and supervisord. 
The goal is to decrease the time to run featureform locally

No

### Select type(s) of change

- [ ] Documentation update

# Checklist:


sdreyer:
@Sami1309 if you have time to look over this again

newserving/Dockerfile
```diff
@@ -29,5 +29,6 @@ FROM alpine
 COPY --from=0 ./app/main ./main
 COPY --from=0 ./app/provider/scripts/offline_store_spark_runner.py /app/provider/scripts/offline_store_spark_runner.py
 
+ENV SERVING_PORT "0.0.0.0:8080"
```
Sami1309, line 32: would this make address 0.0.0.0:0.0.0.0:8080 then?

metadata/server/server.go
```diff
@@ -6,38 +6,49 @@ package main
...
-		Logger:  logger,
-		Address: fmt.Sprintf(":%s", addr),
-		TypeSenseParams: &search.TypeSenseParams{
-			Port:   help.GetEnv("TYPESENSE_PORT", "8108"),
-			Host:   help.GetEnv("TYPESENSE_HOST", "localhost"),
-			ApiKey: help.GetEnv("TYPESENSE_APIKEY", "xyz"),
-		},
-		StorageProvider: storageProvider,
+	var config *metadata.Config
+	if disableTypesense == "false" {
```
Sami1309, line 32: config = &metadata.Config{
			Logger:          logger,
			Address:         fmt.Sprintf(":%s", addr),
			StorageProvider: storageProvider,
		}
repeats in either block. Should it be declared before and then TypeSennseParams can just be added in a single if block after to the config?

charts/featureform/charts/feature-server/templates/deployment.yaml
```diff
@@ -29,7 +29,7 @@ spec:
             - containerPort: {{ .Values.metrics.port }}
           env:
             - name: SERVING_PORT
-              value: 0.0.0.0:{{ .Values.serving.port }}
+              value: "{{ .Values.serving.port }}"
```
Sami1309, line 32: why make this a different convention than metadata port?

Sami1309 (commented):
lgtm, just minor comments

charts/featureform/charts/feature-server/templates/deployment.yaml
sdreyer: changed to be consistent

metadata/server/server.go
sdreyer: removed duplicate code

newserving/Dockerfile
sdreyer: fixed

PR #468, opened 2022-09-22
add implementation for training sets and tests

Sami1309:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -523,7 +523,7 @@ class Dataset:
         stream = Stream(self._stream, name, version)
         return Dataset(stream)
     
-    def from_dataframe(dataframe):
+    def from_dataframe(self, dataframe):
         stream = LocalStream(dataframe.values.tolist())
         return Dataset(stream, dataframe)
 
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -586,7 +586,7 @@ class Dataset:
                     second_index = random.randint(0, len(self._dataframe)-1)
                 indices.add(first_index)
                 indices.add(second_index)
-                temp = df.iloc[first_index].copy()
+                temp = self._dataframe.iloc[first_index].copy()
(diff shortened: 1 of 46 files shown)


client/src/featureform/serving.py
```diff
@@ -567,6 +576,8 @@ def shuffle(self, buffer_size):
         if buffer_size <= 0:
             raise Exception("Buffer size must be greater than or equal to 1")
         self._stream = Shuffle(buffer_size, self._stream)
+        if self._dataframe is not None:
+            self._dataframe = self._dataframe.sample(frac=1)
```
sdreyer, line 580: We dont want this to always shuffle all the rows

Sami1309: Should this replicate the way the normal stream shuffles from a buffer?

sdreyer: The stream shuffles as many rows as specified in the buffer_size I believe. This should act similarly

Sami1309: Shuffle technically shuffles every row, the buffer is the "memory" for how many rows it keeps in a pool to randomly select from. The shuffling logic though applies to every row.  I think this was done to avoid needing to load every row before shuffling, but since every row is read ahead of time for the dataframe it wouldn't be necessary.

To replicate the logic exactly I could implement it like this:
buffer = pd.DataFrame()
shuffled_df = pd.DataFrame()
for i, row in df.iterrows():
    if i < buffer_size:
        buffer = buffer.append(row)
    else:
        random_index = random.randrange(len(buffer))
        suffled_df = shuffled_df.append(buffer[random_index])
buffer = buffer.sample(frac=1)
for row in buffer.iterrows():
    shuffled_df = shuffled_df.append(row)
df = shuffled_df

sdreyer: Gotcha, it doesn't need the exact logic, just the same outcome. If this satisfies that then its fine

sdreyer (approved):

PR #481, opened 2022-10-11
Feature/include label timestamp for training set

imanthorpe:
# Description

The goal of this feature is to give users the ability to include the label timestamp in the training set. Currently, the label does not include any timestamps when being served in the training set. The expected behavior is that we exclude the timestamp by default but if users provide a flag then we will include it.

### Select type(s) of change

- [ ] Documentation update

# Checklist:


ahmadnazeri (approved):
LGTM

PR #494, opened 2022-10-12
implement blob online store for azure blob store

Sami1309:
# Description

Add functionality so an azure blob container can be registered as an online store. Also seed functionality for it to be used as a parameter for separated compute offline stores (via spark or k8s)

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -332,6 +332,21 @@ class OnlineProvider:
     def name(self) -> str:
         return self.__provider.name
 
+class StoreProvider:
+    def __init__(self, registrary, provider, config, store_type):
+        self.__registrar = registrar
+        self.__provider = provider
+        self.__config = config
+        self.__store_type = store_type
+    
+    def name(self) -> str:
+        return self.__provider.name
+    
+    def store_type(self) -> str:
+        return self.__store_type
+
+    def config(self):
+        return self.__config
 
 class LocalProvider:
     """
@@ -543,7 +558,7 @@ class LocalSource:
             entity: Union[str, EntityRegistrar],
             entity_column: str,
             owner: Union[str, UserRegistrar] = "",
-            inference_store: Union[str, OnlineProvider] = "",
+            inference_store: Union[str, OnlineProvider, BlobProvider] = "",
             features: List[ColumnMapping] = None,
             labels: List[ColumnMapping] = None,
             timestamp_column: str = ""
@@ -567,7 +582,7 @@ class LocalSource:
             entity (Union[str, EntityRegistrar]): The name to reference the entity by when serving features
             entity_column (str): The name of the column in the source to be used as the entity
             owner (Union[str, UserRegistrar]): The owner of the resource(s)
-            inference_store (Union[str, OnlineProvider]): Where to store the materialized feature for serving. (Use the local provider in Localmode)
+            inference_store (Union[str, OnlineProvider, BlobProvider]): Where to store the materialized feature for serving. (Use the local provider in Localmode)
             features (List[ColumnMapping]): A list of column mappings to define the features
             labels (List[ColumnMapping]): A list of column mappings to define the labels
             timestamp_column: (str): The name of an optional timestamp column in the dataset. Will be used to match the features and labels with point-in-time correctness
@@ -641,7 +656,7 @@ class SQLTransformationDecorator:
             entity: Union[str, EntityRegistrar],
             entity_column: str,
             owner: Union[str, UserRegistrar] = "",
-            inference_store: Union[str, OnlineProvider] = "",
+            inference_store: Union[str, OnlineProvider, BlobProvider] = "",
             features: List[ColumnMapping] = None,
             labels: List[ColumnMapping] = None,
             timestamp_column: str = "",
@@ -712,7 +727,7 @@ class DFTransformationDecorator:
             entity: Union[str, EntityRegistrar],
             entity_column: str,
             owner: Union[str, UserRegistrar] = "",
-            inference_store: Union[str, OnlineProvider] = "",
+            inference_store: Union[str, OnlineProvider, BlobProvider] = "",
             features: List[ColumnMapping] = None,
             labels: List[ColumnMapping] = None,
             timestamp_column: str = "",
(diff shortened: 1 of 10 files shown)


client/src/featureform/register.py
```diff
@@ -983,6 +998,45 @@ def get_redis(self, name):
...
+        ```
+        Args:
+            name (str): Name of Azure blob provider to be retrieved
+
+        Returns:
+            azure_blob (StoreProvider): Provider
+            bm22
+        """
+
+        # azure_config = AzureBlobConfig(account_name=account_name, account_key=account_key,container_name=container_name,root_path=root_path)
```
ahmadnazeri, line 1025: do we want to remove this commented out code?

client/src/featureform/register.py
```diff
@@ -2999,6 +3102,7 @@ def list_providers(self, local=False):
 clear_state = global_registrar.clear_state
 register_user = global_registrar.register_user
 register_redis = global_registrar.register_redis
+register_blob_store = global_registrar.register_blob_store
```
ahmadnazeri, line 3105: do we need to add `get_azure_blob` here as well? And should both functions be named similarly?

client/src/featureform/resources_test.py
```diff
@@ -40,6 +40,30 @@ def redis_config():
         db=3,
     )
 
+@pytest.fixture
+def online_blob_config(
```
ahmadnazeri, line 44: with fixtures, you can pass in another fixture to it. Basically, define the `azure_blob_store_config` fixture first then for this one, just pass it in as a parameter. 

```
@pytest.fixture
def azure_blob_store_config():
      return AzureBlobStoreConfig(
         account_name="<account_name>",
         account_key="<account_key>",
         container_name="examplecontainer",
         root_path="example/path",
     )

@pytest.fixture
def online_blob_config(azure_blob_store_config):
     return OnlineBlobConfig(
         store_type="AZURE",
         store_config=azure_blob.serialize(),
     )
```

client/src/featureform/register.py
```diff
@@ -332,6 +332,21 @@ def __init__(self, registrar, provider):
     def name(self) -> str:
         return self.__provider.name
 
+class StoreProvider:
```
sdreyer, line 335: Do we want this as StoreProvider? Maybe FilestoreProvider instead for clarity

client/src/featureform/register.py
```diff
@@ -983,6 +998,45 @@ def get_redis(self, name):
         fakeProvider = Provider(name=name, function="ONLINE", description="", team="", config=fakeConfig)
         return OnlineProvider(self, fakeProvider)
 
+    def get_azure_blob(self, name):
```
sdreyer, line 1001: Should this be get_blob_store instead so it matches the register_blob_store call?

client/src/featureform/resources.py
```diff
@@ -80,6 +80,43 @@ def serialize(self) -> bytes:
         }
         return bytes(json.dumps(config), "utf-8")
 
+@typechecked
+@dataclass
+class AzureBlobStoreConfig:
```
sdreyer, line 85: Should AzureBlobStoreConfig and OnlineBlobConfig be the same?

provider/blob_online.go
```diff
@@ -0,0 +1,204 @@
...
+func blobOnlineStoreFactory(serialized SerializedConfig) (Provider, error) {
+	onlineBlobConfig := &OnlineBlobConfig{}
+	if err := onlineBlobConfig.Deserialize(serialized); err != nil {
+		return nil, err
+	}
+	return NewOnlineBlobStore(onlineBlobConfig)
+}
+
+func NewOnlineBlobStore(config *OnlineBlobConfig) (*OnlineBlobStore, error) {
+	blobStore, err := CreateBlobStore(string(config.Type), Config(config.Config))
```
sdreyer, line 45: This handles if we were to use S3 or something else correct?

provider/blob_online.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+	}, nil
+}
+
+func (store *OnlineBlobStore) AsOnlineStore() (OnlineStore, error) {
+	return store, nil
+}
+
+func blobTableKey(feature, variant string) string {
+	return fmt.Sprintf("tables/%s/%s", feature, variant)
```
sdreyer, line 63: Are normal files and featureform files differentiated with this key? Just so people know which ones we own

provider/blob_online.go
```diff
@@ -0,0 +1,204 @@
...
+func (store OnlineBlobStore) DeleteTable(feature, variant string) error {
+	exists, err := store.tableExists(feature, variant)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return &TableNotFound{feature, variant}
+	}
+	return store.deleteTable(feature, variant)
+	//TODO should this also cycle through all values and delete them too?
```
sdreyer, line 135: What values is this referring to? In the file?

client/src/featureform/resources.py
Sami1309: onlineblobconfig is the config for registering an offline store using a variable blob-compatible store. OnlineBlobConfig could also have an s3 blob or other cloud provider compatible with the backend blob api as a parameter.

client/src/featureform/register.py
Sami1309: yes

client/src/featureform/register.py
Sami1309: good idea. I was thinking about the right term. It's something that offers offline and online storage. Perhaps hybrid store?

provider/blob_online.go
Sami1309: yeah i should prepend .featureform to both this and the entity keys.

provider/blob_online.go
Sami1309: yes. The "index" for all the tables is in /tables while the actual values for the entities are in /values. I can delete the tables but the values persist. Really it makes sense to get rid of everything so I'll add a directory delete as par tof the delete table

sdreyer: Yeah makes sense

sdreyer (approved):

PR #512, opened 2022-10-24
Adding the client changes for lag features

ahmadnazeri:
# Description
Adding the client and protobuf changes

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+from multiprocessing.sharedctypes import Value
 from typeguard import typechecked, check_type
 from typing import Tuple, Callable, List, Union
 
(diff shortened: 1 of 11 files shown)


client/src/featureform/register.py
```diff
@@ -1967,25 +1970,47 @@ def register_column_resources(
...
+                    feature_nv_list.append(feature_nv)
             elif isinstance(feature, list):
                 feature_nv_list.extend(self.__get_feature_nv(feature))
             else:
                 feature_nv_list.append(feature)
-        return feature_nv_list
+
+        unique_feature_nv_list = list(set(feature_nv_list))
+        unique_feature_lags = list(set(feature_lags))
+        return unique_feature_nv_list, unique_feature_lags
```
Sami1309, line 2006: Is this actually needed? If we register feature lags as part of a training set at this point, they don't actually have any definition except their derivation from the original features. Could they just remain in the training set definition, untouched until they get to the coordiantor?

Sami1309 (approved):
lgtm, though one comment about some change, but if it works we can address it later.

PR #515, opened 2022-10-26
Bugfix/parquet iterator nil value

sdreyer:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/provider/k8s.go
+++ b/provider/k8s.go
@@ -484,7 +484,7 @@ func (store genericBlobStore) Serve(key string) (Iterator, error) {
 	keyParts := strings.Split(key, ".")
 	switch fileType := keyParts[len(keyParts)-1]; fileType {
 	case "parquet":
-		return parquetIteratorFromReader(b)
+		return parquetIteratorFromBytes(b)
 	default:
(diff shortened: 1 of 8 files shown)


provider/k8s.go
```diff
@@ -474,80 +477,64 @@ func (store genericBlobStore) Read(key string) ([]byte, error) {
 }
 
 func (store genericBlobStore) Serve(key string) (Iterator, error) {
-	r, err := store.bucket.NewReader(ctx, key, nil)
+	b, err := store.bucket.ReadAll(ctx, key)
```
Sami1309, line 480: For 1M row files I assume this is a bottleneck, as it would have been in the original version too.  Should the server in the future read chunks of bytes at a time so it doesn't need to read everything before writing?

Sami1309 (approved):
Good solve on this, lgtm just one comment on a possible performance issue.

provider/k8s.go
sdreyer: The other implementation did this too. So far theres no issues with it

PR #522, opened 2022-11-03
Adding the azure blob store support for pyspark script 

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -111,13 +129,73 @@ def get_code_from_file(file_path, aws_region=None):
...
+
+    for key, value in configs.items():
+        spark.conf.set(key, value)
+
+
+def split_key_value(values):
+    arguments = {}
+    for value in values:
+        # split it into key and value; parse the first value
+        key, value = value.split('=', 1)
```
Sami1309, line 193: Using the = sign as a delimeter may be a problem, the azure key we're using ends in two ==

Sami1309 (approved):
lgtm

provider/scripts/spark/offline_store_spark_runner.py
ahmadnazeri: so we are splitting on the first `=`. We should be good as long as the keys don't have `=` in them. The reason I did it this way is because it is the expected norm. But can modify it if we want.

Sami1309: Yeah I actually meant to delete this comment as I saw you're only doing the first, so it shouldn't be an issue

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -8,47 +9,58 @@
...
 
 
+real_path = os.path.realpath(__file__)
+dir_path = os.path.dirname(real_path)
+LOCAL_DATA_PATH = f"{dir_path}/.featureform/data"
+
 def main(args):
     if args.transformation_type == "sql": 
-        output_location = execute_sql_query(args.job_type, args.output_uri, args.sql_query, args.source_list)
+        output_location = execute_sql_query(args.job_type, args.output_uri, args.sql_query, args.spark_config, args.source_list)
```
sdreyer, line 21: Does the offline store know to pass these variables in?

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -8,47 +9,58 @@
...
     elif args.transformation_type == "df":
-        output_location = execute_df_job(args.output_uri, args.code, args.aws_region, args.source)
+        output_location = execute_df_job(args.output_uri, args.code, args.store_type, args.spark_config, args.credential, args.source)
```
sdreyer, line 23: Was this change made in the offline store?

PR #544, opened 2022-11-17
Fix for invalid onlinestore casting

sdreyer:
# Description

Type casting for the online store was not standardized. This resulted in:
1. Invalid types being passed from the client to the online store
2. The online store returning nil values in most providers when not able to deserialize values

Solutions implemented:
1. Check on the client side at `apply` time to verify types are valid
2. Checks on the online store side to verify types are valid
3. Default cases for casts and errors if the casts fail, instead of returning nil

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/offline.go
```diff
@@ -37,9 +37,36 @@ const (
...
+	// Check if value type is a valid type
+	switch vt {
+	case NilType, Int, Int32, Int64, Float32, Float64, String, Bool, Timestamp, Datetime:
+		return nil
+	default:
+		return fmt.Errorf("invalid table type: %v", vt)
+	}
+}
+
+func (vt ValueType) doesMatch(value interface{}) bool {
```
sdreyer, line 54: Forces exact types to be used because of discrepancies in ability to cast with different providers

PR #549, opened 2022-11-21
[Bugfix] Redis invalid reads

sdreyer:
# Description

Fixes a bug with redis where a feature column type could be any string. If not a type recognized by redis, it would return an empty string.
This PR implements two changes:
1. Checking the types in the client to validate that they match with known types on the online store
2. Making unknown types in redis default to String to not break already created tables

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/redis.go
```diff
@@ -117,9 +121,13 @@ func (table redisOnlineTable) Get(entity string) (interface{}, error) {
 		result, err = val.Float64()
```
Sami1309, line 121: should it return float32 when it's a float 32 rather than float64?

Sami1309 (approved):
lgtm just one small comment, not on commit actually

provider/redis.go
sdreyer: Unsure what you mean. This case is Float64 returning Float64

PR #570, opened 2022-12-08
Fixing Support for EMR

ahmadnazeri:
# Description
Goal is to fix the support for EMR and Databricks. 

### Select type(s) of change

- [ ] Documentation update

# Checklist:


client/src/featureform/register.py
```diff
@@ -20,14 +20,18 @@
...
     SQLTransformation, DFTransformation, Entity, Feature, Label, ResourceColumnMapping, TrainingSet, ProviderReference, \
-    EntityReference, SourceReference, ExecutorCredentials, ResourceRedefinedError, ResourceStatus, Transformation
+    EntityReference, SourceReference, ExecutorCredentials, ResourceRedefinedError, ResourceStatus, Transformation, AWSCredentials
 
 from .proto import metadata_pb2_grpc as ff_grpc
 
+
 NameVariant = Tuple[str, str]
 
+s3_config = S3StoreConfig("", "", AWSCredentials("id", "secret"))
```
sdreyer, line 32: Is there a way to do this without having to create the actual object?

client/src/featureform/resources.py
```diff
@@ -104,6 +104,29 @@ def serialize(self) -> bytes:
...
 
 
+@typechecked
+@dataclass
+class AWSCredentials:
+    def __init__(self,
+                 aws_access_key_id: str = "",
+                 aws_secret_access_key: str = "",):
+        empty_strings = aws_access_key_id == "" or aws_secret_access_key == ""
+        assert not empty_strings, Exception(
```
sdreyer, line 114: Are asserts typically used in non-testing situations?

client/src/featureform/register.py
ahmadnazeri: We can hardcode the types but we will have to modify it if we modify the type.

client/src/featureform/resources.py
ahmadnazeri: Modified it to exceptions.

client/src/featureform/register.py
sdreyer: Does it make sense to make all the types an enum?

client/src/featureform/register.py
```diff
@@ -3681,4 +3743,5 @@ def list_providers(self, local=False):
 get_spark_aws = global_registrar.get_spark
```
sdreyer, line 3743: This should just be "get_spark"

client/tests/test_spark_provider.py
```diff
@@ -6,20 +6,30 @@
...
+    executor = request.getfixturevalue(executor_fixture)
+    filestore = request.getfixturevalue(filestore_fixture)
 
-    databricks = DatabricksCredentials(username="a", password="b", cluster_id="c_id")
-    azure_blob = AzureFileStoreConfig(account_name="", account_key="", container_name="", root_path="")   
+    provider_name = "test_offline_spark_provider"
+    r = Registrar() 
     
-    config = SparkConfig(executor_type=databricks.type(), executor_config=databricks.config(), store_type=azure_blob.store_type(), store_config=azure_blob.config())
+    config = SparkConfig(executor_type=executor.type(), executor_config=executor.config(), store_type=filestore.store_type(), store_config=filestore.config())
```
sdreyer, line 25: Is there a check that register_spark() will typecheck/accept these types?

ahmadnazeri: The `register_spark` method requires a `executor: ExecutorCredentials` and `filestore: FileStoreProvider`. That's the check on the type.

sdreyer: Ahh the register functions return those. All good

PR #587, opened 2023-01-09
adding filestore upload and download

ahmadnazeri:
# Description
Modified the filestore interface to upload and download files. 

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/provider/k8s.go
+++ b/provider/k8s.go
@@ -5,6 +5,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
 	dp "github.com/novln/docker-parser"
 	"github.com/segmentio/parquet-go"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


provider/k8s.go
```diff
@@ -633,6 +637,37 @@ func (store genericFileStore) ServeDirectory(dir string) (Iterator, error) {
 	return parquetIteratorOverMultipleFiles(fileParts, store)
 }
 
+func (store genericFileStore) Upload(sourcePath string, destPath string) error {
+	content, err := ioutil.ReadFile(sourcePath)
+	if err != nil {
+		return fmt.Errorf("cannot read %s file because %v", sourcePath, err)
```
sdreyer, line 643: typically the convention in go is to have <outer error>: <inner error>
`Errorf("cannot read file %s: %v"), sourcePath, err`

provider/k8s.go
```diff
@@ -633,6 +637,37 @@ func (store genericFileStore) ServeDirectory(dir string) (Iterator, error) {
 	return parquetIteratorOverMultipleFiles(fileParts, store)
 }
 
+func (store genericFileStore) Upload(sourcePath string, destPath string) error {
```
sdreyer, line 640: Since the filestore already takes a container path as an argument, does it make sense to chance `destPath` to `filename` or allow it to be a subpath?

ahmadnazeri: I think the subpath makes sense because you can store it in different locations especially if later on, we support manual backups.

sdreyer (approved):

PR #590, opened 2023-01-11
Adding Backup Script

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	err = backupProvider.Upload(name, name)
+	if err != nil {
+		return fmt.Errorf("cannot upload snapshot to filestore: %v", err)
+	}
+	return nil
+}
+
+func (b *Backup) Restore() error {
+	return nil
```
simba-git, line 53: Not implemented yet on purpose yeah? Random note but I like to throw in a panic instead of a return nil if I don't implement something to avoid forgetting. But I am very forgetful

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (b *Backup) getBackupProvider() (Provider, error) {
+	var backupProvider Provider
+	switch b.ProviderType {
+	case AZURE:
+		backupProvider = &Azure{
+			AzureStorageAccount: help.GetEnv("AZURE_STORAGE_ACCOUNT", ""),
```
simba-git, line 61: I don't like using env variables when its not easy to override them in the code. It almost always is the wrong form and makes testing weird.

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// Client Allows ETCD to be tested with mock values
+type Client interface {
+	Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.GetResponse, error)
+}
+
+type Backup struct {
+	ETCDClient   Client
+	ProviderType ProviderType
```
simba-git, line 28: turn this into Provider Provider

The Init method of both could do the env variable fill in if needed.

backup/backup.go
simba-git: Move this code into the Init methods

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	"encoding/json"
+	"fmt"
+	help "github.com/featureform/helpers"
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"io/ioutil"
+	"time"
+)
+
+type ProviderType string
```
simba-git, line 14: Delete this, more info below

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	Value []byte
+}
+
+func (b *Backup) takeSnapshot(filename string) error {
+	resp, err := b.ETCDClient.Get(context.Background(), "", clientv3.WithPrefix())
+	if err != nil {
+		return fmt.Errorf("could get snapshot values: %v", err)
+	}
+
+	values := b.convertValues(resp.Kvs)
```
simba-git, line 88: convertEtcdToBackupFile

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+			Path: "file://./",
+		}
+	default:
+		return nil, fmt.Errorf("the cloud provider '%s' is not supported", b.ProviderType)
+	}
+
+	return backupProvider, nil
+}
+
+type BackupFile struct {
```
simba-git, line 77: type BackupRow struct

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	Key   []byte
+	Value []byte
+}
+
```
simba-git, line 81: type BackupFile []BackupRow

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	for _, row := range resp {
+		values = append(values, BackupFile{
+			Key:   row.Key,
+			Value: row.Value,
+		})
+	}
+	return values
+}
+
+func (b *Backup) writeValues(filename string, values []BackupFile) error {
```
simba-git, line 107: func (b * BackupFile) WriteTo(filename string) error

Make it a method of BackupFile, makes it easier to test.

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	AZURE ProviderType = "AZURE"
+	LOCAL              = "LOCAL"
+)
+
+// Client Allows ETCD to be tested with mock values
+type Client interface {
+	Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.GetResponse, error)
+}
+
+type Backup struct {
```
simba-git, line 26: Maybe called this BackupManager or Backuper or something, it's not really a backup, the backup is the thing you're saving not this struct.

backup/backup.go
simba-git: or
type Backup []BackupRow

If you rename Backup to BackupManager

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+
+	values := b.convertValues(resp.Kvs)
+	err = b.writeValues(filename, values)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (b *Backup) convertValues(resp []*mvccpb.KeyValue) []BackupFile {
```
simba-git, line 96: convertEtcdToBackupFile

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	values := []BackupFile{}
```
simba-git, line 97: minor but you can do:
values := make([]BackupFile, len(resp))
for i, row := range resp {
  values[i] = backupRow{row.Key, row.Value}
}

backup/backup.go
simba-git: Actually this should be lower case, it's only for internal use in this package. Only make capital what you expect to be used in other pacakges. By default, lower case.

backup/backup.go
simba-git: actually:
backupRow not BackupRow

backup/backup.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if err = ioutil.WriteFile(filename, file, 0644); err != nil {
+		return fmt.Errorf("could not write snapshot to file: %v", err)
+	}
+	return nil
+}
+
+func GenerateSnapshotName(currentTime time.Time) string {
+	prefix := "featureform_etcd_snapshot"
+	formattedTime := currentTime.Format("2006-01-02 15:04:05")
```
simba-git, line 120: Minor but can we put a . or underscore between the data and time, I don't like spaces in file names.

backup/main/main.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"time"
+
+	"github.com/featureform/backup"
+	help "github.com/featureform/helpers"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+func main() {
+	etcdHost := help.GetEnv("ETCD_HOSTNAME", "localhost")
```
simba-git, line 13: Not worth changing, but I like allowing people to pass args into it if it's gonna work like a CLI tool instead of setting a ton of env variables.

backup/main/main.go
```diff
@@ -0,0 +1,45 @@
...
+	etcdPort := help.GetEnv("ETCD_PORT", "2379")
+	etcdUsername := help.GetEnv("ETCD_USERNAME", "")
+	etcdPassword := help.GetEnv("ETCD_PASSWORD", "")
+
+	address := fmt.Sprintf("%s:%s", etcdHost, etcdPort)
+
+	etcdConfig := clientv3.Config{
+		Endpoints:   []string{address},
+		DialTimeout: time.Second * 1,
```
simba-git, line 22: Long enough?

backup/main/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+	backupExecutor := backup.Backup{
+		ETCDClient:   client,
+		ProviderType: backup.ProviderType(help.GetEnv("CLOUD_PROVIDER", "LOCAL")),
+	}
+
+	currentTimestamp := time.Now()
+	snapshotName := backup.GenerateSnapshotName(currentTimestamp)
+
+	err = backupExecutor.Save(snapshotName)
```
simba-git, line 40: Why not have a Save method that automatically generates snapshot name.

You can have a: SaveTo(path) as well. In fact the Save method could just call SaveTo. Better interface imo.

provider/k8s.go
```diff
@@ -813,41 +813,41 @@ func (store genericFileStore) Close() error {
...
-func (config *FileFileStoreConfig) Deserialize(data []byte) error {
+func (config *LocalFileStoreConfig) Deserialize(data []byte) error {
 	err := json.Unmarshal(data, config)
 	if err != nil {
 		return fmt.Errorf("deserialize file blob store config: %w", err)
 	}
 	return nil
 }
 
-type FileFileStore struct {
```
simba-git, line 836: LOL how'd this get in

backup/main/main.go
sdreyer: Changed to SaveTo. Believe the idea was that in the future, we may want to use this to allow people manually create/name backups. Ex. before a migration, critical changes, etc

provider/k8s.go
sdreyer: 🥸

backup/backup.go
sdreyer: Was unsure what you meant. Changed it to provider.FilestoreType so it matches with the filestores we have implemented

backup/main/main.go
simba-git: This wasn't changed it looks like?

backup/backup.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Client Allows ETCD to be tested with mock values
+type Client interface {
+	Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.GetResponse, error)
+}
+
+type BackupManager struct {
+	ETCDClient Client
+	Provider   provider.FileStoreType
+	as         provider.Provider
```
simba-git, line 22: better name?

backup/backup.go
```diff
@@ -0,0 +1,113 @@
...
+func (b *BackupManager) Restore() error {
+	panic(fmt.Errorf("Restore() unimplemented"))
+	return nil
+}
+
+func (b *BackupManager) getBackupProvider() (Provider, error) {
+	var backupProvider Provider
+	switch b.Provider {
+	case provider.Azure:
+		backupProvider = &Azure{}
```
simba-git, line 55: The user should set &Azure or &Local in the struct above. This allows them to overwrite any fields they want to.

backup/backup.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return backupProvider, nil
+}
+
+type backupRow struct {
+	Key   []byte
+	Value []byte
+}
+
+type backupFile []backupRow
```
simba-git, line 70: If you can easily rename this to "backup" I would. I worry that someone will make this and think it's already saved in a file given the name.

backup/backup.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (b *BackupManager) takeSnapshot(filename string) error {
+	resp, err := b.ETCDClient.Get(context.Background(), "", clientv3.WithPrefix())
+	if err != nil {
+		return fmt.Errorf("could get snapshot values: %v", err)
+	}
+
+	values := b.convertEtcdToBackupFile(resp.Kvs)
```
simba-git, line 78: convertEtcdToBackup

I worry that someone will think it automatically saves it to a file

backup/backup.go
```diff
@@ -0,0 +1,113 @@
...
+	values.writeTo(filename)
```
simba-git, line 79: Not capturing the error here.

Can also make this one line:
err := b.convertEtcdToBackup(resp.Kvs).WriteTo(filename)

backup/main/main.go
simba-git: Also to be clear there should be two methods:
Save() and SaveTo(path)
Save() is literally SaveTo(GenerateSnapshotName())

sdreyer: Oh gotcha

backup/backup.go
sdreyer: Can you clarify? Are you saying to remove the switch and just set the struct directly?

simba-git: Yes
BackupManager{client, &Azure{}}

backup/main/main.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	client, err := clientv3.New(etcdConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	backupExecutor := backup.BackupManager{
+		ETCDClient: client,
+		Provider:   provider.FileStoreType(help.GetEnv("CLOUD_PROVIDER", "LOCAL")),
```
simba-git, line 35: Provider: provider.Azure{} or provider.Local{}

You could consider moving the env variables from the Init() all the way up to here actually.

PR #593, opened 2023-01-14
Restore Script

sdreyer:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


backup/backup.go
```diff
@@ -79,19 +138,45 @@ func (b *BackupManager) convertEtcdToBackup(resp []*mvccpb.KeyValue) backup {
...
+	if err != nil {
+		return fmt.Errorf("could not write to ETCD: %v", err)
+	}
+	return nil
+}
+
+func (b *BackupManager) writeToEtcd(data backup) error {
+	for _, row := range data {
+		_, err := b.ETCDClient.Put(context.Background(), string(row.Key), string(row.Value))
+		if err != nil {
```
ahmadnazeri, line 163: Does `ETCDClient.Put` have retries built in?

ahmadnazeri (approved):
Some questions

.github/workflows/e2e.yml
```diff
@@ -348,19 +358,17 @@ jobs:
...
       - name: Create Secret
         env:
-          CLOUD_PROVIDER: "AZURE"
-          AZURE_STORAGE_ACCOUNT: ${{ secrets.AZURE_STORAGE_ACCOUNT }}
-          AZURE_STORAGE_TOKEN: ${{ secrets.AZURE_STORAGE_TOKEN }}
+          AZURE_ACCOUNT_NAME: ${{ secrets.AZURE_ACCOUNT_NAME }}
+          AZURE_STORAGE_KEY: ${{ secrets.AZURE_ACCOUNT_KEY }}
           AZURE_CONTAINER: ${{ secrets.AZURE_CONTAINER_NAME }}
-          STORAGE_PATH: ${{ secrets.STORAGE_PATH }}
+          STORAGE_PATH: "backup"
```
simba-git, line 368: This on purpose? I dont understand

.github/workflows/testing.yml
```diff
@@ -444,12 +444,31 @@ jobs:
           git clone -b v3.4.16 https://github.com/etcd-io/etcd.git
           cd etcd
           ./build
-          export PATH="$PATH:`pwd`/bin"
+          echo "`pwd`/bin"
+          echo "`pwd`/bin" >> $GITHUB_PATH
```
simba-git, line 448: ?

backup/backup.go
```diff
@@ -79,19 +138,45 @@ func (b *BackupManager) convertEtcdToBackup(resp []*mvccpb.KeyValue) backup {
...
+func (b *BackupManager) loadSnapshot(filename string) error {
+	backupData := backup{}
+	err := backupData.readFrom(filename)
 	if err != nil {
-		return fmt.Errorf("could not marshal snapshot: %v", err)
+		return fmt.Errorf("could not read from file %s: %v", filename, err)
 	}
-	if err = ioutil.WriteFile(filename, file, 0644); err != nil {
-		return fmt.Errorf("could not write snapshot to file: %v", err)
+	fmt.Println("Clearing ETCD")
```
simba-git, line 147: on purpose?

backup/create_secret.py
```diff
@@ -0,0 +1,87 @@
...
+        "AZURE_STORAGE_PATH": container_path
+    }
+    with open("./backup_secret.yaml", 'w+') as f:
+        yaml.dump(secretBase, f)
+
+@cli.command()
+@click.argument("aws_access_key", required=True)
+@click.argument("aws_secret_key", required=True)
+@click.argument("bucket_region", required=True)
+@click.argument("bucket_path", required=True)
```
simba-git, line 58: What is a bucket path? Is it different from a bucket name?

backup/save/main.go
```diff
@@ -30,7 +30,7 @@ func main() {
 		panic(err)
 	}
 
-	p := provider.FileStoreType(help.GetEnv("CLOUD_PROVIDER", "LOCAL"))
+	p := provider.FileStoreType(help.GetEnv("CLOUD_PROVIDER", "FILE_SYSTEM"))
```
simba-git, line 33: what is file system?

provider/k8s.go
```diff
@@ -543,9 +543,7 @@ func (store genericFileStore) NewestFile(prefix string) (string, error) {
 	mostRecentKey := ""
 	for {
 		if listObj, err := listIterator.Next(context.TODO()); err == nil {
-			pathParts := strings.Split(listObj.Key, ".")
-			fileType := pathParts[len(pathParts)-1]
-			if fileType == "parquet" && !listObj.IsDir && (listObj.ModTime.After(mostRecentTime) || listObj.ModTime.Equal(mostRecentTime)) {
+			if !listObj.IsDir && (listObj.ModTime.After(mostRecentTime) || listObj.ModTime.Equal(mostRecentTime)) {
```
simba-git, line 546: Make this a variable or helper method

simba-git (commented):
Mostly looks good, the script and config changes confused me a bit

.github/workflows/testing.yml
sdreyer: This is how you add things to PATH in github actions. Its just making the etcd and etcdctl executables available in other steps

backup/save/main.go
sdreyer: File system is the local filesystem. I can change it to LOCAL_FILESYSTEM instead of that clarified. Its just the value that the filestore enum uses

backup/backup.go
sdreyer: I believe it has built in retries but times out after a time limit. It could result in a partial restore. 

I can make it so that if a key fails, it clears etcd again

.github/workflows/e2e.yml
sdreyer: yeah its so its more visible what the path is for debugging purposes. Otherwise you need to remember or have it documented somewhere since its stored as a secret

backup/backup.go
sdreyer: Left it so the user has some reference to if its running rather than waiting for an empty terminal. Can change it to be a log too

backup/create_secret.py
sdreyer: Changed this. The filestore values are name improperly but need to create an RFC for it

.github/workflows/testing.yml
simba-git: Maybe add an echo above these two that says
echo "Adding to Github Actions Path:"
echo "`pwd`/bin"

backup/backup.go
simba-git: Can we make it a log please

backup/save/main.go
simba-git: Please do

simba-git: preferably set up a const even, best to not use strings like this when we mean them to be consts.

backup/backup.go
simba-git: I'm not sure whats best here, probably have it try its best then make sure its very clear that it only partially restored.

backup/save/main.go
sdreyer: There was a const setup, unsure why I didnt use it

.github/workflows/testing.yml
sdreyer: Done

backup/backup.go
sdreyer: Changed to logs

PR #619, opened 2023-02-04
Model registration e2e tests

epps:
# Description

This PR covers the E2E tests for registering models, both prior to and at the time of serving, and getting registered models.

### Select type(s) of change

- [ ] Documentation update
- [x] E2E tests

# Checklist:

### DIFF
--- a/client/tests/conftest.py
+++ b/client/tests/conftest.py
@@ -1,8 +1,27 @@
+import csv
+import os
 import pytest
 import sys
+import shutil
+import stat
+from tempfile import NamedTemporaryFile
 sys.path.insert(0, 'client/src/')
-from featureform.register import Registrar, OfflineSparkProvider
-from featureform.resources import SparkConfig, Provider, DatabricksCredentials, AzureFileStoreConfig, AWSCredentials, S3StoreConfig, EMRCredentials
+from featureform.register import Registrar, OfflineSparkProvider, LocalProvider, OfflineSQLProvider
+from featureform.resources import (
+    AWSCredentials,
+    AzureFileStoreConfig,
+    DatabricksCredentials,
+    EMRCredentials,
+    LocalConfig,
+    S3StoreConfig,
+    SparkConfig,
+    PostgresConfig,
+    Provider,
+)
+import featureform as ff
+
+real_path = os.path.realpath(__file__)
+dir_path = os.path.dirname(real_path)
 
 pytest_plugins = [
     'connection_test',
@@ -120,3 +139,74 @@ def azure_blob():
 @pytest.fixture(scope="module")
 def s3(aws_credentials):
     return S3StoreConfig("bucket_path", "bucket_region", aws_credentials)
+
+@pytest.fixture(scope="module")
+def local_provider():
+    ff.register_user("test_user").make_default_owner()
+    return ff.register_local()
+
+@pytest.fixture(scope="module")
+def local_provider_and_source(local_provider):
+    local_source = local_provider.register_file(
+                name="transactions",
+                variant="quickstart",
+                description="A dataset of fraudulent transactions.",
+                path=f"{dir_path}/test_files/input_files/transactions.csv"
+            )
+    return (local_provider, local_source)
+
+@pytest.fixture(scope="module")
+def resource_serving_clients():
+    def get_clients_for_context(is_local):
+            resource_client = ff.ResourceClient(local=True)
+            serving_client = ff.ServingClient(local=True)
+
+            return (resource_client, serving_client)
+    
+    return get_clients_for_context
+
+@pytest.fixture(scope="module")
+def create_temp_file(contents):
+    file = NamedTemporaryFile(delete=False)
+    with open(file.name, 'w') as csvfile:
+        writer = csv.writer(csvfile, delimiter=',', quotechar='|')
+        writer.writerow(contents['columns'])
+        for row in contents['values']:
+            writer.writerow(row)
+        csvfile.close()
+
+    return file.name
+
+@pytest.fixture(scope="module")
+def setup_teardown():
+    def clear_state_and_reset():
+        ff.clear_state()
+        shutil.rmtree('.featureform', onerror=del_rw)
+
+    return clear_state_and_reset
+
+def del_rw(action, name, exc):
+    if os.path.exists(name):
+        os.chmod(name, stat.S_IWRITE)
+        os.remove(name)
+
+@pytest.fixture(scope="module")
+def hosted_sql_provider_and_source():
+    provider = ff.register_postgres(
+        name = "postgres-quickstart",
+        host="0.0.0.0",
+        port="5432",
+        user="postgres",
+        password="password",
+        database="postgres",
+        description = "A Postgres deployment we created for the Featureform quickstart"
+    )
+
+    source = provider.register_table(
+        name = "transactions",
+        variant = "kaggle",
+        description = "Fraud Dataset From Kaggle",
+        table = "Transactions", # This is the table's name in Postgres
+    )
+
+    return (provider, source)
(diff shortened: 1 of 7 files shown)


client/tests/register_test.py
```diff
@@ -90,6 +90,37 @@ def my_function(df):
     # Checks that Transformation definition does not error when converting to source
     dec.to_source()
 
+@pytest.mark.parametrize(
```
epps, line 93: @ahmadnazeri, I added my model registration tests here given this file already follows the patterns we discussed; however, I still think we could clean things up (e.g. reusable fixtures into `conftest.py` and standardize on arrangement, etc.).

Given this, if you think it's important to separate out the "newer" tests into their own files, I can move these to a new file (e.g. `test_model_registration.py`). Let me know what you think.

client/tests/conftest.py
```diff
@@ -120,3 +139,74 @@ def azure_blob():
...
+            resource_client = ff.ResourceClient(local=True)
+            serving_client = ff.ServingClient(local=True)
+
+            return (resource_client, serving_client)
+    
+    return get_clients_for_context
+
+@pytest.fixture(scope="module")
+def create_temp_file(contents):
+    file = NamedTemporaryFile(delete=False)
```
ahmadnazeri, line 170: this is cool, I wasn't aware of it. There is also `tmp_path` with pytest (https://docs.pytest.org/en/6.2.x/tmpdir.html)

client/tests/register_test.py
ahmadnazeri: no worries, this it good!

ahmadnazeri (approved):
just one comment about registering a model with the same name; otherwise, lgtm.

client/tests/conftest.py
epps: @ahmadnazeri, I can't take credit for this: I simply moved a local fixture from `serving_test.py` into `conftest.py`. I noticed the `tmpdir` fixture in the pytest docs. Should I give that a try to avoid rolling custom solutions to a solved problem?

client/tests/register_test.py
```diff
@@ -90,6 +90,37 @@ def my_function(df):
...
+    "resource_serving_client_fxt,is_local",
+    [
+        ('resource_serving_clients', True),
+        ('resource_serving_clients', False),
+    ]
+)
+def test_valid_model_registration(resource_serving_client_fxt, is_local, request):
+    resource_client = request.getfixturevalue(resource_serving_client_fxt)(is_local)[0];
+    model_name = "model_a"
```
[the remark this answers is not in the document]
epps: You're right: I should add a test for this. I believe the behavior is an exception from the API that's simply printed out to the console as: `Resource <RESOURCE> already registered.`. I'll add a test for that now.

client/tests/test_serving_model.py
```diff
@@ -0,0 +1,195 @@
...
+@pytest.fixture(autouse=True)
+def before_and_after_each(setup_teardown):
+    setup_teardown()
+    yield
+    setup_teardown()
+
+
+def arrange_resources(provider, source, resource_client, is_local):
+    if is_local:
+        @provider.df_transformation(variant="quickstart", inputs=[("transactions", "quickstart")])
```
sdreyer, line 164: This is probably okay since its already here, but not necessary to have a transformation for the model test

client/tests/test_serving_model.py
```diff
@@ -0,0 +1,195 @@
...
+            return transactions.groupby("CustomerID")["TransactionAmount"].mean()
+    else:
+        @provider.sql_transformation(variant="quickstart")
+        def average_user_transaction():
+            return "SELECT CustomerID as user_id, avg(TransactionAmount) as avg_transaction_amt from {{transactions.kaggle}} GROUP BY user_id"
+
+    user = ff.register_entity("user")
+    average_user_transaction.register_resources(
+        entity=user,
+        entity_column="CustomerID",
```
sdreyer, line 175: Entity column would be user_id in this case

client/tests/test_serving_model.py
```diff
@@ -0,0 +1,195 @@
...
+    # Arranges the resources context following the Quickstart pattern
+    arrange_resources(provider, source, resource_client, is_local)
+
+    model_name_a = 'fraud_model_a';
+    model_name_b = 'fraud_model_b';
+
+    serving_client.training_set("fraud_training", "quickstart", model=model_name_a)
+    serving_client.training_set("fraud_training", "quickstart", model=model_name_b)
+
+    models = resource_client.list_models(is_local)
```
sdreyer, line 53: Should have a test case for empty `list_models` unless I missed it somewhere else

client/tests/conftest.py
```diff
@@ -1,8 +1,27 @@
+import csv
```
sdreyer, line 1: Add these tests to the makefile so they run in CICD

client/tests/register_test.py
```diff
@@ -90,6 +90,37 @@ def my_function(df):
...
 
+@pytest.mark.parametrize(
+    "resource_serving_client_fxt,is_local",
+    [
+        ('resource_serving_clients', True),
+        ('resource_serving_clients', False),
+    ]
+)
+def test_valid_model_registration(resource_serving_client_fxt, is_local, request):
+    resource_client = request.getfixturevalue(resource_serving_client_fxt)(is_local)[0];
```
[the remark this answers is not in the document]
epps: This is getting the fixture. I add the `_fxt` suffix to denote that the argument is really the name of the fixture rather than the fixture itself. I can change this if it's throwing anyone off.

client/tests/conftest.py
```diff
@@ -120,3 +139,74 @@ def azure_blob():
...
+                name="transactions",
+                variant="quickstart",
+                description="A dataset of fraudulent transactions.",
+                path=f"{dir_path}/test_files/input_files/transactions.csv"
+            )
+    return (local_provider, local_source)
+
+@pytest.fixture(scope="module")
+def resource_serving_clients():
+    def get_clients_for_context(is_local):
```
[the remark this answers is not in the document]
epps: Thanks for catching this! This is a leftover from some testing. I meant to pass the `is_local` arg to the client instantiations.

client/tests/conftest.py
ahmadnazeri: if you have time then that would be great but no worries since `NamedTemporaryFile` is builtin, I believe.

client/tests/register_test.py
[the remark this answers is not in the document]
epps: Never mind. I misspoke here: this logic only seems to apply to resources that can be either be registered as a name (i.e. `str`) or an instance of the resource Class.

client/tests/test_serving_model.py
epps: Awww, interesting, @sdreyer. I wasn't aware this would work. The [Quickstart (Local)](https://docs.featureform.com/quickstart-local) _and_ the [Python SDK docs](https://sdk.featureform.com/register/#src.featureform.register.ColumnSourceRegistrar.register_resources) so `"CustomerID"` as the value for `entity_column`.

Should I update my docs PR with `"user_id"` for the value of `entity_column` to ensure we're consistent?

client/tests/register_test.py
epps: Just noticed I moved this into `conftest.py` and never actually used it as I ended up moving my model serving tests to their own file and registering my primary source from the `/test_files/input_files`.

When I time comes to further refactor our existing tests, I'll definitely look into this builtin fixture.

client/tests/register_test.py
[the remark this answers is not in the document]
epps: It looks like I'd have to run `apply` in the test and capture the output to `stdout`; however, I'm having trouble seeing that output using `capsys`. I've left the test in place as a check that adding the same model name twice wouldn't throw an exception.

client/tests/register_test.py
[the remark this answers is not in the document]
sdreyer: Ah I see

client/tests/register_test.py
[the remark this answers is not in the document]
sdreyer: Yeah I think its fine

sdreyer (approved):

PR #621, opened 2023-02-06
Adding the client changes for Spark Executor

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/resources.py
+++ b/client/src/featureform/resources.py
@@ -1399,7 +1399,7 @@ class EMRCredentials:
 
 @typechecked
 @dataclass
-class SparkExecutor:
+class SparkCredentials:
     def __init__(self,
                  master: str,
                  deploy_mode: str,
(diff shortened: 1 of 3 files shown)


client/src/featureform/resources.py
```diff
@@ -1397,5 +1397,27 @@ def config(self):
             "Credentials": self.credentials.config(),
         }
 
+@typechecked
+@dataclass
+class SparkExecutor:
+    def __init__(self,
```
sdreyer, line 1403: Are there any other args we would want to allow?

client/src/featureform/resources.py
```diff
@@ -1397,5 +1397,27 @@ def config(self):
             "Credentials": self.credentials.config(),
         }
 
+@typechecked
+@dataclass
+class SparkExecutor:
```
sdreyer, line 1402: Should see if @simba-git has any thoughts on what this should be called since right now it'll be 
`register_spark(SparkExecutor())`, which feels weird. The other options are EMRCredentials and DatabricksCredentials, are there spark credentials that someone might need, SparkCredentials? Otherwise something like SparkConfig?

client/src/featureform/resources.py
ahmadnazeri: the only thing that I can think of is configurations like memory, executors, etc.

sdreyer (approved):

PR #625, opened 2023-02-08
Model Registration: Local Client Implementation

epps:
# Description

The PR introduces the complete local implementation of registering, getting and listing models, even those that are registered at the time of serving training sets or features.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/format.py
+++ b/client/src/featureform/format.py
@@ -1,3 +1,9 @@
+from typing import List
+from rich.console import Console
+from rich.table import Table
+from .resources import Resource
+
+
 two_row_spacing = "{:<30} {:<25}"
 three_row_spacing = "{:<30} {:<30} {:<30}"
 four_row_spacing = "{:<30} {:<30} {:<30} {:<30}"
@@ -17,3 +23,30 @@ def format_rows(format_obj, format_obj_2=None, format_obj_3=None, format_obj_4=N
 def format_pg(s=""):
     print(divider)
     print(s)
+
+
+class CliFormatter:
+
+    def __init__(self) -> None:
+        self.__console = Console()
+
+    def print_to_table(self, resources: List[Resource]):
+        if len(resources) == 0:
+            return
+
+        first_resource = resources[0]
+
+        table = Table(f"{first_resource.type().upper()} NAME", "TYPE", "STATUS")
+
+        for resource in resources:
+            status = resource.get_status() if self._has_status(resource) else ""
+
+            table.add_row(resource.name, resource.type(), status)
+
+        self.__console.print(table)
+
+    def _has_status(self, resource: Resource):
+        return hasattr(resource, 'status') and callable(getattr(resource, 'status'))
+
+
+CliFormatter = CliFormatter()
--- a/client/src/featureform/format.py
+++ b/client/src/featureform/format.py
@@ -1,9 +1,3 @@
-from typing import List
-from rich.console import Console
-from rich.table import Table
-from .resources import Resource
-
-
 two_row_spacing = "{:<30} {:<25}"
 three_row_spacing = "{:<30} {:<30} {:<30}"
 four_row_spacing = "{:<30} {:<30} {:<30} {:<30}"
@@ -23,30 +17,3 @@ def format_rows(format_obj, format_obj_2=None, format_obj_3=None, format_obj_4=N
 def format_pg(s=""):
     print(divider)
     print(s)
-
-
-class CliFormatter:
-
-    def __init__(self) -> None:
-        self.__console = Console()
-
-    def print_to_table(self, resources: List[Resource]):
-        if len(resources) == 0:
-            return
-
-        first_resource = resources[0]
-
-        table = Table(f"{first_resource.type().upper()} NAME", "TYPE", "STATUS")
-
-        for resource in resources:
-            status = resource.get_status() if self._has_status(resource) else ""
-
-            table.add_row(resource.name, resource.type(), status)
-
-        self.__console.print(table)
-
-    def _has_status(self, resource: Resource):
-        return hasattr(resource, 'status') and callable(getattr(resource, 'status'))
-
-
-CliFormatter = CliFormatter()
--- a/client/src/featureform/format.py
+++ b/client/src/featureform/format.py
@@ -4,7 +4,10 @@ four_row_spacing = "{:<30} {:<30} {:<30} {:<30}"
 divider = "-----------------------------------------------"
 
 def format_rows(format_obj, format_obj_2=None, format_obj_3=None, format_obj_4=None):
-    if format_obj_2 is None:
+    # Base case for when `format_obj` is a string
+    if format_obj_2 is None and type(format_obj) == str:
+        print(format_obj)
+    elif format_obj_2 is None:
         for s in format_obj:
             format_rows(*s)
     elif format_obj_2 is not None and format_obj_3 is None:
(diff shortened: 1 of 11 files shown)


client/src/featureform/format.py
```diff
@@ -16,4 +22,31 @@ def format_rows(format_obj, format_obj_2=None, format_obj_3=None, format_obj_4=N
 
 def format_pg(s=""):
     print(divider)
-    print(s)
\ No newline at end of file
+    print(s)
+
+
+class CliFormatter:
```
epps, line 28: @sdreyer, I think [Rich](https://github.com/Textualize/rich) is a great candidate for a full-featured CLI formatting library and I originally thought I'd include it in this PR as we discussed; however, given the scope of changes required to parse out formatting logic from data access logic, I think it would be best to remove it at this time.

I left this example here to demonstrate how we could wrap Rich in our own class to expose a generalized formatting API to the Python CLI.

client/src/featureform/register.py
```diff
@@ -2415,9 +2440,9 @@ def apply(self):
             return
 
         if self.local:
-            state().create_all_local()
+            self.state().create_all_local()
         else:
-            state().create_all(self._stub)
+            self.state().create_all(self._stub)
```
epps, line 2445: @sdreyer, this is the issue that caused me to chase my own tail for a bit. I've run the tests and it doesn't appear to impact the hosted mode, but let me know what you think about this change.

client/src/featureform/register.py
```diff
@@ -2540,7 +2565,11 @@ def get_model(self, name):
         Returns:
             model (Model): Model
         """
-        return get_resource_info(self._stub, "model", name)
+        model = get_model_info_local(name) if local else get_resource_info(self._stub, "model", name)
+
+        CliFormatter.print_to_table([model])
```
epps, line 2570: TODO: remove before merging.

client/src/featureform/register.py
simba-git: interesting, def looks good to me

client/src/featureform/register.py
```diff
@@ -3424,15 +3453,20 @@ def list_training_sets(self, local=False):
...
         """List all models. Prints a list of all models.
 
         Returns:
             models (List[Model]): List of Model Objects
         """
+        models = []
         if local:
-            return list_local("model", [ColumnName.NAME, ColumnName.STATUS, ColumnName.DESCRIPTION])
-        return list_name_status_desc(self._stub, "model")
+            rows = list_local("model", [ColumnName.NAME, ColumnName.STATUS])
```
simba-git, line 3464: Models shouldn't have a status

client/src/featureform/serving.py
```diff
@@ -461,6 +482,14 @@ def check_missing_values(self, resource, df):
             raise KeyError(f"Timestamp column does not exist: {resource['source_timestamp']}")
 
 
+    def _register_model(self, model: Union[str, Model], look_up_table: str, association_name: str, association_variant: str):
+        name = model if isinstance(model, str) else model.name
+        type = "Model" if isinstance(model, str) else model.type()
+
+        self.db.insert("models", name, type, "ready")
```
simba-git, line 489: What should the status be here @sdreyer

client/src/featureform/sqlite_metadata.py
```diff
@@ -175,9 +175,22 @@ def createTables(self):
...
           type        text,
-          description text,
           status      text);''')
 
+        # models training sets table
+        self.__conn.execute('''CREATE TABLE IF NOT EXISTS model_training_sets(
+          model_name           text NOT NULL,
+          training_set_name    text NOT NULL,
+          training_set_variant text NOT NULL,
+          UNIQUE(model_name, training_set_name, training_set_variant));''')
```
simba-git, line 185: Do we have a test for if we register a model to the same feature twice?

client/src/featureform/register.py
epps: A model's status didn't make much sense to me either; however, several other record types has a status field (e.g. entity and user) and the DDL for the `models` table already contained it so I wasn't certain whether I should remove it or not.

I'll remove it.

client/src/featureform/sqlite_metadata.py
epps: Yes, we do. In `client/tests/test_serving_model.py`, I've included this test for both training sets and features:
* `test_registering_same_model_twice_while_serving_training_set`
* `test_registering_same_model_twice_while_serving_features`

I haven't, however, included a test for registering the same model to two different features/training sets. I assumed this test would cover that behavior as well, but if you feel otherwise, I can add tests for that case, too.

client/src/featureform/register.py
sdreyer: The status for user and entity should probably be removed as well. I would remove it from models and leave the others for now just incase other things (like prints) need to be modified so that those arent broken (out of scope)

client/src/featureform/register.py
```diff
@@ -2358,6 +2370,19 @@ def register_training_set(self,
         )
         self.__resources.append(resource)
 
+    def register_model(self, name: str) -> ModelRegistrar:
```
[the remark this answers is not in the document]
epps: I wondered about this, too. Given my frame of reference when starting this work was _"... it'll look a lot like the implementation for `User` ..."_, I was following that pattern. So given `register_user` returns an instance of `UserRegistrar`, I assumed it would be the same for model.

Based on your comment, I'm assuming `register_user` should return an instance of `User` instead, right?

client/src/featureform/register.py
epps: @sdreyer, I think the issue was that even when I wasn't using the global `ff` (i.e. the global registrar) my resources were being applied to it because the code path was calling the global `state` method even when I had a "local" resource client. When I traced down the issue, I saw this and figured it should be `self.state` in both cases given that would still allow the global registrar to behave as it has been, but a local instance of the resource client would hold its own registrar. Does this make sense?

client/src/featureform/get_local.py
```diff
@@ -371,3 +372,11 @@ def get_variant(resource_type, name, variant):
 def format_resource_list(res_list, name_col="name", var_col="variant"):
     formatted_res = [{ "name": r[name_col], "variant": r[var_col]} for r in res_list]
     return formatted_res
+
+def get_model_info_local(name) -> Model:
+    model = get_resource("model", name)
+
+    format_rows("MODEL NAME: ", model["name"])
```
simba-git, line 379: Always thought it was a bad idea to make the get call also print stuff, seems like it'd be annoying if I was just using FF normally

simba-git: No need to change in this PR

client/src/featureform/format.py
```diff
@@ -4,7 +4,10 @@
 divider = "-----------------------------------------------"
 
 def format_rows(format_obj, format_obj_2=None, format_obj_3=None, format_obj_4=None):
```
simba-git, line 6: No need to change in this PR, but I'm very much not a fan of these parameter names

simba-git (approved):

client/src/featureform/get_local.py
epps: I've been discussing this with @sdreyer here and there, and I agree it would be nice to separate out concerns here, as well as introduce more formatting options. But agreed: well out of scope for this PR, as well as the RFC.

PR #628, opened 2023-02-09
Task/spark generic executor/backend

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/resources.py
+++ b/client/src/featureform/resources.py
@@ -1403,13 +1403,35 @@ class SparkCredentials:
     def __init__(self,
                  master: str,
                  deploy_mode: str,
+                 python_version: str,
                 ):
 
         if deploy_mode != "cluster" and deploy_mode != "client":
             raise Exception(f"Spark does not support '{deploy_mode}' deploy mode. It only supports 'cluster' and 'client'.")
 
+        python_version = self._verify_python_version(python_version)
+
         self.master = master
         self.deploy_mode = deploy_mode
+        self.python_version = python_version
+    
+    def _verify_python_version(self, version):
+        if version.count(".") == 2:
+            major, minor, _ = version.split(".")
+        elif version.count(".") == 1:
+            major, minor = version.split(".")    
+        else:
+            raise Exception("Please specify your Python version on the Spark cluster. Accepted formats: Major.Minor or Major.Minor.Patch; ex. '3.7' or '3.7.16")
+
+        if major != "3" or minor not in ["7", "8", "9", "10"]:
+            raise Exception(f"The Python version {version} is not supported. Currently, supported versions are 3.7-3.10.")
+        
+        if minor == "10":
+            patch = "10"
+        else:
+            patch = "16"
+        
+        return f"{major}.{minor}.{patch}"
 
     def type(self):
         return "SPARK"
@@ -1418,6 +1440,7 @@ class SparkCredentials:
         return {
             "Master": self.master,
             "DeployMode": self.deploy_mode,
+            "PythonVersion": self.python_version
         }
 
 ExecutorCredentials = Union[EMRCredentials, DatabricksCredentials, SparkCredentials]
--- a/client/src/featureform/resources.py
+++ b/client/src/featureform/resources.py
@@ -21,6 +21,11 @@ from featureform.format import *
 NameVariant = Tuple[str, str]
 
 
+# Constants for Pyspark Versions
+MAJOR_VERSION = "3"
+MINOR_VERSIONS = ["7", "8", "9", "10"]
+
+
 class ColumnTypes(Enum):
     NIL = ""
     INT = "int"
@@ -1423,7 +1428,7 @@ class SparkCredentials:
         else:
             raise Exception("Please specify your Python version on the Spark cluster. Accepted formats: Major.Minor or Major.Minor.Patch; ex. '3.7' or '3.7.16")
 
-        if major != "3" or minor not in ["7", "8", "9", "10"]:
+        if major != MAJOR_VERSION or minor not in MINOR_VERSIONS:
             raise Exception(f"The Python version {version} is not supported. Currently, supported versions are 3.7-3.10.")
         
         if minor == "10":
--- a/client/src/featureform/resources.py
+++ b/client/src/featureform/resources.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import sys
 import json
 import time
 from enum import Enum
@@ -1408,19 +1409,30 @@ class SparkCredentials:
     def __init__(self,
                  master: str,
                  deploy_mode: str,
-                 python_version: str,
+                 python_version: str = "",
                 ):
+        
+        self.master = master.lower()
+        self.deploy_mode = deploy_mode.lower()
 
-        if deploy_mode != "cluster" and deploy_mode != "client":
-            raise Exception(f"Spark does not support '{deploy_mode}' deploy mode. It only supports 'cluster' and 'client'.")
+        if self.deploy_mode != "cluster" and self.deploy_mode != "client":
+            raise Exception(f"Spark does not support '{self.deploy_mode}' deploy mode. It only supports 'cluster' and 'client'.")
 
-        python_version = self._verify_python_version(python_version)
-
-        self.master = master
-        self.deploy_mode = deploy_mode
-        self.python_version = python_version
+        self.python_version = self._verify_python_version(self.deploy_mode, python_version)
     
-    def _verify_python_version(self, version):
+    def _verify_python_version(self, deploy_mode, version):
+        if deploy_mode == "cluster" and version == "":
+            client_python_version = sys.version_info
+            client_major = str(client_python_version.major)
+            client_minor = str(client_python_version.minor)
+
+            if client_major != MAJOR_VERSION:
+                client_major = "3"
+            if minor not in MINOR_VERSIONS:
+                client_minor = "7"
+
+            version = f"{client_major}.{client_minor}"
+
         if version.count(".") == 2:
             major, minor, _ = version.split(".")
         elif version.count(".") == 1:
@@ -1431,6 +1443,10 @@ class SparkCredentials:
         if major != MAJOR_VERSION or minor not in MINOR_VERSIONS:
             raise Exception(f"The Python version {version} is not supported. Currently, supported versions are 3.7-3.10.")
         
+        """
+        The Python versions on the Docker image are 3.7.16, 3.8.16, 3.9.16, and 3.10.10. 
+        This conditional statement sets the patch number based on the minor version. 
+        """
         if minor == "10":
             patch = "10"
         else:
--- a/coordinator/Dockerfile
+++ b/coordinator/Dockerfile
@@ -28,7 +28,7 @@ COPY ./coordinator/main/main.go ./coordinator/main/main.go
 ENV CGO_ENABLED=0
 RUN go build ./coordinator/main/main.go
 
-FROM ubuntu:latest
+FROM ubuntu:22.04
 
 COPY --from=0 ./app/main /app/main
 COPY ./provider/scripts/spark/offline_store_spark_runner.py /app/provider/scripts/spark/offline_store_spark_runner.py
--- a/coordinator/Dockerfile
+++ b/coordinator/Dockerfile
@@ -24,21 +24,35 @@ COPY ./kubernetes ./kubernetes
 COPY ./types ./types
 COPY ./coordinator/main/main.go ./coordinator/main/main.go
 
+ENV CGO_ENABLED=0
 RUN go build ./coordinator/main/main.go
 
-FROM alpine
+FROM ibmjava:8
 
-COPY --from=0 ./app/main ./main
+COPY --from=0 ./app/main /app/main
 COPY ./provider/scripts/spark/offline_store_spark_runner.py /app/provider/scripts/spark/offline_store_spark_runner.py
 COPY ./provider/scripts/spark/python_packages.sh /app/provider/scripts/spark/python_packages.sh
 COPY ./provider/scripts/spark/requirements.txt /app/provider/scripts/spark/requirements.txt
 
 # Installing pyenv
-RUN apk --update add libffi-dev ncurses-dev openssl-dev readline-dev \
-tk-dev xz-dev zlib-dev curl git bash build-base bzip2-dev sqlite-dev \
-openjdk11 coreutils procps
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    checkinstall \
+    libreadline-gplv2-dev \
+    libncursesw5-dev \
+    libssl-dev \
+    libsqlite3-dev \
+    libgdbm-dev \
+    libc6-dev \
+    libbz2-dev \
+    libffi-dev \
+    zlib1g-dev \
+    liblzma-dev \
+    curl \
+    git \
+    && rm -rf /var/lib/apt/lists/*
 
-ENV ENV="/root/.ashrc"
+ENV ENV="/root/.bashrc"
 ENV PYENV_ROOT="/.pyenv"
 ENV PATH="$PYENV_ROOT/bin:$PATH"
 RUN echo "PATH=${PATH}" > "${ENV}"
@@ -55,4 +69,4 @@ ENV SPARK_SCRIPT_PATH="/app/provider/scripts/spark/offline_store_spark_runner.py
 ENV PYTHON_INIT_PATH="/app/provider/scripts/spark/python_packages.sh"
 
 EXPOSE 8080
-ENTRYPOINT ["./main"]
+ENTRYPOINT ["/app/main"]
--- a/coordinator/Dockerfile
+++ b/coordinator/Dockerfile
@@ -24,6 +24,7 @@ COPY ./kubernetes ./kubernetes
 COPY ./types ./types
 COPY ./coordinator/main/main.go ./coordinator/main/main.go
 
+# CGD_ENABLED=0 is used to avoid the libc differences in alpine and ubuntu
 ENV CGO_ENABLED=0
 RUN go build ./coordinator/main/main.go
 
--- a/coordinator/Dockerfile
+++ b/coordinator/Dockerfile
@@ -29,9 +29,9 @@ RUN go build ./coordinator/main/main.go
 FROM alpine
 
 COPY --from=0 ./app/main ./main
-COPY ./app/provider/scripts/spark/offline_store_spark_runner.py /app/provider/scripts/spark/offline_store_spark_runner.py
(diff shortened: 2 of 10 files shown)


provider/spark.go
```diff
@@ -37,8 +38,9 @@ import (
 type SparkExecutorType string
 
 const (
-	EMR        SparkExecutorType = "EMR"
-	Databricks                   = "DATABRICKS"
+	EMR          SparkExecutorType = "EMR"
+	Databricks                     = "DATABRICKS"
+	SparkGeneric                   = "SPARK"
 )
```
epps, line 44: My Go language server complains that only the first constant in the list is typed. I'm assuming that all three of these entries should be a string of type `SparkExecutorType`, right?

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+
+func (sc *SparkGenericConfig) Deserialize(config SerializedConfig) error {
+	err := json.Unmarshal(config, sc)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (sc *SparkGenericConfig) Serialize() ([]byte, error) {
```
epps, line 647: @sdreyer, I'm seeing a variance in our config struct's `Serialize` method, should we standardize these? For example:
* `PostgresConfig.Serialize` returns `[]byte` only because it panics if it encounters an error
* `RedisConfig.Serialize` returns `SerializedConfig` (type alias for `[]byte`) and also panics if it encounters an error
* `DatabricksConfig.Serialize` returns `([]byte, error)`

The other configs fall into one of these 3 buckets. We should standardize around 1, yeah?

epps: P.S. @ahmadnazeri, I'm not proposing that this should fall within the scope of your PR, just wanted to point this out.

client/src/featureform/resources.py
```diff
@@ -1403,13 +1403,35 @@ class SparkCredentials:
...
+    
+    def _verify_python_version(self, version):
+        if version.count(".") == 2:
+            major, minor, _ = version.split(".")
+        elif version.count(".") == 1:
+            major, minor = version.split(".")    
+        else:
+            raise Exception("Please specify your Python version on the Spark cluster. Accepted formats: Major.Minor or Major.Minor.Patch; ex. '3.7' or '3.7.16")
+
+        if major != "3" or minor not in ["7", "8", "9", "10"]:
```
sdreyer, line 1426: Maybe make this a const at the top of the file

sdreyer: Along with the minor version, just so when the compatibility increases in the future we're not hunting for it

coordinator/Dockerfile
```diff
@@ -24,16 +24,50 @@ COPY ./kubernetes ./kubernetes
 COPY ./types ./types
 COPY ./coordinator/main/main.go ./coordinator/main/main.go
 
+# CGD_ENABLED=0 is used to avoid the libc differences in alpine and ubuntu
+ENV CGO_ENABLED=0
 RUN go build ./coordinator/main/main.go
 
-FROM alpine
+FROM ibmjava:8
```
sdreyer, line 31: Do we know what license this uses? Is this from IBM?

provider/k8s.go
```diff
@@ -523,6 +523,9 @@ func (store AzureFileStore) PathWithPrefix(path string, remote bool) string {
 		}
 	}
 	if remote {
+		if path[:len("abfss://")] == "abfss://" {
```
sdreyer, line 526: Should define all these prefixes as consts somewhere

provider/spark.go
sdreyer: We should probably remove the panics, but out of scope for now

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+
+type SparkGenericExecutor struct {
+	master        string
+	deployMode    string
+	pythonVersion string
+	logger        *zap.SugaredLogger
+}
+
+func (s *SparkGenericExecutor) InitializeExecutor(store FileStore) error {
+	s.logger.Info("uploading pyspark script to filestore")
```
sdreyer, line 667: Capitalize

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	sparkScriptPath := helpers.GetEnv("SPARK_SCRIPT_PATH", "/scripts/offline_store_spark_runner.py")
+	err := readAndUploadFile(sparkScriptPath, store.PathWithPrefix(sparkScriptPath, false), store)
+	sparkExists, _ := store.Exists(store.PathWithPrefix(sparkScriptPath, false))
+	if err != nil && !sparkExists {
+		return fmt.Errorf("could not upload spark script: Path: %s, Error: %v", store.PathWithPrefix(sparkScriptPath, false), err)
```
sdreyer, line 672: assign pathwithprefix to a var instead of calling it 3 times, also so it removes the function nesting above.

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	deployMode    string
+	pythonVersion string
+	logger        *zap.SugaredLogger
+}
+
+func (s *SparkGenericExecutor) InitializeExecutor(store FileStore) error {
+	s.logger.Info("uploading pyspark script to filestore")
+	sparkScriptPath := helpers.GetEnv("SPARK_SCRIPT_PATH", "/scripts/offline_store_spark_runner.py")
+	err := readAndUploadFile(sparkScriptPath, store.PathWithPrefix(sparkScriptPath, false), store)
+	sparkExists, _ := store.Exists(store.PathWithPrefix(sparkScriptPath, false))
```
sdreyer, line 670: Confusing naming, something like "scriptExists"

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	return nil
+}
+
+func (s *SparkGenericExecutor) RunSparkJob(args *[]string, store FileStore) error {
+	bashCommand := "bash"
+	sparkArgs := *args
+	sparkArgsString := strings.Join(sparkArgs, " ")
+	bashCommandArgs := []string{"-c", fmt.Sprintf("pyenv global %s && pyenv exec %s", s.pythonVersion, sparkArgsString)}
+
+	s.logger.Infow("Executing the", bashCommand, "command")
```
sdreyer, line 683: Does this always just say "bash"?

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+}
 
+func (s *SparkGenericExecutor) PythonFileURI(store FileStore) string {
+	return ""
+}
+
+func (s *SparkGenericExecutor) SparkSubmitArgs(destPath string, cleanQuery string, sourceList []string, jobType JobType, store FileStore) []string {
+	sparkScriptPath := helpers.GetEnv("SPARK_SCRIPT_PATH", "/scripts/offline_store_spark_runner.py")
+
+	argList := []string{
```
sdreyer, line 707: this needs a store_type arg as well

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+		"--output_uri",
+		fmt.Sprintf("\"%s\"", destPath),
+		"--sql_query",
+		fmt.Sprintf("\"%s\"", cleanQuery),
+		"--job_type",
+		fmt.Sprintf("\"%s\"", string(jobType)),
+	}
+
+	var remoteConnectionArgs []string
+	azureStore := store.AsAzureStore()
```
sdreyer, line 726: This is weird but I fixed it in my branch

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	for _, source := range sourceList {
+		argList = append(argList, fmt.Sprintf("\"%s\"", source))
+	}
+	return argList
+}
+
+func (s *SparkGenericExecutor) GetDFArgs(outputURI string, code string, sources []string, store FileStore) ([]string, error) {
+	sparkScriptPath := helpers.GetEnv("SPARK_SCRIPT_PATH", "/scripts/offline_store_spark_runner.py")
+
+	argList := []string{
```
sdreyer, line 745: Needs store_type arg

provider/spark.go
```diff
@@ -626,8 +630,170 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+		"--deploy-mode",
+		s.deployMode,
+		"--packages",
+		"\"org.apache.hadoop:hadoop-azure:3.2.0\"",
+		sparkScriptPath,
+		"df",
+		"--output_uri",
+		fmt.Sprintf("\"%s\"", outputURI),
+		"--code",
+		code, // TODO: Need to use the prefix
```
sdreyer, line 758: still todo?

provider/spark.go
sdreyer: Actually don't worry about the store_type arg, I changed how it works in the hdfs branch

provider/spark.go
```diff
@@ -968,7 +1141,7 @@ func (d *DatabricksExecutor) GetDFArgs(outputURI string, code string, sources []
 		"--output_uri",
 		outputURI,
 		"--code",
-		code,
+		code, // TODO: Need to use the prefix
```
sdreyer, line 1144: still todo?

provider/spark.go
ahmadnazeri: yes

client/src/featureform/resources.py
```diff
@@ -1403,13 +1408,35 @@ class SparkCredentials:
...
+            major, minor, _ = version.split(".")
+        elif version.count(".") == 1:
+            major, minor = version.split(".")    
+        else:
+            raise Exception("Please specify your Python version on the Spark cluster. Accepted formats: Major.Minor or Major.Minor.Patch; ex. '3.7' or '3.7.16")
+
+        if major != MAJOR_VERSION or minor not in MINOR_VERSIONS:
+            raise Exception(f"The Python version {version} is not supported. Currently, supported versions are 3.7-3.10.")
+        
+        if minor == "10":
```
simba-git, line 1434: can you comment why this is here? Seems arbritrary

client/src/featureform/resources.py
```diff
@@ -1403,13 +1408,35 @@ class SparkCredentials:
     def __init__(self,
                  master: str,
                  deploy_mode: str,
+                 python_version: str,
                 ):
 
         if deploy_mode != "cluster" and deploy_mode != "client":
             raise Exception(f"Spark does not support '{deploy_mode}' deploy mode. It only supports 'cluster' and 'client'.")
 
+        python_version = self._verify_python_version(python_version)
```
simba-git, line 1417: Isn't this only necessary for client mode? Couldn't we get this info from the spark cluster? Could we make this default to the user's current version?

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+
+	err := readAndUploadFile(sparkScriptPath, sparkScriptPathWithPrefix, store)
+	scriptExists, _ := store.Exists(sparkScriptPathWithPrefix)
+	if err != nil && !scriptExists {
+		return fmt.Errorf("could not upload spark script: Path: %s, Error: %v", sparkScriptPathWithPrefix, err)
+	}
+	return nil
+}
+
+func (s *SparkGenericExecutor) RunSparkJob(args *[]string, store FileStore) error {
```
simba-git, line 679: Why is this a pointer to a slice?

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	bashCommand := "bash"
+	sparkArgs := *args
+	sparkArgsString := strings.Join(sparkArgs, " ")
+	bashCommandArgs := []string{"-c", fmt.Sprintf("pyenv global %s && pyenv exec %s", s.pythonVersion, sparkArgsString)}
```
simba-git, line 683: This is fine but in the future you don't have to and shouldn't wrap this with bash. You can run pyenv directly

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	err = cmd.Wait()
+	if err != nil {
+		return fmt.Errorf("spark job failed: %v", err)
+	}
+
+	return nil
+}
+
+func (s *SparkGenericExecutor) PythonFileURI(store FileStore) string {
+	return ""
```
simba-git, line 703: Can you comment why this is?

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	sparkScriptPath := helpers.GetEnv("SPARK_SCRIPT_PATH", "/scripts/offline_store_spark_runner.py")
+
+	argList := []string{
+		"spark-submit",
+		"--master",
+		s.master,
+		"--deploy-mode",
+		s.deployMode,
+		"--packages",
+		"\"org.apache.hadoop:hadoop-azure:3.2.0\"",
```
simba-git, line 716: ?

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+		"--output_uri",
+		fmt.Sprintf("\"%s\"", outputURI),
+		"--code",
+		code,
+	}
+
+	var remoteConnectionArgs []string
+	azureStore := store.AsAzureStore()
+
+	if azureStore != nil {
```
simba-git, line 766: What is this doing?

client/src/featureform/resources.py
ahmadnazeri: So, on our docker image, we support 3.7 - 3.10 and this value will be used for `pyenv global {major}.{minor}.{patch}`. I can add a comment in the code too?

provider/spark.go
ahmadnazeri: No particular reason except that the interface was designed this way. I can change it so it accepts `[]string`

provider/spark.go
ahmadnazeri: This is part of the interface and method is used for Databricks to get the full path for the Python Script.

provider/spark.go
ahmadnazeri: needed for reading data from azure blob store.

provider/spark.go
ahmadnazeri: checks to see if azureStore in order to add credentials for it which will be used to pull the df pickle.

client/src/featureform/resources.py
ahmadnazeri: yes, only needed it for client mode. We will need access to spark cluster in order to determine the type. The user's current version can be different from the cluster version.

provider/spark.go
simba-git: Could we write the code to make this cleaner?
Maybe add an IsAzureStore method?

provider/spark.go
simba-git: I know, I mean put a comment in here. Also shows that the abstraction is broken, will want to fix at some point

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+
+	var remoteConnectionArgs []string
+	azureStore := store.AsAzureStore()
+	if azureStore != nil {
+		remoteConnectionArgs = []string{
+			"--spark_config",
+			fmt.Sprintf("\"%s\"", azureStore.configString()),
+		}
+	}
+	argList = append(argList, remoteConnectionArgs...)
```
simba-git, line 735: Why not put the append in the if statemnet above

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+		"--output_uri",
+		fmt.Sprintf("\"%s\"", destPath),
+		"--sql_query",
+		fmt.Sprintf("\"%s\"", cleanQuery),
+		"--job_type",
+		fmt.Sprintf("\"%s\"", string(jobType)),
+	}
+
+	var remoteConnectionArgs []string
+	azureStore := store.AsAzureStore()
```
simba-git, line 728: Seems weird that AsAzureStore would just return null if its not possible. I'd expect it to have two parameters:
Store, bool or Store, error or something

provider/spark.go
simba-git: But why add it to the main arg list why not add it to the ones you do if its azure

provider/spark.go
```diff
@@ -626,8 +630,172 @@ func (e EMRExecutor) InitializeExecutor(store FileStore) error {
...
+	sparkScriptPath := helpers.GetEnv("SPARK_SCRIPT_PATH", "/scripts/offline_store_spark_runner.py")
+
+	argList := []string{
+		"spark-submit",
+		"--master",
+		s.master,
+		"--deploy-mode",
+		s.deployMode,
+		"--packages",
+		"\"org.apache.hadoop:hadoop-azure:3.2.0\"",
```
simba-git, line 754: put this in the if statement if its only used by azure

provider/spark.go
simba-git: Can we abstract this into its own interface of struct.

Example: have a new struct called
SparkArgs
- SubmitArgs
- DFArgs

Have two fns:
genericSparkArgs(...)
azureSparkArgs(...)

Then you can do a type switch on the store here.

You don't have to do this, but I personally think this would be much cleaner and extensible that what you did here.

client/src/featureform/resources.py
simba-git: yes please

client/src/featureform/resources.py
simba-git: I know, it just seems a bit weird to have to add the python version of spark as a parameter. I was saying that we could default to the client's current version, if they don't specify

Is there no way to deduce the python version from Spark itself on the backend

provider/spark.go
sdreyer: I removed this in my branch

client/src/featureform/resources.py
ahmadnazeri: So we can make it _required_ if the run it in client mode. Even then if they don't specify it, we can make pick their python version on the client.

provider/spark.go
simba-git: A slice pointer is super weird, I'd change it if its easy to do so

simba-git: Can also leave it to a separate PR for next release if prefered let me know

ahmadnazeri: so databricks executor uses it and I believe that's why we made into a pointer. we will be working on databricks next so we will address it in that PR.

sdreyer (approved):

PR #630, opened 2023-02-13
Model Metadata Upsert and Registration w/ Serving

epps:
# Description

This PR introduces the upsert pattern for resources, with the model resource as an initial example, as well as the implementation for model registration while serving features or training sets.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/metadata/client.go
+++ b/metadata/client.go
@@ -1051,7 +1051,6 @@ func (def ModelDef) ResourceType() ResourceType {
 func (client *Client) CreateModel(ctx context.Context, def ModelDef) error {
 	serialized := &pb.Model{
 		Name:         def.Name,
-		Status:       &pb.ResourceStatus{Status: pb.ResourceStatus_NO_STATUS},
 		Description:  def.Description,
 		Features:     def.Features.Serialize(),
 		Trainingsets: def.Trainingsets.Serialize(),
@@ -1467,16 +1466,10 @@ func (model *Model) Description() string {
 }
 
 func (model *Model) Status() ResourceStatus {
-	if model.serialized.GetStatus() != nil {
-		return ResourceStatus(model.serialized.GetStatus().Status)
-	}
 	return ResourceStatus(0)
 }
 
 func (model *Model) Error() string {
-	if model.serialized.GetStatus() != nil {
-		return model.serialized.GetStatus().ErrorMessage
-	}
 	return ""
 }
 
--- a/metadata/client.go
+++ b/metadata/client.go
@@ -1038,8 +1038,10 @@ func (client *Client) GetModels(ctx context.Context, models []string) ([]*Model,
 }
 
 type ModelDef struct {
-	Name        string
-	Description string
+	Name         string
+	Description  string
+	Features     NameVariants
+	Trainingsets NameVariants
 }
 
 func (def ModelDef) ResourceType() ResourceType {
@@ -1048,9 +1050,11 @@ func (def ModelDef) ResourceType() ResourceType {
 
 func (client *Client) CreateModel(ctx context.Context, def ModelDef) error {
 	serialized := &pb.Model{
-		Name:        def.Name,
-		Status:      &pb.ResourceStatus{Status: pb.ResourceStatus_NO_STATUS},
-		Description: def.Description,
+		Name:         def.Name,
+		Status:       &pb.ResourceStatus{Status: pb.ResourceStatus_NO_STATUS},
+		Description:  def.Description,
+		Features:     def.Features.Serialize(),
+		Trainingsets: def.Trainingsets.Serialize(),
 	}
 	_, err := client.grpcConn.CreateModel(ctx, serialized)
 	return err
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -317,6 +317,11 @@ func (lookup localResourceLookup) HasJob(id ResourceID) (bool, error) {
 	return false, nil
 }
 
+type NameVariantKey struct {
+	Name    string
+	Variant string
+}
+
 type sourceResource struct {
 	serialized *pb.Source
 }
@@ -888,14 +893,16 @@ func (resource *modelResource) Update(lookup ResourceLookup, updateRes Resource)
 }
 
 func unionNameVariants(source, destination []*pb.NameVariant) ([]*pb.NameVariant, error) {
-	set := make(map[string]bool)
+	set := make(map[NameVariantKey]bool)
 
 	for _, nameVariant := range destination {
-		set[fmt.Sprintf("%s|%s", nameVariant.Name, nameVariant.Variant)] = true
+		key := NameVariantKey{Name: nameVariant.Name, Variant: nameVariant.Variant}
+		set[key] = true
 	}
 
 	for _, nameVariant := range source {
-		if _, has := set[fmt.Sprintf("%s|%s", nameVariant.Name, nameVariant.Variant)]; !has {
+		key := NameVariantKey{Name: nameVariant.Name, Variant: nameVariant.Variant}
+		if _, has := set[key]; !has {
 			destination = append(destination, nameVariant)
 		}
 	}
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -193,7 +193,7 @@ type Resource interface {
 	Proto() proto.Message
 	UpdateStatus(pb.ResourceStatus) error
 	UpdateSchedule(string) error
-	Update(ResourceLookup) error
+	Update(ResourceLookup, Resource) error
 }
 
 func isDirectDependency(lookup ResourceLookup, dependency, parent Resource) (bool, error) {
@@ -359,8 +359,8 @@ func (resource *sourceResource) UpdateSchedule(schedule string) error {
 	return fmt.Errorf("not implemented")
 }
 
-func (resource *sourceResource) Update(lookup ResourceLookup) error {
-	return &ResourceExists{resource.ID()}
+func (resource *sourceResource) Update(lookup ResourceLookup, updateRes Resource) error {
+	return &ResourceExists{updateRes.ID()}
 }
 
 type sourceVariantResource struct {
@@ -433,8 +433,8 @@ func (resource *sourceVariantResource) UpdateSchedule(schedule string) error {
 	return nil
 }
 
-func (resource *sourceVariantResource) Update(lookup ResourceLookup) error {
-	return &ResourceExists{resource.ID()}
+func (resource *sourceVariantResource) Update(lookup ResourceLookup, updateRes Resource) error {
+	return &ResourceExists{updateRes.ID()}
 }
 
 type featureResource struct {
@@ -479,8 +479,8 @@ func (resource *featureResource) UpdateSchedule(schedule string) error {
 	return fmt.Errorf("not implemented")
 }
 
-func (resource *featureResource) Update(lookup ResourceLookup) error {
-	return &ResourceExists{resource.ID()}
+func (resource *featureResource) Update(lookup ResourceLookup, updateRes Resource) error {
+	return &ResourceExists{updateRes.ID()}
 }
 
 type featureVariantResource struct {
@@ -557,8 +557,8 @@ func (resource *featureVariantResource) UpdateSchedule(schedule string) error {
 	return nil
 }
 
-func (resource *featureVariantResource) Update(lookup ResourceLookup) error {
-	return &ResourceExists{resource.ID()}
+func (resource *featureVariantResource) Update(lookup ResourceLookup, updateRes Resource) error {
+	return &ResourceExists{updateRes.ID()}
 }
 
 type labelResource struct {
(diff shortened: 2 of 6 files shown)


metadata/metadata.go
```diff
@@ -192,6 +193,7 @@ type Resource interface {
 	Proto() proto.Message
 	UpdateStatus(pb.ResourceStatus) error
 	UpdateSchedule(string) error
+	Update(ResourceLookup) error
```
simba-git, line 196: Update should take a ResourceLookup and a Resource, we need the resource to be updated with.

metadata/metadata.go
```diff
@@ -832,6 +866,45 @@ func (resource *modelResource) UpdateSchedule(schedule string) error {
 	return fmt.Errorf("not implemented")
 }
 
+func (resource *modelResource) Update(lookup ResourceLookup) error {
+	val, err := lookup.Lookup(resource.ID())
```
simba-git, line 870: This won't work since the model you lookup will be the same one you currently have

metadata/metadata.go
```diff
@@ -832,6 +866,45 @@ func (resource *modelResource) UpdateSchedule(schedule string) error {
...
+		return err
+	}
+	resource.serialized.Trainingsets = updated
+	return nil
+}
+
+func unionNameVariants(source, destination []*pb.NameVariant) ([]*pb.NameVariant, error) {
+	set := make(map[string]bool)
+
+	for _, variant := range destination {
```
simba-git, line 895: this is only using the variants not the names, you could make or us a struct NameVariant{name string, variant string} and make set := make(map[NameVariant]bool)

metadata/client.go
```diff
@@ -1048,9 +1050,11 @@ func (def ModelDef) ResourceType() ResourceType {
 
 func (client *Client) CreateModel(ctx context.Context, def ModelDef) error {
 	serialized := &pb.Model{
-		Name:        def.Name,
-		Status:      &pb.ResourceStatus{Status: pb.ResourceStatus_NO_STATUS},
-		Description: def.Description,
+		Name:         def.Name,
+		Status:       &pb.ResourceStatus{Status: pb.ResourceStatus_NO_STATUS},
```
sdreyer, line 1054: Feel like we should just remove the status field entirely @simba-git ?

metadata/metadata_test.go
```diff
@@ -1243,9 +1277,102 @@ func expectedModels() ResourceTests {
...
+			Features:    []NameVariant{},
+			TrainingSets: []NameVariant{
+				{Name: "training-set", Variant: "variant"},
+				{Name: "training-set", Variant: "variant2"},
+			},
+		},
+	}
+}
+
+func modelUpdates() []ResourceDef {
```
sdreyer, line 1314: Should add some cases like adding a trainingset or feature twice

metadata/metadata.go
```diff
@@ -832,6 +866,45 @@ func (resource *modelResource) UpdateSchedule(schedule string) error {
...
+	resource.serialized.Features = updated
+	updated, err = unionNameVariants(resource.serialized.Trainingsets, model.Trainingsets)
+	if err != nil {
+		return err
+	}
+	resource.serialized.Trainingsets = updated
+	return nil
+}
+
+func unionNameVariants(source, destination []*pb.NameVariant) ([]*pb.NameVariant, error) {
```
sdreyer, line 892: Should make a test for this function that tests duplicates, different name variants, nil values, etc

metadata/metadata.go
epps: I could be grossly misunderstanding something here, but if `resource` is the object in memory on which we're calling `Update`, and we already know that a record by the same key exists, isn't the `val` going to be different from `resource`?

metadata/metadata.go
epps: Is this likely due to a misunderstanding on my end, but my thought here was:
* `res` is the object in memory from the request (e.g. `res Resource` in `genericCreate`)
* if we now a record exists in etcd with the key `res.ID()`, then calling `res.Update(serv.lookup)` allows the `Update` method to fetch that existing record and merge the two according to whatever rules we have for the specific resource
* after the call to `Update` completes without err, we know `res` now represents the updated value for the resource, which is set back to etcd

metadata/metadata.go
epps: I originally had the make use a `NameVariant` has a key, but it seems that the comparison for the map isn't performed by value, so I was never seeing updates.

metadata/metadata_test.go
epps: Maybe I should comment this function to make it clearer, but I iterate over these items and call `CreateModel` for each, so the 2nd and the 3rd item do test for adding the same training set twice. Would you like me to add the same cases for features, too?

metadata/metadata.go
epps: @simba-git, never mind: I chatted with @sdreyer about this and I now see the assumption I was making about the `Update` method (i.e. that `res` would always represent the inbound request message and not also (possibly) the existing record in etcd). I'll update the PR to reflect this new understanding.

metadata/metadata.go
simba-git: I see, would rename variant to nameVariant in this loop and the next loop, I got confused

metadata/metadata.go
```diff
@@ -824,14 +858,51 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
+		return err
+	}
+	return nil
+}
+
+func unionNameVariants(source, destination []*pb.NameVariant) ([]*pb.NameVariant, error) {
+	set := make(map[string]bool)
+
+	for _, nameVariant := range destination {
+		set[fmt.Sprintf("%s|%s", nameVariant.Name, nameVariant.Variant)] = true
```
epps, line 894: FYI, @simba-git, after taking a closer look a the proto's `String` method, I noticed a comment about not relying on the output, which _could_ change from version to version:
Screen Shot 2023-02-14 at 9 19 25 AM
I chose to handle the string formatting explicitly myself, but if there's a better pattern I can follow, let me know.

simba-git: I would make a struct and use that

serving/serving.go
```diff
@@ -33,13 +34,20 @@ func NewFeatureServer(meta *metadata.Client, promMetrics metrics.MetricsHandler,
 	}, nil
 }
 
-func (serv *FeatureServer) TrainingData(req *pb.TrainingDataRequest, stream pb.Feature_TrainingDataServer) error {
+func (serv *FeatureServer) TrainingData(ctx context.Context, req *pb.TrainingDataRequest, stream pb.Feature_TrainingDataServer) error {
```
simba-git, line 37: I think the last argument has a context method

metadata/metadata.go
```diff
@@ -315,6 +317,11 @@ func (lookup localResourceLookup) HasJob(id ResourceID) (bool, error) {
 	return false, nil
 }
 
+type NameVariantKey struct {
+	Name    string
+	Variant string
+}
+
```
epps, line 324: @simba-git, let me know if you'd like to change the naming here.

metadata/metadata_test.go
```diff
@@ -1243,9 +1277,112 @@ func expectedModels() ResourceTests {
...
+/*
+Currently, the testing pattern assumes immutability, which made writing
+test for model updates a bit awkward. As we roll out updates to other
+resource types, we should consider refactoring the top-level interfaces
+so we can more neatly encapsulate data/logic/etc. for updates.
+
+For now, the below two functions work in tandem:
+* `modelUpdates` holds 3 payloads that are applied in order
+* `expectedUpdatedModels` holds the expected state after each payload is persisted
+*/
```
epps, line 1289: @sdreyer, does this encapsulate what we discussed?

metadata/metadata.go
simba-git: Would make this lowercased so its not externalized outside the package (in Go, lowercase = private). You could also put this struct in the method where it's used to further limit the scope

metadata/metadata.go
```diff
@@ -824,14 +863,53 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
+		return err
+	}
+	resource.serialized.Trainingsets = updated
+	if err = lookup.Set(resource.ID(), resource); err != nil {
+		return err
+	}
+	return nil
+}
+
+func unionNameVariants(source, destination []*pb.NameVariant) ([]*pb.NameVariant, error) {
```
simba-git, line 895: this never returns an error, can make it just return the array

metadata/metadata.go
```diff
@@ -824,14 +863,53 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
 	return fmt.Errorf("not implemented")
 }
 
+func (resource *modelResource) Update(lookup ResourceLookup, updateRes Resource) error {
+	deserialized := updateRes.Proto()
+	updateModel, ok := deserialized.(*pb.Model)
+	if !ok {
+		return errors.New("failed to deserialize existing model record")
+	}
+	updated, err := unionNameVariants(resource.serialized.Features, updateModel.Features)
```
simba-git, line 879: since this never returns an error you could make it:
resource.serialized.Features = unionNameVariants(...

metadata/metadata.go
```diff
@@ -1331,7 +1421,13 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	if has, err := serv.lookup.Has(id); err != nil {
 		return nil, err
 	} else if has {
-		return nil, &ResourceExists{id}
+		existing, err := serv.lookup.Lookup(id)
```
simba-git, line 1424: I would switch the serv.lookup.Has above into serv.lookup.Lookup
With how this is coded up now, we end up having to do two lookups in the worst case for no reason.
You can check the error of the lookup to make it look like the lookup.Has

metadata/metadata.go
```diff
@@ -1331,7 +1421,13 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
...
+		if err != nil {
+			return nil, err
+		}
+		if err := res.Update(serv.lookup, existing); err != nil {
+			return nil, err
+		}
 	}
 	if err := serv.lookup.Set(id, res); err != nil {
```
simba-git, line 1432: You are setting it in the Update and here. Could remove the Set from the Update.

metadata/metadata.go
```diff
@@ -824,14 +863,53 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
+	if err != nil {
+		return err
+	}
+	resource.serialized.Features = updated
+	updated, err = unionNameVariants(resource.serialized.Trainingsets, updateModel.Trainingsets)
+	if err != nil {
+		return err
+	}
+	resource.serialized.Trainingsets = updated
+	if err = lookup.Set(resource.ID(), resource); err != nil {
```
simba-git, line 889: This can be removed since its done in the genericCreate already

simba-git (commented):
Excited to get this in! A few Go specific comments and some unnecessary operations

metadata/metadata.go
```diff
@@ -824,14 +858,49 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
 	return fmt.Errorf("not implemented")
 }
 
+func (resource *modelResource) Update(lookup ResourceLookup, updateRes Resource) error {
+	deserialized := updateRes.Proto()
+	updateModel, ok := deserialized.(*pb.Model)
+	if !ok {
+		return errors.New("failed to deserialize existing model record")
+	}
+	updated := unionNameVariants(resource.serialized.Features, updateModel.Features)
```
simba-git, line 874: you can combine this and the line below if you'd like, keeps you from re-using updated

metadata/metadata.go
```diff
@@ -824,14 +858,49 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
+func unionNameVariants(source, destination []*pb.NameVariant) []*pb.NameVariant {
+	type nameVariantKey struct {
+		Name    string
+		Variant string
+	}
+
+	set := make(map[nameVariantKey]bool)
+
+	for _, nameVariant := range destination {
+		key := nameVariantKey{Name: nameVariant.Name, Variant: nameVariant.Variant}
```
simba-git, line 890: you can write this as:
key := nameVariantKey{nameVariant.Name, nameVariant.Variant}
more typical in Go when you have a simple struct like this and fields are implied

metadata/metadata.go
```diff
@@ -824,14 +858,49 @@ func (this *modelResource) Notify(lookup ResourceLookup, op operation, that Reso
...
+		set[key] = true
+	}
+
+	for _, nameVariant := range source {
+		key := nameVariantKey{Name: nameVariant.Name, Variant: nameVariant.Variant}
```
simba-git, line 895: same as above

simba-git (approved):
Two super minor nitpicks, you can ignore or implement up to you

Otherwise LGTM

PR #637, opened 2023-02-17
Model Registration: Hosted Client Implementation

epps:
# Description

This PR introduces changes to the Python client, as well as changes to the structure of the E2E tests after discussion around the global registrar, for the Python client implementation for model registration in hosted mode.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -2582,7 +2582,14 @@ class ResourceClient(Registrar):
         Returns:
             model (Model): Model
         """
-        model = get_model_info_local(name) if local else get_resource_info(self._stub, "model", name)
+        if local:
+            model = get_model_info_local(name)
+        else:
+            model_proto = get_resource_info(self._stub, "model", name)
+            if model_proto is None:
+                model = None
+            else:
+                model = Model(model_proto.name)
 
         return model
 
@@ -3479,7 +3486,8 @@ class ResourceClient(Registrar):
             rows = list_local("model", [ColumnName.NAME])
             models = [Model(row["name"]) for row in rows]
         else:
-            models = list_name_status_desc(self._stub, "model")
+            model_protos = list_name(self._stub, "model")
+            models = [Model(proto.name) for proto in model_protos]
 
         return models
 
--- a/client/tests/conftest.py
+++ b/client/tests/conftest.py
@@ -149,24 +149,24 @@ def s3(aws_credentials):
 
 @pytest.fixture(scope="module")
 def local_provider_source():
-        def get_local():
-            ff.register_user("test_user").make_default_owner()
-            provider = ff.register_local()
-            source = provider.register_file(
-                name="transactions",
-                variant="quickstart",
-                description="A dataset of fraudulent transactions.",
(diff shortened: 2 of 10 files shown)


client/tests/conftest.py
```diff
@@ -149,24 +149,24 @@ def s3(aws_credentials):
...
-            source = provider.register_file(
-                name="transactions",
-                variant="quickstart",
-                description="A dataset of fraudulent transactions.",
-                path=f"{dir_path}/test_files/input_files/transactions.csv"
-            )
-            return (provider, source)
-        
-        return get_local
+    def get_local(is_insecure):
```
simba-git, line 152: Is this parameter used?

client/src/featureform/register.py
```diff
@@ -2582,7 +2582,11 @@ def get_model(self, name, local=False) -> Model:
         Returns:
             model (Model): Model
         """
-        model = get_model_info_local(name) if local else get_resource_info(self._stub, "model", name)
+        if local:
+            model = get_model_info_local(name)
+        else:
+            model_proto = get_resource_info(self._stub, "model", name)
+            model = Model(model_proto.name) if model_proto is not None else model_proto
```
simba-git, line 2589: Should this return the model proto or fail?

client/tests/conftest.py
epps: It's not, but given the hosted fixture requires it, I had to pass it to the local counterpart. Is there an idiomatic way of ignoring a parameter in Python?

epps: It is. I needed to pass a param to the local fixture given the hosted fixture requires it. I'll convert this to an `_` to better convey the fact I don't intend to use it.

client/src/featureform/register.py
epps: The pattern with the other resource get methods was to print out an message to the console in the event of a gRPC error. In these cases, the implicit return type is `None` so I tried to maintain that API here.

client/tests/conftest.py
```diff
@@ -149,24 +149,24 @@ def s3(aws_credentials):
...
-            source = provider.register_file(
-                name="transactions",
-                variant="quickstart",
-                description="A dataset of fraudulent transactions.",
-                path=f"{dir_path}/test_files/input_files/transactions.csv"
-            )
-            return (provider, source)
-        
-        return get_local
+    def get_local(_):
```
sdreyer, line 152: Is this underscore supposed to be here?

client/src/featureform/register.py
sdreyer: I would just do `else None` for clarity for now. I blocked off time on my calendar this week to review that lingering PR that would remove this weird logic

client/tests/conftest.py
epps: Yes, previously, it was a named param that I don't actually need for the local fixture, but to ensure I had the same function signature for both local and hosted, I passed the param here, but changed it to an underscore to convey the fact I didn't intend to use it.

sdreyer (approved):

simba-git (approved):

PR #641, opened 2023-02-23
Mutable Providers: E2E Tests

epps:
# Description

This PR adds E2E tests for valid and invalid providers updates.

### Select type(s) of change

- [ ] Documentation update

# Checklist:


client/tests/test_updating_provider.py
```diff
@@ -0,0 +1,153 @@
...
+
+    if is_local:
+        postgres_host = "0.0.0.0"
+        redis_host = "0.0.0.0"
+    else:
+        # The host name for postgres is different between Docker and Minikube
+        postgres_host = "host.docker.internal" if "docker" in custom_marks else "quickstart-postgres"
+        redis_host = "host.docker.internal" if "docker" in custom_marks else "quickstart-redis"
+
+    postgres = ff.register_postgres(
```
sdreyer, line 40: What does this do in localmode? Does it actually store a postgres config?

client/tests/test_updating_provider.py
```diff
@@ -0,0 +1,153 @@
...
+
+    redis = ff.register_redis(
+        name=redis_name,
+        host=redis_host,
+        port=6379,
+        password=redis_password,
+        description=redis_description
+    )
+
+    resource_client.apply()
```
sdreyer, line 58: Feel like this will likely always error in localmode right?

client/tests/test_updating_provider.py
```diff
@@ -0,0 +1,153 @@
...
+        description = "A Postgres deployment we created for the Featureform quickstart"
+    )
+
+    redis = ff.register_redis(
+        name=redis_name,
+        host="updated-quickstart-redis",
+        port=6379,
+    )
+
+    with pytest.raises(ResourceRedefinedError):
```
sdreyer, line 114: I dont think this will raise a resourceredefinederror. I believe it will print something like "Resource Already Exists"

client/tests/test_updating_provider.py
epps: It does. I was skeptical, but I tried it with the local mode quickstart in a separate repo and checked the SQLite DB to ensure it was in the `providers` table.

client/tests/test_updating_provider.py
epps: Due to the postgres/redis providers or another reason?

sdreyer: nm was thinking about serving

client/tests/test_updating_provider.py
sdreyer: bump

epps: You're right. The backend returns the error, but the client swallows and prints it. Let me see how I can adapt the test to account for this. @simba-git pointed out we should really return a different type of error.

sdreyer (approved):

PR #643, opened 2023-02-27
Provider Configs: Refactor to Pkg and Add Mutable Field Methods

epps:
# Description

This PR refactors provider configs to a separate package to avoid a future circular dependency between the metadata and provider packages. Additionally, methods for determining a config's mutable fields and determining a diff between two configs of the same type. New unit tests to cover these config methods are also included.

### Select type(s) of change

- [ ] Documentation update

# Checklist:


helpers/string_set/string_set.go
```diff
@@ -0,0 +1,12 @@
+package string_set
+
+type StringSet map[string]bool
+
+func (a StringSet) Contains(b StringSet) bool {
```
sdreyer, line 5: Can you put a doc string on this so we have hints in the IDE when importing?

provider/provider_config/provider_config.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	bv := reflect.ValueOf(b)
+
+	for aIter.Next() {
+		key := aIter.Key()
+		aVal := aIter.Value()
+		bVal := bv.FieldByName(key).Interface()
```
sdreyer, line 37: You could add this to the struct iterator interface or make a function for it

provider/offline.go
```diff
@@ -8,23 +8,25 @@ import (
...
 
 const (
 	MemoryOffline    Type = "MEMORY_OFFLINE"
-	PostgresOffline       = "POSTGRES_OFFLINE"
-	SnowflakeOffline      = "SNOWFLAKE_OFFLINE"
-	RedshiftOffline       = "REDSHIFT_OFFLINE"
-	SparkOffline          = "SPARK_OFFLINE"
-	BigQueryOffline       = "BIGQUERY_OFFLINE"
-	K8sOffline            = "K8S_OFFLINE"
+	PostgresOffline  Type = "POSTGRES_OFFLINE"
```
simba-git, line 24: these aren't required

simba-git: By these I mean adding the Type, it was fine before

simba-git: nbd if you want to keep it this way, just an FYI

simba-git (approved):

provider/provider_config/provider_config.go
simba-git: Yeah I mentioned making a:
StructMap which is like a struct iterator but allows you to get fields by name, also keeps all the reflect in one package which is preferable

provider/offline.go
epps: I looked into this and thought that it might just be an issue with my language server in VS Code, but after reading [this](https://stackoverflow.com/questions/64826695/set-type-for-entire-const-block) it appears if we want the alias to apply to all of the variants, we need to specify it on each line or use a single-line notation.

simba-git: Today I learned

provider/provider_config/dynamo_config.go
```diff
@@ -0,0 +1,42 @@
...
+	err := json.Unmarshal(config, d)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d DynamodbConfig) MutableFields() ss.StringSet {
+	return ss.StringSet{
+		"Region":    true,
```
epps, line 34: @simba-git and/or @sdreyer, now that I'm looking at this again, I'm not sure it makes sense to allow `Region` to change. My understanding is that you would have to actually replicate data to another instance in the new region, so this would likely be a new provider from our perspective, right?

provider/provider_config/spark_config.go
```diff
@@ -0,0 +1,217 @@
...
+	for field, val := range executor {
+		result["Executor."+field] = val
+	}
+
+	for field, val := range store {
+		result["Store."+field] = val
+	}
+
+	return result, err
+}
```
epps, line 179: @sdreyer, @simba-git, and @ahmadnazeri, let me know if this matches your expectations for a valid spark config update.

provider/provider_config/spark_config.go
```diff
@@ -0,0 +1,217 @@
...
+	return result
+}
+
+func (a SparkConfig) DifferingFields(b SparkConfig) (ss.StringSet, error) {
+	result := ss.StringSet{}
+	var executor ss.StringSet
+	var store ss.StringSet
+	var err error
+
+	switch a.ExecutorType {
```
sdreyer, line 126: Create a default and throw an error for these switches

provider/provider_config/spark_config.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+	}
+
+	if err != nil {
+		return result, err
+	}
+
+	switch a.StoreType {
+	case Azure:
+		if b.StoreType != Azure {
```
sdreyer, line 147: you can pull this out of the switch and just do `if a.StoreType != b.StoreType`

provider/provider_config/spark_config.go
```diff
@@ -0,0 +1,217 @@
...
+	var executor ss.StringSet
+	var store ss.StringSet
+	var err error
+
+	switch a.ExecutorType {
+	case EMR:
+		if b.ExecutorType != EMR {
+			err = fmt.Errorf("executor config mismatch: a = %v; b = %v", a.ExecutorType, b.ExecutorType)
+		} else {
+			executor, err = a.ExecutorConfig.(*EMRConfig).DifferingFields(*b.ExecutorConfig.(*EMRConfig))
```
sdreyer, line 131: Change this to executorFields (and storeFields below). This looks like its returning an executor

provider/provider_config/spark_config.go
sdreyer: I think the logic makes sense

provider/provider_config/spark_config.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return nil
+}
+
+func (s SparkConfig) MutableFields() ss.StringSet {
+	result := ss.StringSet{}
+	var executorFields ss.StringSet
+	var storeFields ss.StringSet
+
+	switch s.ExecutorType {
```
sdreyer, line 93: Add defaults

sdreyer (approved):

PR #655, opened 2023-03-04
adds hosted search implementation

epps:
# Description

This PR implements hosted search in the Featureform CLL.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/search.py
+++ b/client/src/featureform/search.py
@@ -3,11 +3,7 @@ import requests
 from .format import *
 
 
-def search(phrase):
-    host = os.getenv('FEATUREFORM_HOST')
-    if host is None:
-        raise RuntimeError("Host not found.")
-    
+def search(phrase, host):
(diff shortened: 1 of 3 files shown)


client/src/featureform/search.py
```diff
@@ -0,0 +1,26 @@
+import os
+import requests
+from .format import *
+
+
+def search(phrase):
+    host = os.getenv('FEATUREFORM_HOST')
```
epps, line 7: @sdreyer, there is a possible issue here that is somewhat novel: Given all other methods in the resource client use gRPC, which means they rely on the check in the constructor and then `self._stub`, we don't save the value of `host` on the RC instance. If a user is simply passing the `--host` flag on the command line, this won't work. They need to export the value in the session instead.

Are there any objections to storing `host` on the RC instance?

sdreyer: Storing it is fine

PR #679, opened 2023-03-11
Tags & Properties: Local Implementation + E2E tests

epps:
# Description

This PR introduces the ability to add tags (i.e. `List[str]`) and properties (i.e. `dict`) to resources. Additionally, users may modify the tags/properties on a previously registered resource (i.e. append only).

### Select type(s) of change

- [x] Documentation update

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,6 +222,7 @@ pytest:
 	pytest -m 'local' client/tests/test_serving_model.py
 	pytest -m 'local' client/tests/test_getting_model.py
 	pytest -m 'local' client/tests/test_search.py
+	pytest -m 'local' client/tests/test_tags_and_properties.py
 	-rm -r .featureform
 	-rm -f transactions.csv
 
--- a/client/src/featureform/dashboard_metadata.py
+++ b/client/src/featureform/dashboard_metadata.py
@@ -80,7 +80,9 @@ def feature_variant(variantData):
                 "value": variantRow['source_value'], 
                 "timestamp": variantRow['source_timestamp']},
                 {"Name":variantRow['source_name'],
-                "Variant":variantRow['source_variant']} 
+                "Variant":variantRow['source_variant']},
+                json.loads(variantRow['tags']) if variantRow['tags'] is not None else [],
+                json.loads(variantRow['properties']) if variantRow['properties'] is not None else {}
             ).toDictionary()
 
         allVariantList.append(variantRow['variant'])
@@ -89,7 +91,7 @@ def feature_variant(variantData):
     return variantsDict, allVariantList, variants
 
 def features(featureRow):
-    variantData = feature_variant(sqlObject.query_resource("feature_variant", "name",featureRow['name']))
+    variantData = feature_variant(sqlObject.query_resource_variant("feature_variant", "name", featureRow['name']))
     
     return FeatureResource(
                 featureRow['name'], 
@@ -119,6 +121,8 @@ def training_set_variant(variantData):
                 "Variant":variantRow['label_variant']}, 
                 variantRow['status'], 
                 variant_organiser(feature_variant(getTrainingSetFeatures(feature_list))[2]),
+                json.loads(variantRow['tags']) if variantRow['tags'] is not None else [],
+                json.loads(variantRow['properties']) if variantRow['properties'] is not None else {}
             ).toDictionary()
         allVariantList.append(variantRow['variant'])
         variantDict[variantRow['variant']] = trainingSetVariant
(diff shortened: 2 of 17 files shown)


Makefile
```diff
@@ -222,6 +222,7 @@ pytest:
 	pytest -m 'local' client/tests/test_serving_model.py
 	pytest -m 'local' client/tests/test_getting_model.py
 	pytest -m 'local' client/tests/test_search.py
+	pytest -m 'loacl' client/tests/test_tags_and_properties.py
```
sdreyer, line 225: Spelling

client/src/featureform/sqlite_metadata.py
```diff
@@ -300,96 +342,154 @@ def fetch_data_safe(self, query, type, name, variant):
           return []
         return variant_data_list
 
-    def get_user(self, name):
-      return self.query_resource("users", "name", name)[0]
+    def get_user(self, name, should_fetch_tags_properties):
```
sdreyer, line 345: why do we need this flag?

client/src/featureform/sqlite_metadata.py
```diff
@@ -424,3 +538,29 @@ def search(self, phrase, resource_type=None):
...
+        is_update = False
+        name = args[0]
+        variant = args[1]
+
+        cursor = self.__conn.execute(f"SELECT * FROM {tablename} WHERE name = '{name}' AND variant = '{variant}';")
+        self.__conn.commit()
+        existing = cursor.fetchall()
+
+        if len(existing):
+          existing_val = json.loads(existing[0][updatable_column]) if existing[0][updatable_column] is not None else default_value
```
sdreyer, line 555: Should make const or assign existing[0] to a var instead of using a #. A lot of magic numbers in this function

client/src/featureform/sqlite_metadata.py
```diff
@@ -300,96 +342,154 @@ def fetch_data_safe(self, query, type, name, variant):
...
+        LEFT JOIN properties p ON p.name = v.name AND p.variant = v.variant
+        WHERE v.label_name = '{0}' AND v.label_variant = '{1}';
+        '''.format(name, variant)
         return self.fetch_data_safe(query, "training_set_variant", name, variant)
 
     def get_training_set_features(self, name, variant):
-        query = f"SELECT * FROM training_set_features WHERE training_set_name = '{name}' AND training_set_variant = '{variant}'"
+        query = '''SELECT v.*, t.tag_list as tags, p.property_list as properties
+        FROM training_set_features v
+        LEFT JOIN tags t ON t.name = v.training_set_name AND t.variant = v.training_set_variant
```
aolfat, line 470: just a thought, we could separate out the join str and the base string

epps: You mean into constants to reduce repetition?

aolfat: yeahhh, the motivation was all the repitition

epps: Not all these join conditions are the same, but you're right that I could save some repetition. Let me circle back around to this so I can tackle the hosted implementation first. I'll leave this comment unresolved in the meantime.

client/src/featureform/dashboard_metadata.py
```diff
@@ -119,6 +121,8 @@ def training_set_variant(variantData):
                 "Variant":variantRow['label_variant']}, 
                 variantRow['status'], 
                 variant_organiser(feature_variant(getTrainingSetFeatures(feature_list))[2]),
+                json.loads(variantRow['tags']) if variantRow['tags'] is not None else [],
```
simba-git, line 124: Why are we storing this as JSON? Where is it JSON?

client/src/featureform/sqlite_metadata.py
```diff
@@ -267,23 +267,65 @@ def createTables(self):
         VALUES ('provider', new.name, '', new.description, new.status);
         END;''')
 
+        # Tags and Properties tables
+        self.__conn.execute('''CREATE TABLE IF NOT EXISTS tags(
+          name text NOT NULL,
+          variant text,
+          type text,
+          tag_list text,
+          UNIQUE(name,variant));''')
```
simba-git, line 276: This should also include resource types? Imagine we have a source and a feature with the same name/variant

client/src/featureform/sqlite_metadata.py
```diff
@@ -267,23 +267,65 @@ def createTables(self):
...
+
+        self.__conn.execute('''CREATE TABLE IF NOT EXISTS properties(
+          name text NOT NULL,
+          variant text,
+          type text,
+          property_list text,
+          UNIQUE(name,variant));''')
```
simba-git, line 283: Same as above

simba-git (approved):

client/src/featureform/dashboard_metadata.py
epps: I figured it would simpler to store JSON rather that handling multiple rows (e.g. a record per tag). But honestly, it was an arbitrary decision.

client/src/featureform/sqlite_metadata.py
epps: Good catch! I missed this. I'll update the unique constraint.

PR #688, opened 2023-03-16
Tags & Properties: Hosted Implementation in Client

epps:
# Description

This PR adds the hosted implementation for tags and properties in the client, as well as a bug fix for avoiding rescheduling jobs in the coordinator.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -1476,7 +1476,8 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	if err := resourceNamedSafely(id); err != nil {
 		return nil, err
 	}
-	if existing, err := serv.lookup.Lookup(id); err == nil {
+	existing, _ := serv.lookup.Lookup(id)
+	if existing != nil {
 		if err := existing.Update(serv.lookup, res); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


metadata/metadata.go
```diff
@@ -1476,7 +1476,8 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	if err := resourceNamedSafely(id); err != nil {
 		return nil, err
 	}
-	if existing, err := serv.lookup.Lookup(id); err == nil {
+	existing, _ := serv.lookup.Lookup(id)
```
sdreyer, line 1479: Why cancel the error out?

epps: I waffled on that. I thought that the previous error was due to an existing resource, so now the presence/absence of the resource determines behavior, but errors could occur for other reasons, too. I'll add that back. Thanks for catching this.

sdreyer (approved):

metadata/metadata.go
```diff
@@ -1476,7 +1476,11 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	if err := resourceNamedSafely(id); err != nil {
 		return nil, err
 	}
-	if existing, err := serv.lookup.Lookup(id); err == nil {
+	existing, err := serv.lookup.Lookup(id)
+	if _, ok := err.(*ResourceNotFound); err != nil && !ok {
```
sdreyer, line 1480: lets change `ok` to `isReasourceError` or something similar. Have to stare at it for a bit rn to figure out the logic with the `ok`

PR #699, opened 2023-03-24
Localmode Caching

aolfat:
# Description

Added local mode caching

- On serve, all features, labels, transformations, and training sets are cached as .pkl files (we can do parquet too). If the cache file exists the fetch is done from that.
- When requesting a resource, if the source file for that resource has changed, the cache is deleted.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/local_cache.py
+++ b/client/src/featureform/local_cache.py
@@ -72,40 +72,55 @@ class LocalCache:
         Caches the result of a training set to a local file. Difference between this one and the one above
         is how this needs to fetch all the source files for the training set.
         """
+        resource_type = "training_set"
+
         file_path = self._cache_file_path(
-            "training_set", training_set_name, training_set_variant
+            resource_type, training_set_name, training_set_variant
         )
+
+        # check db for source files
+        source_files_from_db = self.db.get_source_files_for_resource(
+            resource_type, training_set_name, training_set_variant
+        )
+
+        # Only check to invalidate the cache if we have source files in the db
+        if source_files_from_db:
+            self._invalidate_cache_if_source_files_changed(
+                source_files_from_db, file_path
+            )
 
         source_files = set()
-
-        ts_variant = self.db.get_training_set_variant(
-            training_set_name, training_set_variant
-        )
-        label_variant = self.db.get_label_variant(
-            ts_variant["label_name"], ts_variant["label_variant"]
-        )
-        source_files.update(
-            self.get_source_files_for_source(
-                label_variant["source_name"], label_variant["source_variant"]
+        if source_files_from_db:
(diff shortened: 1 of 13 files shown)


client/src/featureform/local_cache.py
```diff
@@ -0,0 +1,169 @@
...
+from featureform import SQLiteMetadata  # fix to do client.source.import
+from featureform.resources import SourceType  # fix to do client.source.import
+from pandas.core.generic import NDFrame
+from typeguard import typechecked
+
+
+class LocalCache:
+    def __init__(self, db: SQLiteMetadata):
+        self.db = db
+        self.cache_dir = os.environ.get("FEATUREFORM_CACHE_DIR", ".featureform/cache")
```
aolfat, line 17: this should possibly be elsewhere. We need an env var for .featureform directory so that this is a relative path

sdreyer: Can you make the featureform directory an env var then set the default to ~/.featureform and set this inside of that?

client/tests/test_files/input_files/transactions.csv
sdreyer, line 1: ?

client/tests/test_localmode_caching.py
```diff
@@ -0,0 +1,175 @@
...
+    resource_client = ff.ResourceClient(local=True)
+    resource_client.apply()
+
+    serving_client = ff.ServingClient(local=True)
+    serving_client.training_set("fraud_training", "quickstart")
+
+    return serving_client
+
+
+# @pytest.fixture(autouse=True)
```
sdreyer, line 77: delete if unused

client/src/featureform/local_cache.py
```diff
@@ -0,0 +1,208 @@
...
+                    ["updated_at"],
+                    resource_type,
+                    resource_name,
+                    resource_variant,
+                    source_file,
+                    str(os.path.getmtime(source_file)),
+                )
+            return df
+
+    @lru_cache(maxsize=128)
```
aolfat, line 160: we actually probably don't need this anymore because we're storing this to db

client/src/featureform/local_utils.py
```diff
@@ -0,0 +1,147 @@
+import re
+
+import pandas as pd
```
aolfat, line 3: These functions were bloating the LocalClient file so I moved them out in a utils file. Downside is we have to switch off files but upside is a leaner file with the core functionality in the file (and class)

aolfat: This is a refactor!

sdreyer (approved):

sdreyer (approved):

PR #701, opened 2023-03-25
Class API for Entities, Features and Labels

epps:
# Description

This PR introduces a new API for defining entities, features and labels.

### Select type(s) of change

- [x] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/cli.py
+++ b/client/src/featureform/cli.py
@@ -231,7 +231,11 @@ def read_url(url):
 
 def exec_file(file, name):
     code = compile(file.read(), name, "exec")
-    exec(code)
+    # Create a new global namespace for each file to ensure that
+    # global variables, such as `ff`, are not undefined in the
+    # context of class attribute assignments (e.g. `label = ff.Label()`)
+    file_globals = {}
+    exec(code, file_globals)
 
 
 if __name__ == '__main__':
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -625,8 +625,9 @@ class LocalSource:
         return fn
 
     def __getitem__(self, columns: List[str]):
-        if len(columns) < 2:
-            raise Exception("Missing entity and source columns")
+        col_len = len(columns)
+        if col_len < 2:
+            raise Exception(f"Expected 2 columns, but found {col_len}. Missing entity and/or source columns")
         return (self.registrar, self.name_variant(), columns)
 
     def name_variant(self):
@@ -713,8 +714,9 @@ class SubscriptableTransformation:
         pass
 
     def __getitem__(self, columns):
-        if len(columns) < 2:
-            raise Exception("Missing entity and source columns")
+        col_len = len(columns)
+        if col_len < 2:
+            raise Exception(f"Expected 2 columns, but found {col_len}. Missing entity and/or source columns")
         return (self.registrar, self.name_variant(), columns)
 
     def __call__(self, *args, **kwds):
@@ -901,6 +903,12 @@ class DFTransformationDecorator:
 
 
 class ColumnSourceRegistrar(SourceRegistrar):
+
+    def __getitem__(self, columns: List[str]):
+        col_len = len(columns)
+        if col_len < 2:
+            raise Exception(f"Expected 2 columns, but found {col_len}. Missing entity and/or source columns")
+        return (self.registrar(), self.id(), columns)
 
     def register_resources(
             self,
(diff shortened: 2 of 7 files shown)


client/src/featureform/cli.py
```diff
@@ -231,7 +231,11 @@ def read_url(url):
 
 def exec_file(file, name):
     code = compile(file.read(), name, "exec")
-    exec(code)
+    # Create a new global namespace for each file to ensure that
+    # global variables, such as `ff`, are not undefined in the
+    # context of class attribute assignments (e.g. `label = ff.Label()`)
+    file_globals = {}
```
simba-git, line 237: lol, I'm sorry you had to debug this, but great work

client/src/featureform/register.py
```diff
@@ -802,9 +849,19 @@ def __call__(self, fn):
             if self.name is nv[0] and self.variant is nv[1]:
                 raise ValueError(f"Transformation cannot be input for itself: {self.name} {self.variant}")
         self.query = dill.dumps(fn.__code__)
-        fn.register_resources = self.register_resources
-        fn.name_variant = self.name_variant
-        return fn
+        # Given the function type does not implement __getitem__ we need to wrap it
```
simba-git, line 852: Noticed that you copy and pasted this, would it make sense to make it its own function or put the logic in the constructor?

client/src/featureform/register.py
```diff
@@ -3947,6 +4010,194 @@ def search(self, raw_query, local=False):
...
+            entity_column=self.entity_column,
+            owner=self.owner,
+            inference_store=self.inference_store,
+            features=features,
+            labels=labels,
+            timestamp_column=self.timestamp_column,
+            schedule=self.schedule,
+        )
+    
+    def get_features_and_labels(self) -> Tuple[List[ColumnMapping], List[ColumnMapping]]:
```
simba-git, line 4069: would it make sense to separate these into two methods? If not would it make sense to drop the get_ in the name. Super minor nitpick feel free to ignore

client/src/featureform/register.py
```diff
@@ -3947,6 +4010,194 @@ def search(self, raw_query, local=False):
...
+            if resource.variant != variant_key:
+                raise ValueError(
+                    f"Variant name {variant_key} does not match resource variant name {resource.variant}"
+                )
+
+    def register(self):
+        for resource in self.resources.values():
+            resource.register()
+
+# Class API
```
simba-git, line 4108: Would cut this comment

simba-git (dismissed):
Looks great, nice work

Left some minor nitpicks that you can totally ignore as well

aolfat:
I downloaded black formatting for my stuff: https://black.readthedocs.io/en/stable/. I've been using it and trying to get everyone else aboard.

client/src/featureform/register.py
```diff
@@ -3947,6 +4010,194 @@ def search(self, raw_query, local=False):
             return search(processed_query, self._host)
 
 
+class ColumnResource:
```
aolfat, line 4013: You're effectively creating an abstract class here right? You can force the inability to instantiate with this -> https://www.geeksforgeeks.org/abstract-classes-in-python/

client/src/featureform/register.py
```diff
@@ -3947,6 +4010,194 @@ def search(self, raw_query, local=False):
...
+    """
+    Base class for all column resources. This class is not meant to be instantiated directly.
```
aolfat, line 4015: > This class is not meant to be instantiated directly.

How about we make this an abstract class? https://www.geeksforgeeks.org/abstract-classes-in-python/

client/src/featureform/register.py
epps: Definitely cleaner to wrap this up in the wrapper class. Thanks for the suggestion!

client/src/featureform/register.py
epps: I took a look at this as I was unfamiliar with the module, but I'm not sure it fits the use case here. I actually don't want to reimplement the methods in the derived classes, so having a functional base class is what I want, no?

client/tests/test_class_api.py
```diff
@@ -0,0 +1,222 @@
...
+        custom_marks
+    )
+    label_entity_column = "CustomerID" if is_local else "customerid"
+
+    with pytest.raises(Exception, match="Expected 2 columns"):
+        source[[label_entity_column]]
+
+
+@pytest.fixture(autouse=True)
+def before_and_after_each(setup_teardown):
```
aolfat, line 119: this has been copied pasta'd in a lot of places. I'm not sure what the correct solution is but ideally we don't' have to.

sdreyer (approved):

PR #702, opened 2023-03-28
Adding the client changes for OnDemand Features

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/resources.py
+++ b/client/src/featureform/resources.py
@@ -60,7 +60,6 @@ class SourceType(Enum):
     PRIMARY_SOURCE = "PRIMARY"
     DF_TRANSFORMATION = "DF"
     SQL_TRANSFORMATION = "SQL"
-    ONDEMAND_FEATURE = "ONDEMAND_FEATURE"
 
 
 @typechecked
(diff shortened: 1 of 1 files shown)


client/src/featureform/register.py
```diff
@@ -2479,6 +2479,45 @@ def df_transformation(self,
         self.__resources.append(decorator)
         return decorator
 
+    def ondemand_feature(self,
+                          tags: List[str],
+                          properties: dict,
```
epps, line 2484: Do we want to make these default arguments so users don't have to specify tags and properties?

client/src/featureform/register.py
```diff
@@ -2479,6 +2479,45 @@ def df_transformation(self,
...
+                          variant: str = "default",
+                          name: str = "",
+                          owner: Union[str, UserRegistrar] = "",
+                          description: str = "",
+                          ):
+        """On Demand Feature decorator.
+
```
epps, line 2491: Do we want to add an example section here?

```
    **Examples**
    ```python
    @ondemand_feature
    def avg_user_transactions():
        pass
    ```
```

client/src/featureform/register.py
ahmadnazeri: addressed this in local mode PR

PR #703, opened 2023-03-28
adding support for hosted mode

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -129,7 +129,7 @@ class HostedClientImpl:
     def training_set(self, name, variation, include_label_timestamp, model: Union[str, Model] = None):
         return Dataset(self._stub).from_stub(name, variation, model)
 
-    def features(self, features, entities, model: Union[str, Model] = None):
+    def features(self, features, entities, model: Union[str, Model] = None, params: list = []):
         req = serving_pb2.FeatureServeRequest()
         for name, value in entities.items():
             entity_proto = req.entities.add()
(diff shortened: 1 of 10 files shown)


client/src/featureform/serving.py
```diff
@@ -129,7 +129,7 @@ def _create_channel(self, host, insecure, cert_path):
     def training_set(self, name, variation, include_label_timestamp, model: Union[str, Model] = None):
         return Dataset(self._stub).from_stub(name, variation, model)
 
-    def features(self, features, entities, model: Union[str, Model] = None):
+    def features(self, features, entities, model: Union[str, Model] = None, params: list = []):
```
sdreyer, line 132: think @aolfat had a post about using empty lists as defaults and how you shouldnt do it

client/src/featureform/serving.py
```diff
@@ -142,7 +142,20 @@ def features(self, features, entities, model: Union[str, Model] = None):
         if model is not None:
             req.model.name = model if isinstance(model, str) else model.name
         resp = self._stub.FeatureServe(req)
-        return [parse_proto_value(val) for val in resp.values]
+
+        feature_values = []
+        for val in resp.values:
+            parsed_value = parse_proto_value(val)
+            
+            is_ondemand_feature = type(parsed_value) == bytes
```
sdreyer, line 150: Is there a different way to check the type? We may want to return bytes as a value in the future

client/src/featureform/serving.py
aolfat: https://featureform-workspace.slack.com/archives/C022U1JHDB3/p1679592789364619

sdreyer: Also how does this handle multiple on-demand features in one call? are params a list of tuples?

ahmadnazeri: params will be used by the user. It can essentially be a list of whatever they want.

client/src/featureform/serving.py
ahmadnazeri: the alternative will be for each val, we will have to pull the metadata for that feature in that position and see the category if is on demand_feature. 

```
for i, val in enumerate(resp.values):
      feature = features[i]
      category = self._stub.PullFeatureMetadata(feature) # need to figure out right method for this
      
      if category == "ONDEMAND_FEATURE":
```

another approach is we modify FeatureServe to return values and categories.

epps (approved):

client/src/featureform/serving.py
sdreyer: Isnt there a way to check what the oneof type is?

PR #730, opened 2023-04-11
Uploading PySpark Script to correct path

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/spark.go
```diff
@@ -629,17 +632,17 @@ type EMRExecutor struct {
...
-	_, err = scriptFile.Read(buff)
-	if err != nil {
-		return err
+	e.logger.Info("Uploading PySpark script to filestore")
+	sparkLocalScriptPath := config.GetSparkLocalScriptPath()
+	sparkRemoteScriptPath := config.GetSparkRemoteScriptPath()
+	sparkScriptPathWithPrefix := store.PathWithPrefix(sparkRemoteScriptPath, false)
+
+	err := readAndUploadFile(sparkLocalScriptPath, sparkScriptPathWithPrefix, store)
+	scriptExists, _ := store.Exists(sparkScriptPathWithPrefix)
```
simba-git, line 641: Are we dropping an error here that we should check?

.github/workflows/testing.yml
```diff
@@ -19,6 +19,8 @@ env:
   ETCD_PORT: 2379
   REDSHIFT_PORT: 5439
   REDSHIFT_DATABASE: dev
+  SPARK_LOCAL_SCRIPT_PATH: scripts/spark/offline_store_spark_runner.py
```
sdreyer, line 22: Are these used for integration testing?

client/src/featureform/resources.py
```diff
@@ -216,6 +218,7 @@ def serialize(self) -> bytes:
             "Credentials": self.credentials.config(),
```
sdreyer, line 218: Can we put this dict in a function so its consistent?

dashboard/src/components/resource-list/ResourceListView.js
```diff
@@ -467,7 +467,7 @@ const NoDataMessage = ({ type }) => {
   const classes = useStyles();
 
   function redirect() {
-    window.location.href = "https://docs.featureform.com/quickstart";
+    window.location.href = "https://docs.featureform.com/getting-started/overview";
```
sdreyer, line 470: In the future, should just put this is its own pr

provider/filestore.go
```diff
@@ -114,19 +114,21 @@ func (store AzureFileStore) AsAzureStore() *AzureFileStore {
 }
 
 func (store AzureFileStore) PathWithPrefix(path string, remote bool) string {
+	pathContainsAzureBlobPrefix := strings.HasPrefix(path, azureBlobPrefix)
```
sdreyer, line 117: Can you add comments in these functions for what the expected path should look like and/or add tests that enforce it

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -80,7 +80,7 @@ def execute_sql_query(job_type, output_uri, sql_query, spark_configs, source_lis
 
         dt = datetime.now()
         safe_datetime = dt.strftime("%Y-%m-%d-%H-%M-%S-%f")
-        output_uri_with_timestamp = f'{output_uri}{safe_datetime}/'
+        output_uri_with_timestamp = f"{output_uri}{safe_datetime}" if output_uri[-1] == "/" else f"{output_uri}/{safe_datetime}"
```
sdreyer, line 83: Whats this do?

.github/workflows/testing.yml
ahmadnazeri: yes!

client/src/featureform/resources.py
ahmadnazeri: the other configs have serialize method the same way. we can call `self.config()` here but we will have to change all of the other ones too to stay consistent.

dashboard/src/components/resource-list/ResourceListView.js
ahmadnazeri: I can remove the change and open a new PR with just this change.

provider/scripts/spark/offline_store_spark_runner.py
ahmadnazeri: basically, if the output location doesn't end with `/`, it adds so date time will be separate directory.

dashboard/src/components/resource-list/ResourceListView.js
simba-git: Don't bother

provider/filestore_test.go
```diff
@@ -0,0 +1,342 @@
...
+				Remote:       true,
+				ExpectedPath: fmt.Sprintf("s3://%s/featureform-unit-test/path_to_file/file.csv", s3BucketName),
+			},
+			"path_without_slash_remote": TestCase{
+				Path:         "path_to_file/file.csv",
+				Remote:       true,
+				ExpectedPath: fmt.Sprintf("s3://%s/featureform-unit-test/path_to_file/file.csv", s3BucketName),
+			},
+		},
+		// "hdfs": map[string]TestCase{
```
aolfat, line 105: we keeping this?

provider/filestore_test.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get local filestore: %v", err)
+	}
+
+	s3FileStore, err := getS3FileStore()
+	if err != nil {
+		return nil, fmt.Errorf("could not get s3 filestore: %v", err)
+	}
+
+	// hdfsFileStore, err := getHDFSFileStore()
```
aolfat, line 204: and this?

provider/filestore_test.go
simba-git, line 105: ?

provider/k8s.go
```diff
@@ -396,7 +396,7 @@ type genericFileStore struct {
 func (store genericFileStore) PathWithPrefix(path string, remote bool) string {
 	// What does this mean? Change this as check for local file
 	if len(store.path) > 4 && store.path[0:4] == "file" {
-		return fmt.Sprintf("%s%s", store.path[len("file:///"):], path)
+		return fmt.Sprintf("%s%s", store.path[len("file:///"):], strings.TrimPrefix(path, "/"))
```
simba-git, line 399: I've now seen this line of code like 5x, should this be done differently?

provider/scripts/spark/offline_store_spark_runner.py
simba-git: Make this two different lines:
one line should add a / at the end, call it output_uri_with_trailing_slash
Then just add the safe_datetime to that

provider/spark.go
```diff
@@ -136,17 +136,21 @@ func (s3 SparkS3FileStore) Type() string {
 }
 
 func (s3 SparkS3FileStore) PathWithPrefix(path string, remote bool) string {
-	noS3Prefix := !strings.HasPrefix(path, "s3a://")
+	pathContainsS3Prefix := strings.HasPrefix(path, s3aPrefix)
```
simba-git, line 139: I think we need to make a struct called Path that handles all this stuff across all the different files. Thoughts @sdreyer ? We can do it later but this just seems like a mess, I can see why path keeps being an issue.

provider/scripts/spark/offline_store_spark_runner.py
aolfat: could rewrite it as this too:
```suggestion
        output_uri_with_timestamp = f"{output_uri.rstrip('/')}/{safe_datetime}""
```

provider/k8s.go
simba-git: I think we really need a Path struct that manages all this stuff.

simba-git (dismissed):

client/src/featureform/resources.py
sdreyer: What i mean is that why doesnt serialized just call `config=self.config()` instead of rewriting it again right below

provider/filestore.go
sdreyer: Is this referenced somewhere?

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -118,7 +118,7 @@ def execute_df_job(output_uri, code, store_type, spark_configs, credentials, sou
 
         dt = datetime.now()
         safe_datetime = dt.strftime("%Y-%m-%d-%H-%M-%S-%f")
-        output_uri_with_timestamp = f"{output_uri}{safe_datetime}" if output_uri[-1] == "/" else f"{output_uri}/{safe_datetime}"
+        output_uri_with_timestamp = f"{output_uri.rstrip('/')}/{safe_datetime}"
```
sdreyer, line 121: Can we add a comment here? Unclear what why this exists by looking at it

provider/spark.go
sdreyer: agreed

provider/filestore.go
ahmadnazeri: I added tests for `PathWithPrefix` and the `pathContainsAzureBlobPrefix` is referenced in the conditional statement below.

sdreyer (approved):

PR #734, opened 2023-04-12
Python 3.11.2 Support

epps:
# Description

This PR introduces minor changes required to support the latest version of Python (i.e. 3.11.2).

### Select type(s) of change

- [ ] Documentation update
- [x] Language Support

# Checklist:

### DIFF
--- a/client/setup.cfg
+++ b/client/setup.cfg
@@ -26,7 +26,7 @@ install_requires =
     protobuf>=3.20.0
     typeguard<3.0.0
     grpcio>=1.47.0
-    numpy==1.21.6
+    numpy>=1.21.6
     pandas==1.3.5
     pandasql>=0.7.3
     typing_extensions>=4.1.1
@@ -34,7 +34,7 @@ install_requires =
     flask==2.2.1
     Flask-Cors==3.0.10
     validators>=0.20.0
-    dill==0.3.5.1
+    dill>=0.3.6
     pandasql==0.7.3
     sqlalchemy<2.0.0
     requests
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -736,15 +736,16 @@ class SubscriptableTransformation:
         self.fn = fn
         self.registrar = registrar
         self.provider = provider
-        # Binds the register_resources and name_variant methods to the subscriptable_fn
-        # using the descriptor protocol. This allows us to call the methods on the
-        # subscriptable_fn instance. **NOTE** the MethodType could also be used to bind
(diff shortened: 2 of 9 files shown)


client/setup.cfg
```diff
@@ -26,15 +26,15 @@ install_requires =
     protobuf>=3.20.0
     typeguard<3.0.0
     grpcio>=1.47.0
-    numpy==1.21.6
+    numpy
```
simba-git, line 29: Any argument to set this to a >=

client/setup.cfg
```diff
@@ -26,15 +26,15 @@ install_requires =
...
     pandas==1.3.5
     pandasql>=0.7.3
     typing_extensions>=4.1.1
     dataclasses==0.6
     flask==2.2.1
     Flask-Cors==3.0.10
     validators>=0.20.0
-    dill==0.3.5.1
+    dill
```
simba-git, line 37: Same

client/src/featureform/register.py
```diff
@@ -736,15 +736,8 @@ def __init__(self, fn, registrar, provider, decorator_register_resources_method,
...
-        # Binds the register_resources and name_variant methods to the subscriptable_fn
-        # using the descriptor protocol. This allows us to call the methods on the
-        # subscriptable_fn instance. **NOTE** the MethodType could also be used to bind
-        # the methods to the subscriptable_fn instance; however, the descriptor protocol
-        # produces the same result while also being compatible with `classmethod`
-        # and `staticmethod`.
-        self.register_resources = decorator_register_resources_method.__get__(self)
-        self.name_variant = decorator_name_variant_method.__get__(self)
-        pass
+        self.register_resources = decorator_register_resources_method
```
simba-git, line 739: Ironically this is originally how I thought we should do this, and the whole __get__ thing confused the hell out of me. I just assume you can use fns and methods however you want in python.

simba-git (approved):

client/src/featureform/register.py
epps: I probably overcomplicated this based on a misguided bias from JS. If this were JS, I would never expect the value of `self` to be correct unless I bound it. When I searched for how to do this in Python, I happened upon the approach, which was touted as better than just a "raw" assignment or using `types.MethodType`. Lesson learned.

client/setup.cfg
epps: I'm not sure I'm following your comment here. Are you saying you'd prefer some `>=` specification?

simba-git (approved):

aolfat (approved):

PR #738, opened 2023-04-13
Bugfix: Consistent global state (disallow resource client to register)

aolfat:
# Description

Resource Client had it's own local state separate from the global state. We're no longer providing the ability for the resource client to register resources.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -2988,7 +2988,7 @@ class ResourceClient:
             resources = resource_state.sorted_list()
             display_statuses(self._stub, resources)
 
-        self.clear_state()
+        clear_state()
 
 
     def get_user(self, name, local=False):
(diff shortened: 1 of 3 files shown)


sdreyer (dismissed):

client/src/featureform/register.py
```diff
@@ -2921,7 +2921,7 @@ def register_model(self, name: str, tags: List[str] = [], properties: dict = {})
         return model
 
 
-class ResourceClient(Registrar):
+class ResourceClient:
```
ahmadnazeri, line 2924: how would these changes affect the docs? would it be a breaking change?

aolfat: good question. I made sure to check. The fact that you can use the Resource Client to register shit is pretty much a hidden feature/easter egg. Removing it is fine

PR #740, opened 2023-04-14
adding support for spark sql transformation with 'default' variant

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/tests/test_spark_provider.py
+++ b/client/tests/test_spark_provider.py
@@ -87,6 +87,32 @@ def test_sql_transformation(name, variant, sql, spark_provider):
 
 
 @pytest.mark.parametrize(
+    "sql",
+    [
+        ("SELECT * FROM {{test_name.test_variant}}"),
+    ]
+)
+def test_sql_transformation_without_variant(sql, spark_provider):
+    def transformation():
+        """doc string"""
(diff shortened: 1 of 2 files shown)


client/tests/test_spark_provider.py
```diff
@@ -86,6 +86,32 @@ def transformation():
...
+    def transformation():
+        """doc string"""
+        return sql
+
+    decorator = spark_provider.sql_transformation()
+    decorator(transformation)
+
+    assert decorator.to_source() == Source(
+        name=transformation.__name__,
+        variant="default",
```
aolfat, line 105: we should also be able to not pass in a variant right?

aolfat (approved):

client/tests/test_spark_provider.py
epps: This does already default to `"default"`, so we _could_ leave it out.

epps (approved):

client/tests/test_spark_provider.py
ahmadnazeri: It does, just wanted to be explicit because it is a test case

PR #758, opened 2023-04-20
Adding Spark error messages for Databricks provider

ahmadnazeri:
# Description
Making sure that Databricks returns the error of each run whenever it fails. 

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/spark.go
```diff
@@ -478,12 +479,47 @@ func (db *DatabricksExecutor) RunSparkJob(args []string, store SparkFileStore) e
...
+	runID := runs[0].RunId
+	request := jobs.GetRunRequest{
+		RunId: runID,
+	}
+
+	// in order to get the status of the run output, we need to
+	// use API v2.0 instead of v2.1. The following code leverages
+	// version 2.0 of the API to get the run output.
+	var runOutput jobs.RunOutput
+	path := "/api/2.0/jobs/runs/get-output"
```
epps, line 514: Couple of questions here:
1. Why is this? Or I should ask, will the "why" eventually change, and if so, will we have to update `path`?
2. Do we want to move the value of `path` into configuration with a constant default? Is this overkill?

epps (approved):
I left a couple of questions that you may choose to ignore, but otherwise, this looks good to me.

provider/spark.go
```diff
@@ -351,9 +354,10 @@ const (
 )
 
 type DatabricksExecutor struct {
-	client  *databricks.WorkspaceClient
-	cluster string
-	config  pc.DatabricksConfig
+	client    *databricks.WorkspaceClient
+	cluster   string
+	config    pc.DatabricksConfig
+	runClient *dbClient.DatabricksClient
```
aolfat, line 360: what is the diff between client & runclient?

aolfat: A few q's
- Is the run clients primary purpose to fetch logs from dbrix?
- Is there a case where the regular client can still connect but the runClient can fail? My concern is, if the answer to above is yes (runClient purely for log fetching), if we run on dbricks and there are no failures but there is a problem with connecting to the run client our entire dbricks executor fails.

provider/spark.go
```diff
@@ -478,12 +479,47 @@ func (db *DatabricksExecutor) RunSparkJob(args []string, store SparkFileStore) e
...
 }
 
+func (db *DatabricksExecutor) getErrorMessage(jobId int64) (string, error) {
+	ctx := context.Background()
+
+	runRequest := jobs.ListRunsRequest{
+		JobId: jobId,
+	}
+
+	runs, err := db.client.Jobs.ListRunsAll(ctx, runRequest)
```
aolfat, line 500: is there a way to only fetch with runrequest?

provider/spark.go
```diff
@@ -430,29 +434,26 @@ func NewDatabricksExecutor(databricksConfig pc.DatabricksConfig) (SparkExecutor,
 			Username: databricksConfig.Username,
 			Password: databricksConfig.Password,
 		}))
+
+	runClient, err := dbClient.New(&dbConfig.Config{
```
aolfat, line 438: should we only create this when we fetch errors? From my naive understanding I imagine the flow should be like this.

- We run on dbrix
- Something fails (we return the error message that we get from fmt.error)
- We attempt to get the more verbose message from dbrix but if it fails its fine and we can still default to the other error?

aolfat (approved):

PR #760, opened 2023-04-20
Source Dataframes in the Client (Local Mode)

epps:
# Description

This PR introduces the ability to fetch sources (either files or transformation) to the client for experimentation purposes. Currently, only local mode is supported; however, the hosted implementation is the next priority.

### Select type(s) of change

- [x] Documentation update

# Checklist:


client/src/featureform/register.py
```diff
@@ -3252,7 +3254,13 @@ def __init__(
             insecure (bool): True if connecting to an insecure Featureform endpoint. False if using a self-signed or public TLS certificate
             cert_path (str): The path to a public certificate if using a self-signed certificate.
         """
-        super().__init__()
```
simba-git, line 3255: Why remove the super __init__?

client/src/featureform/serving.py
```diff
@@ -67,6 +69,13 @@ class ServingClient:
     """
 
     def __init__(self, host=None, local=False, insecure=False, cert_path=None):
```
simba-git, line 71: probably should add the super init here

ahmadnazeri (dismissed):

client/src/featureform/register.py
epps: Pretty sure this was leftover from when we separated `ResourceClient` from `Registrar`. As it stands now, there's no purpose in calling `super().__init__()`, though, if I'm not mistaken, it's benign to leave it. I can return it if you're concerned about it.

client/src/featureform/serving.py
epps: Maybe I'm missing something important, but if `ServingClient` doesn't inherit from another class, what purpose would calling `super` have here?

client/src/featureform/client.py
```diff
@@ -0,0 +1,75 @@
...
+            dry_run=dry_run,
+        )
+        # Given both ResourceClient and ServingClient are instantiated together, if dry_run is True, then
+        # the ServingClient cannot be instantiated due to a conflict the local and host arguments.
+        if not dry_run:
+            ServingClient.__init__(
+                self, host=host, local=local, insecure=insecure, cert_path=cert_path
+            )
+
+    def compute_df(
```
simba-git, line 47: I know this is a change up about what I said before but I think we name this datafarme( instead of compute_df, the logic being:
to get a training set you'd call: client.training_set, it follows that for a dataframe you'd call: client.dataframe(

client/src/featureform/client.py
```diff
@@ -0,0 +1,75 @@
...
+from .register import (
+    ResourceClient,
+    SourceRegistrar,
+    LocalSource,
+    SubscriptableTransformation,
+)
+from .serving import ServingClient
+
+
+class Client(ResourceClient, ServingClient):
```
simba-git, line 11: This is a hack which makes sense for now but when we deprecate we should fit all the code into here

client/src/featureform/client.py
```diff
@@ -0,0 +1,75 @@
...
+    redis = client.get_provider("redis-quickstart")
+
+    # Example 2: Compute a dataframe from a registered source
+    transactions_df = client.compute_df("transactions", "quickstart")
+    """
+
+    def __init__(
+        self, host=None, local=False, insecure=False, cert_path=None, dry_run=False
+    ):
+        ResourceClient.__init__(
```
simba-git, line 32: this should be:
super(ResourceClient, self).__init__

client/src/featureform/client.py
```diff
@@ -0,0 +1,75 @@
...
+            host=host,
+            local=local,
+            insecure=insecure,
+            cert_path=cert_path,
+            dry_run=dry_run,
+        )
+        # Given both ResourceClient and ServingClient are instantiated together, if dry_run is True, then
+        # the ServingClient cannot be instantiated due to a conflict the local and host arguments.
+        if not dry_run:
+            ServingClient.__init__(
```
simba-git, line 43: super(ServingClient, self). __init

Can you explain why we can't do this if its a dry run? I don't understand the comment. We should also call super __init__ whenever possible

client/src/featureform/register.py
simba-git: I think you're right we can leave it out

client/src/featureform/serving.py
simba-git: I think you're fine to leave it out, mixed answers from my reading

client/src/featureform/client.py
simba-git: The reason being:
If you were to cast this as a ServingClient and use it, you'd end up with some weird behavior if dry run is set

client/src/featureform/client.py
epps: @simba-git, I'd originally written it this way; however, I ran into errors relating to differing params (e.g. `dry_run`) and wasn't able to access `impl` on instances of `Client`. After reading [this SO answer](https://stackoverflow.com/questions/26927571/multiple-inheritance-in-python3-with-different-signatures) it seemed to me that I was reaching for the wrong tool for this job. Let me know what you think.

client/src/featureform/client.py
epps: Currently, the logic in `ServingClient`'s constructor will throw an exception is `local` is `False` and `host` is `None`, which is true in the case when a user calls `> featureform apply definitions.py --dry-run`. Given the presence of `dry_run` indicates that nothing can be done with `ServingClient`, I followed the pattern I saw in `ResouceClient` and put an early return here to avoid the unindented exception.

sdreyer: Dry run was a kind of hacky thing I added to test things out a while ago. I think we should have a larger conversation outside of this task about what dry run should actually do and how we want to support it

client/src/featureform/serving.py
```diff
@@ -630,6 +653,9 @@ def _register_model(
         self.db.insert("models", name, type)
         self.db.insert(look_up_table, name, association_name, association_variant)
 
+    def get_source_as_df(self, name, variant):
+        return self.get_input_df(name, variant)
```
sdreyer, line 657: Curious if this works with the caching @aolfat added

client/src/featureform/serving.py
```diff
@@ -180,6 +190,12 @@ def features(
 
         return feature_values
 
+    def get_source_as_df(self, name, variant):
```
sdreyer, line 193: was this the agreed upon naming?

sdreyer (approved):
lgtm, just some questions

client/src/featureform/serving.py
epps: No, and now that I'm thinking about it, I should have added an underscore prefix to the method name.

My thought here was to define a method that was shared across both implementations that would be wrapped by the "public" `dataframe` method we're defining on `Client`. Eventually, when we refactor, the implementation will be added to `Client` and we can work with a standard API for both hosted and local mode.

client/src/featureform/serving.py
epps: It does because I'm calling into the same local method that's used when serving. I've tested this manually and checked `.featureform/cache`.

ahmadnazeri (dismissed):

client/docs/client.md
```diff
@@ -0,0 +1 @@
+:::src.featureform.client
```
aolfat, line 1: what is this?

Makefile
```diff
@@ -234,6 +234,7 @@ pytest:
 	pytest -m 'local' client/tests/test_ondemand_features.py
 	pytest client/tests/test_localmode_caching.py
 	pytest -m 'local' client/tests/test_resource_registration.py
+	pytest -m 'local' client/tests/test_source_dataframe.py
```
aolfat, line 237: for my ocd, can you move client/tests/test_local under (or above) all the pytest -m?

client/src/featureform/client.py
aolfat: Yeahhh I think since this is only temporary it's fine. The more idiomatic way is to use super but it's chaotic just like the SO answer suggested

client/docs/client.md
epps: I believe this is a routing mechanism for [`mkdocs`](https://www.mkdocs.org/), though, I've never actually looked into this. This file was autogenerated by `mkdocs`.

client/src/featureform/enums.py
```diff
@@ -52,9 +54,52 @@ class SourceType(Enum):
...
     S3 = "s3://"
     S3A = "s3a://"
+
+
+class FileFormat(Enum):
+    CSV = "csv"
+    PARQUET = "parquet"
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, str):
```
aolfat, line 77: why do we need this equals? Can't we just leverage the enum equals?

client/src/featureform/serving.py
aolfat: I thought the functionality would be on the source directly? `average_user_transaction.df()`

client/src/featureform/enums.py
epps: I wasn't aware the `Enum` derived classes have an `equals` method. I got hung up on `'csv'  == FileFormat.CSV.value` and when I looked for a way to avoid indexing into the enum, all of the SO articles described adding dunder methods for comparison operations. Seemed cleaner, but I'm happy to revert it.

aolfat: ohhhh I see what you're doing. okk i like thi

aolfat: FileFormat(str, Enum)

client/src/featureform/enums.py
```diff
@@ -52,9 +54,52 @@ class SourceType(Enum):
     DF_TRANSFORMATION = "DF"
     SQL_TRANSFORMATION = "SQL"
 
+    def __eq__(self, other):
```
aolfat, line 57: same here. I think you can just do ColumnTypes(str, Enum) do avoid this equals

client/src/featureform/enums.py
simba-git: I wouldn't inherit from str, seems like a super hack (always prefer composition over inheritence), and never seen it in python code before, I think this is fine as-is

client/src/featureform/serving.py
simba-git: No it's via the Client now

simba-git: I think method name is ok

client/src/featureform/serving.py
simba-git: nice!

simba-git (dismissed):

client/src/featureform/enums.py
aolfat: it's actually really common to do this. I saw it a ton at doordash. It's actually so common that python released StrEnum in 3.11

aolfat (approved):

aolfat (approved):

PR #761, opened 2023-04-21
Adding Spark error messages for EMR and Databricks

ahmadnazeri:
# Description
Previously, we could not see the Spark error messages whenever a job failed on EMR or Databricks. With this PR, we will be able to pull the error message from EMR and Databricks and return it to the user. 

### Select type(s) of change

- [ ] Documentation update

# Checklist:


helpers/decompress/decompress_message.go
```diff
@@ -0,0 +1,29 @@
+package decompress
+
+import (
+	"bytes"
+	"compress/gzip"
+	"fmt"
+)
+
+func GZipMessageToString(message []byte) (string, error) {
```
simba-git, line 9: Could just rename to ParseGZip or even jsut GZip since the package name is decompress so this becomes: decompress.GZip(...

helpers/decompress/decompress_message.go
```diff
@@ -0,0 +1,29 @@
...
+	buffer := bytes.NewBuffer(message)
+
+	gr, err := gzip.NewReader(buffer)
+	if err != nil {
+		return "", fmt.Errorf("could not create a new gzip Reader: %v", err)
+	}
+
+	var decompressed bytes.Buffer
+	if _, err := decompressed.ReadFrom(gr); err != nil {
+		return "", fmt.Errorf("could not read from gzip Reader: %v", err)
```
simba-git, line 22: if this fails, gr remains open

helpers/decompress/decompress_message.go
```diff
@@ -0,0 +1,29 @@
...
+func GZipMessageToString(message []byte) (string, error) {
+	// this function takes a gzip compressed byte array and
+	// decompresses it into string
+
+	buffer := bytes.NewBuffer(message)
+
+	gr, err := gzip.NewReader(buffer)
+	if err != nil {
+		return "", fmt.Errorf("could not create a new gzip Reader: %v", err)
+	}
```
simba-git, line 18: could throw defer gr.Close() after here

provider/spark.go
```diff
@@ -478,12 +481,50 @@ func (db *DatabricksExecutor) RunSparkJob(args []string, store SparkFileStore) e
...
+		return nil, fmt.Errorf("could not get run id: %v", err)
+	}
+
+	runID := runs[0].RunId
+	request := jobs.GetRunRequest{
+		RunId: runID,
+	}
+
+	// in order to get the status of the run output, we need to
+	// use API v2.0 instead of v2.1. The following code leverages
```
simba-git, line 516: Does this get removed in 2.1? Why?

provider/spark.go
```diff
@@ -868,12 +928,79 @@ func (e *EMRExecutor) RunSparkJob(args []string, store SparkFileStore) error {
...
+		if stepResults.Step.Status.FailureDetails.LogFile != nil {
+			logFile := *stepResults.Step.Status.FailureDetails.LogFile
+
+			bucket, path := getBucketAndPathFromFilePath(logFile)
+			outputFilePath := fmt.Sprintf("%s/stdout.gz", path)
+
+			// wait until log file exists
+			for {
+				fileExists, err := e.logFileStore.Exists(outputFilePath)
+				if err != nil {
```
simba-git, line 975: we're now like 5 layers deep, break this up into a lot of different smaller fns

simba-git, line 975: we're now like 5 layers deep, break this up into a lot of different smaller fns

provider/spark.go
```diff
@@ -478,12 +481,50 @@ func (db *DatabricksExecutor) RunSparkJob(args []string, store SparkFileStore) e
...
+	runRequest := jobs.ListRunsRequest{
+		JobId: jobId,
+	}
+
+	runs, err := db.client.Jobs.ListRunsAll(ctx, runRequest)
+	if err != nil {
+		return nil, fmt.Errorf("could not get run id: %v", err)
+	}
+
+	runID := runs[0].RunId
```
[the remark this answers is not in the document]
ahmadnazeri: yes, because we create a new job for each execution and then run that job. Thus, there will be only one run for that particular jobId.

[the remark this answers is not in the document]
sdreyer: is there every a change it returns an empty list due to an error or something?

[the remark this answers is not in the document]
ahmadnazeri: I don't see a chance for it. But if we want, I can check if length of runs is not zero?

provider/filepath.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+type Filepath interface {
+	Bucket() string
+	Prefix() string
+	Path() string
+	FullPathWithBucket() string
```
sdreyer, line 17: Whats the difference between this and Path?

provider/filepath.go
```diff
@@ -0,0 +1,99 @@
...
+	FullPathWithoutBucket() string
+	ParseFullPath(path string) error
+}
+
+func NewFilepath(storeType string, bucket string, prefix string, path string) (Filepath, error) {
+	switch storeType {
+	case "s3":
```
sdreyer, line 24: Should be an enum

provider/spark.go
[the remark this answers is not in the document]
sdreyer: yeah, better safe than crashing

provider/spark.go
```diff
@@ -868,12 +928,93 @@ func (e *EMRExecutor) RunSparkJob(args []string, store SparkFileStore) error {
...
+		// check if there are any errors
+		if stepResults.Step.Status.FailureDetails.Message != nil {
+			// get the error message
+			errorMsg = *stepResults.Step.Status.FailureDetails.Message
+		}
+		if errorMsg != "" {
+			return errorMsg, nil
+		}
+
+		if stepResults.Step.Status.FailureDetails.LogFile != nil {
```
sdreyer, line 966: Feel like this whole thing could be a function

provider/filepath.go
ahmadnazeri: This includes `s3://bucket/prefix/path` where path is `path`

provider/filepath.go
ahmadnazeri: should we make Filestore type as enum as well?

provider/filepath.go
```diff
@@ -0,0 +1,99 @@
...
+				prefix: prefix,
+				path:   path,
+			},
+		}, nil
+	default:
+		return nil, fmt.Errorf("unknown store type '%s'", storeType)
+	}
+}
+
+type genericFilepath struct {
```
simba-git, line 37: Make sure this handles trailing / vs no trailing / well since I know we've had a handful of bugs come from stuff like that

simba-git: Could also rename this to filePath. Also the idea of a bucket isn't necessarily generic for all file paths right?

provider/spark.go
simba-git: Got it, can you add this context and link the issue in the comment here

provider/spark.go
```diff
@@ -868,12 +928,79 @@ func (e *EMRExecutor) RunSparkJob(args []string, store SparkFileStore) error {
...
+				fileExists, err := e.logFileStore.Exists(outputFilePath)
+				if err != nil {
+					return "", fmt.Errorf("could not determine if file '%s' exists: %v", outputFilePath, err)
+				}
+
+				if fileExists {
+					break
+				}
+
+				time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
ahmadnazeri: I was mistaken. It can take a few minutes to make the log available. 
```
Because log files are uploaded to Amazon S3 every 5 minutes, it can take a few minutes for the log file uploads to complete after the step completes.
```
from [AWS Documentation on Logs](https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-manage-view-web-log-files.html#emr-manage-view-web-log-files-debug)

provider/filepath.go
```diff
@@ -0,0 +1,101 @@
...
+				path:   strings.TrimPrefix(path, "/"),
+			},
+		}, nil
+	default:
+		return nil, fmt.Errorf("unknown store type '%s'", storeType)
+	}
+}
+
+type filePath struct {
+	bucket string
```
ahmadnazeri, line 40: @simba-git, `bucket` is used in the https://pkg.go.dev/gocloud.dev/blob package that we use internal to interact with blob stores. Thus, I wanted to use the same term. But if you think another term is better suited then I am more than happy to change it.

helpers/compression/compression_test.go
```diff
@@ -0,0 +1,68 @@
+//go:build compress
+// +build compress
```
aolfat, line 2: what is this?

ahmadnazeri: this can be used for `--tags` when running tests to run just these tests.

helpers/compression/compression.go
```diff
@@ -0,0 +1,31 @@
...
+
+package compression
+
+import (
+	"bytes"
+	"compress/gzip"
+	"fmt"
+)
+
+func GunZip(message []byte) (string, error) {
```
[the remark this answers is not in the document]
ahmadnazeri: So apparently, the cli is gunzip and Simba thought it was best to use the same name.

provider/filepath.go
aolfat: This seems like it's a filestorepath vs filepath right?

aolfat: my thinking is, if someone looks at this for the first time the question they're most likely going to ask is what's the diff b/w this and go's filepath?

ahmadnazeri: That's fair; so we do have different types like `S3Filepath` and will add additional paths as needed. But I can definitely modify the interface to FilestorePath if we want.

Issue #762, opened 2023-04-21
[Bug]: Search Results do not navigate to resource. 


RedLeader16:
### Expected Behavior

When the user uses search, the result list should navigate to a resource when clicked. 

### Actual Behavior

Clicking a search result does nothing. 

ex. 

https://user-images.githubusercontent.com/34227093/233747449-a7a6c6c3-cd46-427e-9817-d9890307ad59.mov

### Steps To Reproduce

1. Type "user" into the search box. 
2. Click on a result. 
3. Kaboom. Nothing.

### What mode are you running Featureform in?

Local

### What version of Python are you running?

3.7

### Featureform Python Package Version

0.7.2

PR #765, opened 2023-04-24
Fix Search Results navigation for #762

RedLeader16:
# Description
This MR fixes the bug where navigation results do not route the user to the resource. 
It also addresses a bug in the breadcrumbs where the query params display.

**Working Search**

https://user-images.githubusercontent.com/34227093/234102309-17c5a4eb-8f4b-491c-b954-e7d43999a2d0.mov

**BreadCrumbs fix:**
before:
beforeFix

after:
afterFix

https://github.com/featureform/featureform/issues/762

Unit Tests: 
image

### Select type(s) of change

- [ ] Documentation update

# Checklist:


dashboard/src/components/searchresults/SearchResultsView.js
```diff
@@ -112,42 +113,39 @@ const SearchResultsList = ({ type, contents, setVariant }) => {
...
+  "LABEL_VARIANT": "Label",
+  "TRAINING_SET": "TrainingSet",
+  "TRAINING_SET_VARIANT": "TrainingSet",
+  "SOURCE": "Source",
+  "SOURCE_VARIANT": "Source",
+  "PROVIDER": "Provider",
+  "ENTITY": "Entity",
+  "MODEL": "Model",
+  "USER": "User",
 }
```
RedLeader16, line 128: @sdreyer for the purposes of search, do variant mappings point back to the original type?

simba-git: Not sure I understand the question but you also have to think about the situation where I'm searching for a specific variant. we also don't want to clog up a search result with all of the variants if I'm searching for a feature.

Ex.
If I search for "abc"
And I have a feature variant with name F1 and Variants [abc, def, ghi], then the abc variant should show up in the results.

simba-git (approved):
Great work

dashboard/src/components/searchresults/SearchResultsView.js
RedLeader16: Gotcha, that makes sense to me. After the user clicks a variant search result, the existing code has a "hasVariant" flag that handles the routing. All clear!

PR #770, opened 2023-04-26
Source Dataframes in Client (Hosted)

epps:
# Description

This PR adds the ability to fetch registered sources as dataframes in hosted mode.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/api/main.go
+++ b/api/main.go
@@ -680,6 +680,9 @@ func (serv *OnlineServer) TrainingData(req *srv.TrainingDataRequest, stream srv.
 
 func (serv *OnlineServer) SourceData(req *srv.SourceDataRequest, stream srv.Feature_SourceDataServer) error {
 	serv.Logger.Infow("Serving Source Data", "id", req.Id.String())
+	if req.Limit == 0 {
+		return fmt.Errorf("limit must be greater -1 to denote no limit or a positive integer")
+	}
 	client, err := serv.client.SourceData(context.Background(), req)
 	if err != nil {
 		return fmt.Errorf("could not serve source data: %w", err)
@@ -699,7 +702,7 @@ func (serv *OnlineServer) SourceData(req *srv.SourceDataRequest, stream srv.Feat
 	}
 }
 
-func (serv *OnlineServer) SourceColumns(ctx context.Context, req *srv.SourceDataRequest) (*srv.SourceDataColumns, error) {
+func (serv *OnlineServer) SourceColumns(ctx context.Context, req *srv.SourceColumnRequest) (*srv.SourceDataColumns, error) {
 	serv.Logger.Infow("Serving Source Columns", "id", req.Id.String())
 	return serv.client.SourceColumns(ctx, req)
 }
--- a/api/main.go
+++ b/api/main.go
@@ -678,6 +678,32 @@ func (serv *OnlineServer) TrainingData(req *srv.TrainingDataRequest, stream srv.
 	}
 }
 
+func (serv *OnlineServer) SourceData(req *srv.SourceDataRequest, stream srv.Feature_SourceDataServer) error {
+	serv.Logger.Infow("Serving Source Data", "id", req.Id.String())
+	client, err := serv.client.SourceData(context.Background(), req)
+	if err != nil {
+		return fmt.Errorf("could not serve source data: %w", err)
+	}
+	for {
+		row, err := client.Recv()
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return fmt.Errorf("receive error: %w", err)
+		}
+		if err := stream.Send(row); err != nil {
+			serv.Logger.Errorf("failed to write to source data stream: %w", err)
+			return fmt.Errorf("source send row: %w", err)
+		}
+	}
+}
+
+func (serv *OnlineServer) SourceColumns(ctx context.Context, req *srv.SourceDataRequest) (*srv.SourceDataColumns, error) {
+	serv.Logger.Infow("Serving Source Columns", "id", req.Id.String())
+	return serv.client.SourceColumns(ctx, req)
+}
+
 func (serv *ApiServer) Serve() error {
 	if serv.grpcServer != nil {
 		return fmt.Errorf("server already running")
--- a/client/src/featureform/client.py
+++ b/client/src/featureform/client.py
@@ -7,6 +7,8 @@ from .register import (
 )
 from .serving import ServingClient
 
+NO_RECORD_LIMIT = -1
+
 
 class Client(ResourceClient, ServingClient):
     """
@@ -48,6 +50,7 @@ class Client(ResourceClient, ServingClient):
         self,
         source: Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str],
         variant="default",
+        limit=NO_RECORD_LIMIT,
     ):
         """
         Compute a dataframe from a registered source or transformation
@@ -72,4 +75,4 @@ class Client(ResourceClient, ServingClient):
             raise ValueError(
                 f"source must be of type SourceRegistrar, LocalSource, SubscriptableTransformation or str, not {type(source)}"
             )
-        return self.impl.get_source_as_df(name, variant)
+        return self.impl._get_source_as_df(name, variant, limit)
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -37,6 +37,8 @@ from .tls import insecure_channel, secure_channel
 from .version import check_up_to_date
 from .enums import FileFormat
 
+NO_RECORD_LIMIT = -1
+
 
 def check_feature_type(features):
     checked_features = []
@@ -136,42 +138,6 @@ class ServingClient:
         features = check_feature_type(features)
         return self.impl.features(features, entities, model)
 
-    def source_columns(self, name, variant="default"):
-        """Returns the columns for the specified source.
-
-        **Examples**:
-        ``` py
-            client = ff.ServingClient(local=True)
-            columns = client.source_columns("transactions")
-            # columns = ["user", "amount", "timestamp"]
-        ```
-        Args:
-            name (str): Name of source to be retrieved
-            variant (str): Variant of source to be retrieved
-
-        Returns:
-            columns (list[str]): A list of column names
-        """
-        return self.impl.source_columns(name, variant)
-
-    def source_data(self, name, variant="default"):
-        """Returns the data for the specified source.
-
-        **Examples**:
-        ``` py
-            client = ff.ServingClient(local=True)
-            data = client.source_data("transactions")
-            # data = [["C1410926", 100, 1620000000], ["C1410926", 200, 1620000001]]
-        ```
-        Args:
-            name (str): Name of source to be retrieved
-            variant (str): Variant of source to be retrieved
-
-        Returns:
-            data (list[list]): A list of rows of data
-        """
-        return self.impl.source_data(name, variant)
-
 
 class HostedClientImpl:
     def __init__(self, host=None, insecure=False, cert_path=None):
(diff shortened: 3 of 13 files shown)


serving/serialize.go
```diff
@@ -80,6 +99,29 @@ func (row *row) AddFeature(feature interface{}) error {
...
+		}
+	}
+	return r, nil
+}
+
+func (row *sourceRow) Serialized() *pb.SourceDataRow {
+	return row.serialized
+}
+
+func (r *sourceRow) AddRow(row interface{}) error {
```
simba-git, line 116: Should be called AddValue? Or something similar

api/main.go
```diff
@@ -678,6 +678,35 @@ func (serv *OnlineServer) TrainingData(req *srv.TrainingDataRequest, stream srv.
 	}
 }
 
+func (serv *OnlineServer) SourceData(req *srv.SourceDataRequest, stream srv.Feature_SourceDataServer) error {
+	serv.Logger.Infow("Serving Source Data", "id", req.Id.String())
+	if req.Limit == 0 {
+		return fmt.Errorf("limit must be greater -1 to denote no limit or a positive integer")
```
sdreyer, line 684: confusing error

client/src/featureform/serving.py
```diff
@@ -37,6 +37,8 @@
 from .version import check_up_to_date
 from .enums import FileFormat
 
+NO_RECORD_LIMIT = -1
```
sdreyer, line 40: We have this twice, should be in the enum or a consts file

client/src/featureform/client.py
```diff
@@ -48,6 +50,7 @@ def dataframe(
         self,
         source: Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str],
         variant="default",
+        limit=NO_RECORD_LIMIT,
```
aolfat, line 53: can you add the limit to the arg docs.

client/src/featureform/serving.py
```diff
@@ -653,8 +670,16 @@ def _register_model(
         self.db.insert("models", name, type)
         self.db.insert(look_up_table, name, association_name, association_variant)
 
-    def get_source_as_df(self, name, variant):
-        return self.get_input_df(name, variant)
+    def _get_source_as_df(self, name, variant, limit):
+        if limit == 0:
+            raise ValueError(
+                "Limit must be greater -1 to denote no limit or a positive integer"
```
sdreyer, line 676: Confusing message

api/main.go
```diff
@@ -678,6 +678,35 @@ func (serv *OnlineServer) TrainingData(req *srv.TrainingDataRequest, stream srv.
 	}
 }
 
+func (serv *OnlineServer) SourceData(req *srv.SourceDataRequest, stream srv.Feature_SourceDataServer) error {
+	serv.Logger.Infow("Serving Source Data", "id", req.Id.String())
+	if req.Limit == 0 {
```
ahmadnazeri, line 683: should we move or add this check to the client?

client/src/featureform/client.py
```diff
@@ -7,6 +7,8 @@
 )
 from .serving import ServingClient
 
+NO_RECORD_LIMIT = -1
```
ahmadnazeri, line 10: do we want to have a limit? I don't want someone to request a df for 100m file.

client/tests/test_source_dataframe.py
```diff
@@ -28,9 +34,14 @@ def test_dataframe_for_name_variant_args(
     transformation = arrange_transformation(provider, is_local)
 
     client = ff.Client(local=is_local, insecure=is_insecure)
-    client.apply(asynchronous=True)
+    # If we're running in a hosted context, `apply` needs to be synchronous
```
sdreyer, line 37: Do we have a check or warning for this when running manually?

api/main.go
ahmadnazeri: nvm

provider/filepath.go
```diff
@@ -36,6 +37,27 @@ func NewFilepath(storeType pc.FileStoreType, bucket string, prefix string, path
...
+		return &AzureFilepath{}, nil
+	case GCS:
+		return &GCSFilepath{}, nil
+	case Memory:
+		return nil, fmt.Errorf("currently unsupported file store type '%s'", storeType)
+	case FileSystem:
+		return nil, fmt.Errorf("currently unsupported file store type '%s'", storeType)
+	case pc.DB:
+		return nil, fmt.Errorf("currently unsupported file store type '%s'", storeType)
+	case HDFS:
```
sdreyer, line 54: Any reason for not adding this?

provider/bigquery.go
```diff
@@ -97,10 +97,14 @@ func (pt *bqPrimaryTable) GetName() string {
 
 func (pt *bqPrimaryTable) IterateSegment(n int64) (GenericTableIterator, error) {
 	tableName := pt.query.getTableName(pt.name)
-	query := fmt.Sprintf("SELECT * FROM `%s` LIMIT %d", tableName, n)
+	var query string
+	if n == -1 {
+		query = fmt.Sprintf("SELECT * FROM `%s` LIMIT %d", tableName, n)
```
sdreyer, line 102: backwards, this will limit -1 rows

provider/filepath_test.go
```diff
@@ -182,9 +182,16 @@ func TestParseFullPath(t *testing.T) {
 			ExpectedError:   nil,
 		},
 		"AzureBlobStoreFilePath": {
-			StoreType:       Azure,
-			FullPath:        "abfss://container@account.dfs.core.windows.net/elasticmapreduce/path/to/file",
-			ExpectedPath:    nil,
+			StoreType: Azure,
```
ahmadnazeri, line 185: can we add the tests for `GSFilePath` as well?

provider/k8s.go
```diff
@@ -659,6 +661,149 @@ func (store genericFileStore) NumRows(key string) (int64, error) {
...
+		if float, err := strconv.ParseFloat(value, 64); err == nil {
+			records[i] = float
+			continue
+		}
+		records[i] = value
+	}
+	return records
+}
+
+func newCSVIterator(b []byte, limit int64) (GenericTableIterator, error) {
```
ahmadnazeri, line 723: would these be used outside of the package?

provider/k8s.go
```diff
@@ -659,6 +661,149 @@ func (store genericFileStore) NumRows(key string) (int64, error) {
...
+}
+
+func (c *csvIterator) Close() error {
+	return nil
+}
+
+func (c *csvIterator) ParseRow(row []string) GenericRecord {
+	records := make(GenericRecord, len(row))
+	for i, value := range row {
+		if integer, err := strconv.Atoi(value); err == nil {
```
sdreyer, line 710: Does this only parse ints a floats?

serving/serving.go
```diff
@@ -101,6 +129,46 @@ func (serv *FeatureServer) getTrainingSetIterator(name, variant string) (provide
...
 }
 
+func (serv *FeatureServer) getSourceDataIterator(name, variant string, limit int64) (provider.GenericTableIterator, error) {
+	ctx := context.TODO()
+	serv.Logger.Infow("Getting Source Variant Iterator", "name", name, "variant", variant)
+	sv, err := serv.Metadata.GetSourceVariant(ctx, metadata.NameVariant{name, variant})
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get source variant")
+	}
+	// TODO: change this to use another method given this won't work for non-SQL sources
```
ahmadnazeri, line 139: ?

proto/serving.proto
```diff
@@ -64,3 +66,25 @@ message Value {
         bytes on_demand_function = 8;
     }
 }
+
+message SourceID {
+  string name = 1;
+  string version = 2;
```
aolfat, line 72: variant? to keep things consistent?

provider/bigquery.go
```diff
@@ -97,10 +97,14 @@ func (pt *bqPrimaryTable) GetName() string {
 
 func (pt *bqPrimaryTable) IterateSegment(n int64) (GenericTableIterator, error) {
 	tableName := pt.query.getTableName(pt.name)
-	query := fmt.Sprintf("SELECT * FROM `%s` LIMIT %d", tableName, n)
+	var query string
```
aolfat, line 100: maybe out of scope for this but is this something we want to possibly paginate on?

client/tests/test_source_dataframe.py
epps: What do you mean "running manually"? In any case, you need to apply in the context of the tests. And actually, now that I'm seeing this again, I don't even need to pass this arg because `asynchronous=True` is the default.

provider/filepath.go
epps: Purely a matter of time at this point.

provider/k8s.go
epps: Yes, I can extend this but I figured these basic types could cover the most ground.

proto/serving.proto
epps: `version` actually is the consistent naming convention for the protos. I figured it would be better to refactor them later together instead of introducing this inconsistency.

client/tests/test_source_dataframe.py
sdreyer: Just wondering if someone decides to us it and run it async=False, is there a check for it, or will it just error?

provider/k8s.go
sdreyer: Would recommend adding string at least since thats common

sdreyer: may need to change this format. seems like it with either always be cast as an int or error

provider/k8s.go
sdreyer: This should go in the filestore package whenever we actually get that out

client/tests/test_source_dataframe.py
epps: There's not check, so I believe the behavior is it'll error

provider/k8s.go
epps: Sorry, maybe this wasn't clear: if it can't part an int or a float, it'll keep the value as a string.

sdreyer: Ah i see, i was reading it backwards

sdreyer (dismissed):

client/src/featureform/serving.py
```diff
@@ -653,8 +669,14 @@ def _register_model(
         self.db.insert("models", name, type)
         self.db.insert(look_up_table, name, association_name, association_variant)
 
-    def get_source_as_df(self, name, variant):
-        return self.get_input_df(name, variant)
+    def _get_source_as_df(self, name, variant, limit):
+        if limit == 0:
+            raise ValueError("limit must be greater than 0")
+        df = self.get_input_df(name, variant)
```
aolfat, line 675: shouldn't we pass the limit over to the server?

epps: This is the local implementation. We do pass it for the hosted implementation. See line 199 above.

proto/serving.proto
```diff
@@ -11,6 +11,8 @@ package featureform.serving.proto;
 service Feature {
   rpc TrainingData(TrainingDataRequest) returns (stream TrainingDataRow) {}
   rpc FeatureServe(FeatureServeRequest) returns (FeatureRow) {}
+  rpc SourceData(SourceDataRequest) returns (stream SourceDataRow) {}
+  rpc SourceColumns(SourceColumnRequest) returns (SourceDataColumns) {}
```
aolfat, line 15: https://cloud.google.com/apis/design/naming_convention#method_names

It might be too late now but the convention is VerbNoun
rpc GetSourceColumns(GetSourceColumnRequest) returns ...

client/src/featureform/serving.py
aolfat: ohhhhh this is local client, got it

proto/serving.proto
epps: Yeah, these serving protos definitely don't follow the conventions, whereas `metadata.proto` does. We should fix this.

sdreyer (approved):

PR #793, opened 2023-05-08
Improve Error Message for Feature Missing `timestamp_column`

epps:
# Description

If a user provides a value for `timestamp_column` when registering a feature, but the underlying transformation does not contain a column for it, the resulting error is like `KeyError: "['Timestamp'] not in index"`. This PR introduces a more helpful error message for the user.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/serving.py
+++ b/client/src/featureform/serving.py
@@ -510,8 +510,9 @@ class LocalClientImpl:
             df.reset_index(inplace=True)
         if feature["source_timestamp"] != "" and feature["source_timestamp"] not in df:
             raise ValueError(
-                f"Provided timestamp column {feature['source_timestamp']} not found in feature {feature['name']}; "
-                f"either remove 'timestamp_column' from the feature registration or include it in the transformation."
+                f"Provided timestamp column '{feature['source_timestamp']}' for feature "
+                f"'{feature['name']}:{feature['variant']}' not found in source '{feature['source_name']}:{feature['source_variant']}'; "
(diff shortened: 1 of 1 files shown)


client/src/featureform/serving.py
```diff
@@ -508,7 +508,12 @@ def feature_df_from_transformation(self, feature):
         if isinstance(df, pd.Series):
             df = df.to_frame()
             df.reset_index(inplace=True)
-        if feature["source_timestamp"] != "":
+        if feature["source_timestamp"] != "" and feature["source_timestamp"] not in df:
+            raise ValueError(
+                f"Provided timestamp column {feature['source_timestamp']} not found in feature {feature['name']}; "
```
ahmadnazeri, line 513: so we are creating the feature, the column would need to be in the source data.  
```python
f"Provided timestamp column '{feature['source_timestamp']}' for feature, '{feature['name']}:{feature['variant']}' not found in source, '{feature['source_name']}:{feature['source_variant']}';"
```

client/src/featureform/serving.py
```diff
@@ -508,7 +508,12 @@ def feature_df_from_transformation(self, feature):
...
+                f"either remove 'timestamp_column' from the feature registration or include it in the transformation."
```
ahmadnazeri, line 514: My assumption would be that a feature would most likely be build of a transformation. But we would have cases where it uses the primary source. Should we modify this statement to say include it in the source.

ahmadnazeri (approved):

PR #795, opened 2023-05-10
Bugfix: Pushing to online store broken

aolfat:
# Description

Remove the code that catches an error that we use to determine if the online store needs to be updated.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/runner/materialize.go
+++ b/runner/materialize.go
@@ -108,11 +108,7 @@ func (m MaterializeRunner) Run() (types.CompletionWatcher, error) {
 	}
 	m.Logger.Infow("Creating Table", "name", m.ID.Name, "variant", m.ID.Variant)
 	_, err = m.Online.CreateTable(m.ID.Name, m.ID.Variant, m.VType)
-	if err != nil {
-		return nil, fmt.Errorf("could not create table: %w", err)
-	}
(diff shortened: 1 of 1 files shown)


runner/materialize.go
```diff
@@ -108,11 +108,7 @@ func (m MaterializeRunner) Run() (types.CompletionWatcher, error) {
 	}
 	m.Logger.Infow("Creating Table", "name", m.ID.Name, "variant", m.ID.Variant)
 	_, err = m.Online.CreateTable(m.ID.Name, m.ID.Variant, m.VType)
-	if err != nil {
-		return nil, fmt.Errorf("could not create table: %w", err)
-	}
 	_, exists := err.(*provider.TableAlreadyExists)
-
 	if err != nil && !exists {
```
ahmadnazeri, line 112: should it fail if table already exists?

ahmadnazeri (approved):
LGTM, just one question

runner/materialize.go
aolfat: good question. This PR is actually a revert. This change that I'm changing was due to what you thought. The original code written by sam catches the exists error in order to update it

Issue #812, opened 2023-05-17
Display the FF version in the dashboard home page


RedLeader16:
Currently the cluster does not easily expose the FF version for other resources in the cluster. Configure the cluster so that it can easily share the app version where it is needed.

Once done include it the dashboard application (near the lower corners, verify with design first).

PR #817, opened 2023-05-17
Add Featureform version to the CLI and Dashhoard

RedLeader16:
# Description

This PR adds functionality to display the client and cluster version in the `cli`, and the cluster version in the `dashboard` application. It also includes unit tests for the backend and frontend. 

The `--local` flag is required when running in local mode.

## Type of change
`CLI`
Client Package Version and Cluster Version:
Screenshot 2023-05-26 at 2 48 27 PM

`local mode and hosted mode`
version and version local 2

`endpoint`
The metadata-dashboard `/data/version` endpoint now returns a `versionMap`. Currently the map returns a single `version` property, and is expected to be expanded in the future. (The difference in values is just the configuration differences between local and k8s)
Screenshot 2023-05-22 at 5 24 13 PM

image

`docker`
Running in docker mode reads from the `FEATUREFORM_VERSION` env. variable. 
dash version docker
docker env variable

#812 

### Select type(s) of change

- [ ] Documentation update

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation


charts/featureform/charts/dashboard-metadata/values.yaml
```diff
@@ -15,6 +15,9 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+#current application version
+featureformVersion: "v0.8.1"
```
RedLeader16, line 19: thoughts on using the `values.yaml` file as the version prop home? pretty standard but we can store it elsewhere too.

RedLeader16:
Should we drop the "Version:" in the text "Version: v0.8.1"? I think the "v" indicates it enough.

metadata/dashboard/dashboard_metadata.go
```diff
@@ -1031,13 +1031,20 @@ func (m *MetadataServer) GetSearch(c *gin.Context) {
 	c.JSON(200, result)
 }
 
+func (m *MetadataServer) GetVersionMap(c *gin.Context) {
+	versionMap := map[string]string{}
+	versionMap["version"] = help.GetEnv("FEATUREFORM_VERSION", "Default")
```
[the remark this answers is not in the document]
RedLeader16: Let's go with blank.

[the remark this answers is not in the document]
RedLeader16: So something to the effect of, 

If a version is missing: We display nothing on the page. The versionMap returns `blank`. 
If there is a version: We display the version value only. ex. "v0.8.1". And not include the "Version:" prefix.

charts/featureform/charts/dashboard-metadata/values.yaml
RedLeader16: updating from our side chat: 

we'll use the `.Values.global.version | default .Chart.AppVersion` approach. that way we can just set the global as needed during upgrades etc.

metadata/dashboard/dashboard_metadata.go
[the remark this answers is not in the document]
RedLeader16: decided to keep the "Version:" prefix.

charts/featureform/charts/dashboard-metadata/values.yaml
RedLeader16: and done.

metadata/dashboard/dashboard_metadata.go
```diff
@@ -1031,13 +1031,20 @@ func (m *MetadataServer) GetSearch(c *gin.Context) {
 	c.JSON(200, result)
 }
 
+func (m *MetadataServer) GetVersionMap(c *gin.Context) {
+	versionMap := map[string]string{}
```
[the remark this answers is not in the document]
RedLeader16: done.

client/src/featureform/cli.py
```diff
@@ -148,6 +149,11 @@ def list(host, cert, insecure, local, resource_type):
 app.register_blueprint(dashboard_app)
 
 
+@cli.command()
+def version():
+    print("Client Version: {}".format(get_package_version()))
```
sdreyer, line 154: Would we be able to do a GET here and if FEATUREFORM_HOST env var is set then get the cluster version?
Then have:
```
Client Version:
Cluster Version:
```
```
If FEATUREFORM_HOST != ""
try:
    version = Get <host>/data/version
except:
    version = "Could not fetch version"

```

RedLeader16: Yeah that should be possible, I'll give it a shot and report back.

RedLeader16: and done. I'll get with you first to double check some assumptions. 

Screenshot 2023-05-26 at 2 48 27 PM

RedLeader16: and done. I'll get with you first to double check some assumptions.
image

client/src/featureform/tls.py
```diff
@@ -15,3 +20,26 @@ def secure_channel(host, cert_path):
...
+def fetch_cluster_version(version_url=""):
+    requests.packages.urllib3.disable_warnings()
+    res = requests.get(url=version_url, verify=False)
+    response = json.loads(res.text)
+    return response["version"]
+
+
+def get_version_local():
+    local_port = os.getenv("LOCALMODE_DASHBOARD_PORT", 3000)
+    version_url = "localhost:{}/data/version".format(local_port)
```
ahmadnazeri, line 34: I would recommend using `f-strings`, just more readable. 
```
version_url = f"localhost:{local_port}/data/version"

return fetch_cluster_version(f"{insecure_protocol}{version_url}")
```

ahmadnazeri (approved):
lgtm; just a small recommendation of using f-strings on the python side

client/src/featureform/cli.py
```diff
@@ -148,6 +150,32 @@ def list(host, cert, insecure, local, resource_type):
...
+            cluster_version = "No host is set."
+    except:
+        cluster_version = "Cannot retrieve: Check your FEATUREFORM_HOST value. If using local mode, use the --local flag."
+
+    print(
+        """
+    Client Version: {}
+    Cluster Version: {}
+    """.format(
+            client_cersion, cluster_version
```
ahmadnazeri, line 174: is this suppose to be `client_version`?

client/src/featureform/cli.py
```diff
@@ -148,6 +150,32 @@ def list(host, cert, insecure, local, resource_type):
 app.register_blueprint(dashboard_app)
 
 
+@cli.command()
+@click.option("--local", is_flag=True, help="Required for local mode only")
+def version(local):
+    client_cersion = get_package_version()
```
sdreyer, line 156: typo

sdreyer (dismissed):

client/src/featureform/tls.py
RedLeader16: agreed. changing.

client/src/featureform/cli.py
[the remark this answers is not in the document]
RedLeader16: yup changed.

RedLeader16: done

client/src/featureform/dashboard_metadata.py
```diff
@@ -62,6 +62,11 @@ def deliver_static(asset):
     return dashboard_app.send_static_file("static/" + asset)
 
 
+@dashboard_app.route("/data/version")
+def version():
+    return {"version": get_package_version()}
```
ahmadnazeri, line 67: Is this the same as the client version? do we want to know the cluster version?

RedLeader16: IIRC we're using the python package version in local mode.

sdreyer (approved):

PR #829, opened 2023-05-25
Add support for Vector32 to Redis

epps:
# Description

This PR introduce the `Vector32` type such that users can write and read vectors to Redis as features.

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/redis.go
```diff
@@ -95,48 +115,138 @@ func (store *redisOnlineStore) DeleteTable(feature, variant string) error {
...
 	if err != nil {
-		return nil, fmt.Errorf("could not cast value: %s to %s: %w", val.String(), table.valueType, err)
+		return nil, fmt.Errorf("could not cast value: %v to %s: %w", resp, table.valueType, err)
 	}
 	return result, nil
 }
+
+// formatVector takes a map[string]interface{} and returns a string representation of the vector
+// using the VectorString32 method provided by the rueidis Redis package.
+func (table redisOnlineTable) formatVector(value map[string]interface{}) (string, error) {
```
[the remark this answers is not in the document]
epps: I was wondering this, too, especially because the test I wrote for the parquet iterator is nearly identical to the logic here. The issue was that I mistyped the unit test in `redis_test.go`, which led to a failing test even though the E2E flow worked as expected.

I can make this change and have it return `interface{}` with the underlying type being `[]float32`. Does this work for you?

provider/k8s.go
```diff
@@ -1442,31 +1447,92 @@ func (iter *FileStoreFeatureIterator) Next() bool {
...
 }
 
+// Attempts to parse timestamp in one of the following formats:
+// 1. "2006-01-02 15:04:05.000000 UTC"
+// 2. "2006-01-02 15:04:05.000000"
+// 3. "2006-01-02 15:04:05.000000 +0000 UTC"
+// If any one of the three formats is valid, returns the parsed timestamp, otherwise it
+// returns an error
+func (iter *FileStoreFeatureIterator) parseTimestamp(ts string) (time.Time, error) {
+	// Attempt to parse timestamp as UTC first
```
simba-git, line 1479: Could actually do this in a loop:

formats := []string{"", "", ""}
for format in formats {
  parsed, err := time.Parse(format, ts)
  if err == nil {
    return parsed, nil
  }
}
return time.Time{}, fmt.Errrorf(

provider/redis_test.go
```diff
@@ -9,24 +9,29 @@ package provider
...
-		&redis.Options{
-			Addr:     miniRedis.Addr(),
-			Password: "",
-			DB:       0,
+	redisClient, err := rueidis.NewClient(
+		rueidis.ClientOption{
+			InitAddress:  []string{miniRedis.Addr()},
+			Password:     "",
+			SelectDB:     0,
+			DisableCache: true,
```
epps, line 27: You should add a comment to explain this configuration in the context of the unit tests.

client/src/featureform/serving.py
```diff
@@ -180,12 +180,19 @@ def features(
         feature_values = []
         for val in resp.values:
             parsed_value = parse_proto_value(val)
+            value_type = type(parsed_value)
 
-            is_ondemand_feature = type(parsed_value) == bytes
-            if is_ondemand_feature:
+            # Ondemand features are returned as a byte array
```
simba-git, line 185: I actually preffered the code you deleted, but either is fine

simba-git (approved):

PR #837, opened 2023-05-30
[Task] Add debug env var

aolfat:
# Description

Add a debug env var that can be used for debug specific tasks?

Other options include adding an env var for [prod, sandbox, testing] but I don't think we're there yet

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/helpers/env.go
+++ b/helpers/env.go
@@ -50,3 +50,7 @@ func GetEnvBool(key string, fallback bool) bool {
 		return parsedValue, err
 	}).(bool)
 }
+
+func IsDebugEnv() bool {
+	return GetEnvBool("DEBUG", false)
+}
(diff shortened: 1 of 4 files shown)


helpers/env.go
```diff
@@ -50,3 +50,7 @@ func GetEnvBool(key string, fallback bool) bool {
 		return parsedValue, err
 	}).(bool)
 }
+
+func IsDebugEnv() bool {
```
aolfat, line 54: Other options include adding an env var for [prod, sandbox, testing] but I don't think we're there yet as a company

RedLeader16: good call out, and something to remember in the near future.

RedLeader16 (approved):

sdreyer (approved):

PR #842, opened 2023-06-01
RediSearch as Vector Database Provider

epps:
# Description

This PR introduces support for [RediSearch](https://redis.io/docs/stack/search/) as a vector database provider in Featureform. Users may now:
* persist embeddings in Redis provided they're using the RediSearch module
* perform KNN queries on embeddings

### Select type(s) of change

- [x] Documentation update

# Checklist:


coordinator/Dockerfile.old
```diff
@@ -66,16 +66,16 @@ RUN echo "PATH=${PATH}" > "${ENV}"
...
+# ARG TESTING
+# RUN if [ "$TESTING" = "True" ]; then \
+#     pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
+#   else \
+#     pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#     pyenv install 3.8.16 && pyenv global 3.8.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#     pyenv install 3.9.16 && pyenv global 3.9.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#     pyenv install 3.10.10 && pyenv global 3.10.10 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#     pyenv install 3.11.2 && pyenv global 3.11.2 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
+#   fi
```
epps, line 78: FYI, I'm aware this is commented out. I'm going to leave this as is for the time being to cut down on chart built times.

client/src/featureform/register.py
```diff
@@ -4511,7 +4515,9 @@ class prior to calling `register_column_resources` on the registrar.
     def __init__(
         self,
         transformation_args: tuple,
-        type: Union[ColumnTypes, str],
+        type: Union[ScalarType, str],
+        dims: int,
```
simba-git, line 4519: remove dims and is_embedding here, it shouldn't exist in the base class

client/src/featureform/register.py
```diff
@@ -4561,23 +4569,24 @@ def register(self):
         )
 
     def features_and_labels(self) -> Tuple[List[ColumnMapping], List[ColumnMapping]]:
```
simba-git, line 4571: wtf is this method used for lol

client/src/featureform/register.py
```diff
@@ -4561,23 +4569,24 @@ def register(self):
...
+            "column": self.source_column,
+            "type": self.type,
+            "description": self.description,
+            "tags": self.tags,
+            "properties": self.properties,
+        }
+
         if self.resource_type == "feature":
-            features = resources
+            resource["dims"] = self.dims
```
simba-git, line 4583: So the trick to get this dims / is_embedding out of this base class is to override features_and_labels in the EmbeddingColumnResourceClass.

so then you'd have something like:

def features_and_labels(self) ...
  features, labels = super().features_and_labels()
  features[0]["dims"] = self.dims
  features[0]["is_embedding"] = self.is_embedding
  return features, labels

client/src/featureform/serving.py
```diff
@@ -212,6 +219,13 @@ def _get_source_columns(self, name, variant):
         resp = self._stub.SourceColumns(req)
         return resp.columns
 
+    def _nearest(self, name, variant, vector, k):
```
simba-git, line 222: so this supports nearest by vector. Just out of curiosity, how hard would it be to do nearest by entity ID?

metadata/client.go
```diff
@@ -1349,6 +1351,30 @@ func (fn fetchPropertiesFn) Properties() Properties {
...
+
+type dimensionGetter interface {
+	GetDimension() uint32
+}
+
+type fetchDimensionFn struct {
+	getter dimensionGetter
+}
+
+func (fn fetchDimensionFn) Dimension() uint32 {
```
simba-git, line 1374: god I hate that this is a uint32, just cast it to an int please. I'm willing to just have this do weird shit if someone actually has a vector of a dimension that would require the 32nd bit

metadata/client.go
```diff
@@ -1487,6 +1517,14 @@ func (variant *FeatureVariant) IsOnDemand() bool {
 	}
 }
 
+func (variant *FeatureVariant) IsEmbedding() bool {
+	return variant.fetchIsEmbeddingFn.IsEmbedding()
+}
+
+func (variant *FeatureVariant) Dimension() uint32 {
```
simba-git, line 1524: please cast this to an int

provider/online.go
```diff
@@ -34,6 +33,19 @@ type OnlineStoreTable interface {
 	Get(entity string) (interface{}, error)
 }
 
+type VectorStore interface {
+	CreateIndex(feature, variant, entity string, vectorType VectorType) (VectorStoreTable, error)
+	GetIndex(feature, variant string) (string, error) // TODO: determine the type of index
```
simba-git, line 38: TODO?

simba-git: I was thinking about this some more and had another fun idea:

Let make all VectorStores OnlineStores,

So it becomes:
type VectorStore interface {
  OnlineStore
  GetIndex
  DeleteIndex
  }
  
  Note that Provider and Close should be removed since you'll get those via OnlineStore's interface

simba-git: Doing this won't change any other code, it's so that every VectorStore implements GetTable which makes life easier. We can talk through it if its still confusing

provider/online.go
```diff
@@ -34,6 +33,19 @@ type OnlineStoreTable interface {
...
+	DeleteIndex(feature, variant string) error
+	Close() error
+	Provider
+}
+
+type VectorStoreTable interface {
+	OnlineStoreTable
```
simba-git, line 45: This was genius, good thinking

provider/redis.go
```diff
@@ -34,58 +38,95 @@ func redisOnlineStoreFactory(serialized pc.SerializedConfig) (Provider, error) {
...
+}
+
 func (store *redisOnlineStore) Close() error {
-	// from the docs:
-	//
-	// "It is rare to Close a Client, as the Client is meant to be
-	// long-lived and shared between many goroutines.""
-	//
-	return store.client.Close()
+	store.client.Close()
```
simba-git, line 70: Does this .Close() not return anything?

provider/redis.go
```diff
@@ -34,58 +38,95 @@ func redisOnlineStoreFactory(serialized pc.SerializedConfig) (Provider, error) {
...
+	if _, isScalarType := ScalarTypes[ScalarType(vType)]; isScalarType {
+		valueType = ScalarType(vType)
+	} else {
+		valueType = &VectorType{}
+		err = json.Unmarshal([]byte(vType), valueType)
+		if err != nil {
+			return nil, err
+		}
+	}
+	table := &redisOnlineIndex{
```
simba-git, line 95: You only return reidsOnlineIndex if its specifically and embedding otherwise you should return a redisOnlineTable. The reason being that normal tables don't have the Nearest method. It should also use the other Get and Set in that situation.

simba-git: It is funny that this works anyway but its not backward compatible

provider/redis.go
```diff
@@ -34,58 +38,95 @@ func redisOnlineStoreFactory(serialized pc.SerializedConfig) (Provider, error) {
...
-	if err := store.client.HSet(context.TODO(), fmt.Sprintf("%s__tables", store.prefix), key.String(), string(valueType)).Err(); err != nil {
+	serialized, err := json.Marshal(ValueTypeJSONWrapper{valueType})
+	if err != nil {
 		return nil, err
 	}
+	cmd = store.client.B().
+		Hset().
+		Key(fmt.Sprintf("%s__tables", store.prefix)).
+		FieldValue().
+		FieldValue(key.String(), string(serialized)).
```
simba-git, line 125: Is this backwards compatible? What happens if something was in a cluster where it was stored the other way?

provider/redis.go
```diff
@@ -94,49 +135,211 @@ func (store *redisOnlineStore) DeleteTable(feature, variant string) error {
...
+		FtCreate().
+		Index(key.String()).
+		Schema().
+		FieldName(vectorField).
+		Vector("HNSW", int64(len(requiredParams)), requiredParams...).
+		Build()
+}
+
+func (store *redisOnlineStore) GetIndex(feature, variant string) (string, error) {
+	return "", nil
```
simba-git, line 166: ?

simba-git (commented):
Nice!

provider/online.go
epps: 🤣 This was _your_ idea, but I'll take whatever I can get.

provider/redis.go
epps: Unimplemented as of yet. I have question about if we want to use a separate naming convention for the index key and if we want a different struct to represent the index (e.g. `algo: "HNSW"` etc.)

client/src/featureform/register.py
epps: When I implemented the class API and was debugging some of the logic, I got annoyed with having the either/or logic of features/labels in my way, so I pushed it to another method. I obviously should name this something better than what I decided on at the time.

client/src/featureform/serving.py
epps: I don't want to promise anything given the time remaining before the release, but I honestly don't think it would be too hard to implement.

provider/redis.go
epps: With the new Redis client, the `Close` method returns nothing, oddly.

provider/redis.go
epps: So, if you look at how I'm handling the value type logic in `GetTable`, I believe it's backwards compatible. Let's say a user has a preexisting Redis instance with the value `int32`. I'll first try to determine if this is a "legacy" scalar value, and if it is, I'll use that. Otherwise, I'll try to unmarshal the string to `ValueType` and then determine the instance type of that. Full disclosure, this explanation isn't actually what you see above as I was pushing to test this last night and didn't actually implement completely. I'm add this now.

epps: This is now working and covered by unit tests.

provider/redis.go
```diff
@@ -34,58 +38,126 @@ func redisOnlineStoreFactory(serialized pc.SerializedConfig) (Provider, error) {
...
+			Given we're not making use of this feature (i.e. via the commands `DoCach` or
+			`DoMultiCache`), and given this is normally a feature that a user would have to
+			explicitly opt into for a connection via the Redis command `CLIENT TRACKING ON`,
+			we disable this feature by default to avoid any issues with Redis servers versions
+			under 6.0.
+
+			See details here:
+				* rueidis client-side caching: https://github.com/redis/rueidis#client-side-caching
+				* Redis CLIENT TRACKING command: https://redis.io/commands/client-tracking/
+		*/
```
epps, line 60: @simba-git and @sdreyer, let me know if this comment doesn't fully explain the decision here. Basically, the `rueidis` library makes a decision to opt users into a configuration that _may_ not be available for all users, and given we're not making use of this feature, I'm disabling it here to return our use of Redis back to its default.

sdreyer (approved):

PR #862, opened 2023-06-13
Task/abstract filestore provider

sdreyer:
# Description

Abstracts the filestore so that other providers can be added. This breaks on demand features, which will get fixed in a later PR

### Select type(s) of change

- [ ] Documentation update

# Checklist:


client/examples/local_serving.py
```diff
@@ -1,15 +1,15 @@
 import featureform as ff
 
-client = ff.ServingClient(local=True)
+client = ff.Client(local=True)
 
-training_set = client.training_set("fraud_training", "quickstart")
+training_set = client.training_set("fraud_training", "courageous_jones")
```
ahmadnazeri, line 5: the variant would change each time you run the `local_quickstart.py`? should we hardcode the variant?

client/src/featureform/providers/filestore.py
```diff
@@ -0,0 +1,139 @@
...
+from .online_store import OnlineStore, OnlineTable
+
+
+class LocalFileTable(OnlineTable):
+    def __init__(self, name, variant, root_path, stype):
+        self.name = name
+        self.variant = variant
+        self.root_path = root_path
+        self.dir = f"{root_path}/{name}/"
+        self.filepath = f"{root_path}/{name}/{variant}.csv"
```
ahmadnazeri, line 13: would these paths be an issue with windows?

ahmadnazeri (approved):

client/src/featureform/enums.py
```diff
@@ -17,6 +17,14 @@ class ScalarType(Enum):
     BOOL = "bool"
     DATETIME = "datetime"
 
+    @classmethod
+    def has_value(cls, value):
+        return value in cls._value2member_map_
```
aolfat, line 22:
```suggestion
          try:
            cls(value)
            return True
        except ValueError:
            return False
```

1. Doesn't depend on internal implementation of value2member map (that could change between python versions)
2. Conforms to the pythonic "easier to ask for forgiveness than permission" https://realpython.com/python-lbyl-vs-eafp/#the-easier-to-ask-forgiveness-than-permission-eafp-style

client/src/featureform/providers/online_store.py
```diff
@@ -0,0 +1,28 @@
...
+    def create_table(self, name, variant, entity_type):
+        raise NotImplementedError()
+
+    def get_table(self, name, variant):
+        raise NotImplementedError()
+
+    def table_exists(self, name, variant):
+        raise NotImplementedError()
+
+    def is_vector_store(self):
```
aolfat, line 11: why do we need this?

It's like having an implementation of List, then having an implementation of LinkedList then inherits from List, and then adding a function on the List interface that says `isLinkedList`.

client/src/featureform/providers/filestore.py
```diff
@@ -0,0 +1,139 @@
...
+
+        types_df = pd.DataFrame(
+            data={"name": [name], "variant": [variant], "type": [entity_type.value]}
+        ).astype({"name": "string", "variant": "string", "type": "string"})
+
+        types_df.to_csv(self.type_table, mode="a", header=header, index=False)
+
+    def get_table(self, name, variant):
+        if not os.path.exists(f"{self.path}/{name}/{variant}.csv"):
+            raise ValueError(f"Table {name} with variant {variant} does not exist")
```
aolfat, line 103: I don't think we need this since self.__get_type will do this for you

client/src/featureform/providers/filestore.py
```diff
@@ -0,0 +1,139 @@
...
+from ..register import ScalarType
+from .online_store import OnlineStore, OnlineTable
+
+
+class LocalFileTable(OnlineTable):
+    def __init__(self, name, variant, root_path, stype):
+        self.name = name
+        self.variant = variant
+        self.root_path = root_path
+        self.dir = f"{root_path}/{name}/"
```
aolfat, line 12: should we use pathlib so we don't have to deal with pathing stuff on our own? i.e. someone passes root path with a ton of backslashes or something?

client/src/featureform/providers/filestore.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        self.df.to_csv(self.filepath, index=False)
+
+    def __init_update_df(self):
+        df = pd.DataFrame(columns=["entity", "value"]).astype(
+            {"entity": "str", "value": self.stype}
+        )
+        return df
+
+    def get_path(self):
```
aolfat, line 37: nit: you can mark this as a @property

client/examples/local_serving.py
aolfat, line 5: i love courageous jones

client/src/featureform/providers/filestore.py
```diff
@@ -0,0 +1,139 @@
...
+
+        return ScalarType(matched_rows["type"].values[0])
+
+    def table_exists(self, name, variant):
+        return os.path.exists(f"{self.path}/{name}/{variant}.csv")
+
+    def is_vector_store(self):
+        return False
+
+    def __convert_type(self, stype: ScalarType):
```
aolfat, line 127: do we need this method?

client/src/featureform/providers/online_store.py
epps: We'll need some way of having the equivalent Golang type assertion. Would you suggest/prefer something like this?

```python
from abc import ABC

class OnlineStore(ABC):
    pass

class VectorStore(OnlineStore, ABC):
    pass

class Pinecone(VectorStore)
    pass

p = Pinecone()

if isinstance(store_instance, VectorStore):
    # do something vector specific like create an index
```

client/tests/test_online_providers.py
```diff
@@ -0,0 +1,80 @@
...
+import pytest
+import uuid
+from datetime import datetime
+
+
+@pytest.mark.parametrize(
+    "provider",
+    ["file"],
+)
+class TestOnlineProvider:
```
epps, line 12: This might be a nit, but @ahmadnazeri was suggesting we move away from the class-based test pattern.

client/src/featureform/providers/online_store.py
aolfat: yeah, just take advantage of the abstraction. Could also have a static method or a function that does it: `isVectorStore(store)`.Making it part of the interface is too much boilerplate.

client/src/featureform/providers/online_store.py
```diff
@@ -0,0 +1,28 @@
...
+        raise NotImplementedError()
+
+
+class OnlineTable:
```
aolfat, line 15: I wonder if it makes sense to use python's contextlibmanager for these so that we can do something like. Depends on @epps 's usage

```
with PineconeTable() as p:
   ...
```

to do that we can add the __enter__ and __exit__ methods here (https://realpython.com/python-with-statement/#the-with-statement-approach).

client/examples/local_serving.py
sdreyer: This is for my own testing, will change later before we merge

client/src/featureform/enums.py
sdreyer: I hate this but sure

client/src/featureform/providers/filestore.py
sdreyer: All the pathing stuff needs to change, this is temporary bc i need to interface merged

client/src/featureform/providers/filestore.py
sdreyer: Dont think so

client/src/featureform/providers/filestore.py
sdreyer: this function not needed

client/src/featureform/providers/filestore.py
sdreyer: Would like to leave it. It differentiates whether there was a failure with getting the table itself, or if it was a failure with getting metadata around the table

client/src/featureform/providers/filestore.py
sdreyer: Yeah this converts featureform types to pandas types. we want to be strict on what gets converted to what

client/src/featureform/providers/online_store.py
```diff
@@ -0,0 +1,28 @@
+class OnlineStore:
```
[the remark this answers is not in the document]
sdreyer: done

client/src/featureform/providers/providers.py
```diff
@@ -0,0 +1,12 @@
+from .filestore import LocalFileStore
+from .online_store import OnlineStore
+
+
+def get_provider(provider) -> OnlineStore:
```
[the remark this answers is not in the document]
sdreyer: Its a lookup that returns the specified provider based on name

client/tests/test_online_providers.py
sdreyer: Whats the alternative. I have nested params I want to go through

client/src/featureform/providers/online_store.py
sdreyer: Ill come back to this, really just need something basic working for now

PR #869, opened 2023-06-14
Added ability to read entire directory

sdreyer:
# Description

using `ff.register_directory()` will read all files from a directory into a dataframe

### Select type(s) of change

- [ ] Documentation update

# Checklist:


client/src/featureform/enums.py
```diff
@@ -51,6 +51,7 @@ class OperationType(Enum):
 @dataclass
 class SourceType(str, Enum):
     PRIMARY_SOURCE = "PRIMARY"
+    DIRECTORY = "DIRECTORY"
```
aolfat, line 54: what's the diff between primary and directory? is primary just a regular file?

client/src/featureform/local_cache.py
```diff
@@ -181,11 +181,19 @@ def get_source_files_for_source(self, source_name, source_variant) -> Set[str]:
             dependencies = json.loads(source["inputs"])
             for name, variant in dependencies:
                 sources.update(self.get_source_files_for_source(name, variant))
+        elif transform_type == SourceType.DIRECTORY.value:
+            path = source["definition"]
+            for absolute_file, _ in self._absolute_file_paths(path):
+                sources.update([absolute_file])
```
aolfat, line 187: nit: just do sources.add(absolute_file)

client/src/featureform/local_cache.py
```diff
@@ -181,11 +181,19 @@ def get_source_files_for_source(self, source_name, source_variant) -> Set[str]:
             dependencies = json.loads(source["inputs"])
             for name, variant in dependencies:
                 sources.update(self.get_source_files_for_source(name, variant))
+        elif transform_type == SourceType.DIRECTORY.value:
+            path = source["definition"]
```
aolfat, line 185: super nit: 

directory = source["definition"]
for file in self.get_files_in_dir(directory):
   sources.add(file)

client/src/featureform/resources.py
```diff
@@ -876,13 +885,14 @@ class Source:
     tags: list
     properties: dict
     variant: str
+    is_directory: bool = False
```
aolfat, line 888: this is fine for now, but i think eventually we want abstractions so we don't need all these checks

client/src/featureform/local_cache.py
```diff
@@ -181,11 +181,19 @@ def get_source_files_for_source(self, source_name, source_variant) -> Set[str]:
...
+                sources.update([absolute_file])
         else:
             raise Exception(f"Unknown source type: {transform_type}")
-
         return sources
 
+    def _absolute_file_paths(self, directory):
+        for dirpath, _, filenames in os.walk(directory):
+            for f in filenames:
+                yield os.path.abspath(os.path.join(dirpath, f)), f
```
aolfat, line 195: nit: probz don't need the the second return var

client/src/featureform/register.py
```diff
@@ -571,6 +573,67 @@ def register_file(
...
+        owner="",
+        tags: List[str] = [],
+        properties: dict = {},
+    ):
+        """Register a directory.
+        When registering a directory, files can be interacted with as a table with columns "filename" and "body".
+        For example:
+        filename                  |            body
+        --------------------------|---------------------------------------
+        featureform_docs.txt      | Featureform allows data scientists....
```
aolfat, line 591: I wonder what the df constraints are for a column. If you have a large ass file will that fit in the column. Something to think about

client/src/featureform/register.py
```diff
@@ -571,6 +573,67 @@ def register_file(
...
+        if owner == "":
+            owner = self.__registrar.must_get_default_owner()
+        if variant == "":
+            variant = self.__registrar.get_run()
+        # Store the file as a source
+        self.__registrar.register_primary_data(
+            name=name,
+            variant=variant,
+            location=Directory(path),
+            is_directory=True,
```
aolfat, line 626: I feel as though instead of adding a "is_directory" one could just check if location is a directory

client/src/featureform/serving.py
```diff
@@ -387,10 +387,41 @@ def get_input_df(self, source_name, source_variant):
...
+
         else:
             df = self.process_transformation(source_name, source_variant)
         return df
 
+    def read_directory(self, directory):
+        if not os.path.isdir(directory):
+            raise Exception(f"Path {directory} is not a directory")
+
+        def absolute_file_paths(directory):
```
aolfat, line 406: if we're using this function in multiple places we can add this somewhere else instead of copying it over to the cache as well. file_utils.py sounds like it could work. or even a utils.

client/src/featureform/serving.py
```diff
@@ -387,10 +387,41 @@ def get_input_df(self, source_name, source_variant):
...
+        file_names = []
+        file_body = []
+        for absolute_fn, relative_fn in absolute_file_paths(directory):
+            file_names.append(relative_fn)
+            f = open(absolute_fn, "r")
+            try:
+                file_body.append(f.read())
+            except Exception as e:
+                raise Exception(
+                    f"Cannot read file {absolute_fn}: {e}\nFiles must be text files"
```
aolfat, line 420: This error won't always throw due to it being a text file, could also throw if there aren't permissions.

client/src/featureform/serving.py
```diff
@@ -387,10 +387,41 @@ def get_input_df(self, source_name, source_variant):
...
+        def absolute_file_paths(directory):
+            for dirpath, _, filenames in os.walk(directory):
+                for f in filenames:
+                    yield os.path.abspath(os.path.join(dirpath, f)), f
+
+        file_names = []
+        file_body = []
+        for absolute_fn, relative_fn in absolute_file_paths(directory):
+            file_names.append(relative_fn)
+            f = open(absolute_fn, "r")
```
aolfat, line 415: use with statement here

aolfat, line 415: with open(absolute_fn, "r") as f:

client/src/featureform/serving.py
aolfat, line 420: you're capturing all generic exceptions into one here. Do we want to fail the whole thing if one of the files is corrupt/not a text file or do we want to skip over it. 

Also, feels like this check should be done during the register.

Could also do it like this:

```
    try:
        file_data = [
            (relative_fn, Path(absolute_fn).read_text())
            for absolute_fn, relative_fn in absolute_file_paths(directory)
        ]
    except Exception as e:
        raise IOError(f"Cannot read file {absolute_fn}: {e}")

    file_names, file_bodies = zip(*file_data)
    df = pd.DataFrame(data={"filename": file_names, "body": file_bodies})

    return df
```

if you use read_text you don't have to open it but you have to wrap it in Path, it takes care of it for you.

client/src/featureform/register.py
aolfat: but minor

client/src/featureform/enums.py
aolfat: coming back to this, I feel like directory would be a type of primary source

aolfat (approved):
if it works, it works lol. I'm approving

client/src/featureform/local_cache.py
sdreyer: done

client/src/featureform/local_cache.py
sdreyer: changed

client/src/featureform/register.py
sdreyer: This is limited by RAM

client/src/featureform/serving.py
sdreyer: Updated

client/src/featureform/serving.py
sdreyer: fixed

client/src/featureform/resources.py
sdreyer: Changed this to just check the type instead

client/src/featureform/enums.py
sdreyer: yeah same, not a good format for this atm though

PR #894, opened 2023-06-17
Entities Without Labels

epps:
# Description

This PR introduces a fix for an assumption in the localmode code (i.e. entities must have associated labels.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/dashboard_metadata.py
+++ b/client/src/featureform/dashboard_metadata.py
@@ -381,9 +381,13 @@ def labels(rowData):
 
 
 def entities(rowData):
-    label_list = sqlObject.query_resource_variant(
-        "label_variant", "entity", rowData["name"]
-    )
+    try:
+        label_list = sqlObject.query_resource_variant(
+            "label_variant", "entity", rowData["name"]
+        )
+    except Exception as e:
+        print(f"No labels found for entity {rowData['name']} - {e}")
+        label_list = []
     training_set_list = set()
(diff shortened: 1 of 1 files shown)


client/src/featureform/dashboard_metadata.py
```diff
@@ -381,9 +381,13 @@ def labels(rowData):
...
 def entities(rowData):
-    label_list = sqlObject.query_resource_variant(
-        "label_variant", "entity", rowData["name"]
-    )
+    try:
+        label_list = sqlObject.query_resource_variant(
+            "label_variant", "entity", rowData["name"]
+        )
+    except Exception as e:
+        print(f"No labels found for entity {rowData['name']} - {e}")
```
ahmadnazeri, line 389: should we throw an error? If yes, can we use the `LabelNotFound`?

ahmadnazeri (approved):

RedLeader16 (approved):

client/src/featureform/dashboard_metadata.py
RedLeader16: I'd say no because in the "valid" scenario where we don't have a label registered, the entity list will throw a 500 instead of gracefully returning an empty list.

RedLeader16: So it needs to be a "soft" exception as I understand the issue.

PR #907, opened 2023-07-05
Resource Tag Updates

RedLeader16:
# Description

This PR allows the creation and updates for resources tags in the dashboard. 
Each resource variant has it's own separate set of tags.

It includes POST endpoints in the metadata layer, and chip functionality in the FE.

`example` 

https://github.com/featureform/featureform/assets/34227093/530a7a36-c954-48ec-87ee-b15ec1175231

Unit Tests:
`FE`
Screenshot 2023-07-06 at 4 19 49 PM

`BE`
Screenshot 2023-07-05 at 3 30 33 PM

### Select type(s) of change

- [ ] Documentation update

# Checklist:


metadata/dashboard/dashboard_metadata.go
```diff
@@ -1074,16 +1121,133 @@ func (m *MetadataServer) getSourceDataIterator(name, variant string, limit int64
...
+	case "providers":
+		foundVariant, err := m.client.GetProvider(context.Background(), name)
+		m.SetFoundVariantJSON(foundVariant, err, c, resourceType)
+	}
+}
+
+type TagRequestBody struct {
+	Tags []string `json:"tags"`
+}
+
```
RedLeader16, line 1193: will make PostTags generic next. kept it as sourceVariant for quicker iteration.

RedLeader16: made generic using a switch. exploring using interfaces/structs to cut down on the boilerplate but Go is being a diva.

metadata/dashboard/dashboard_metadata.go
```diff
@@ -12,32 +12,79 @@ import (
...
-	logger *zap.SugaredLogger
+	lookup          metadata.ResourceLookup
+	client          *metadata.Client
+	logger          *zap.SugaredLogger
+	StorageProvider StorageProvider
 }
 
 func NewMetadataServer(logger *zap.SugaredLogger, client *metadata.Client) (*MetadataServer, error) {
 	logger.Debug("Creating new metadata server")
+	etcdHost := help.GetEnv("ETCD_HOST", "featureform-etcd")
```
sdreyer, line 70: Typically we want to pass configs from a top level main function to this function (add a storage provider param), so we dont have env variables in the function itself

metadata/dashboard/dashboard_metadata.go
```diff
@@ -1074,16 +1121,219 @@ func (m *MetadataServer) getSourceDataIterator(name, variant string, limit int64
 	return primary.IterateSegment(limit)
 }
 
+type VariantDuck interface {
```
sdreyer, line 1124: whats a duck?

metadata/dashboard/dashboard_metadata.go
```diff
@@ -1074,16 +1121,219 @@ func (m *MetadataServer) getSourceDataIterator(name, variant string, limit int64
...
+	c.JSON(http.StatusOK, TagResult{
+		Name:    name,
+		Variant: variant,
+		Tags:    requestBody.Tags,
+	})
+}
+
+func replaceTags(resourceTypeParam string, currentResource metadata.Resource, newTagList *pb.Tags) error {
+	deserialized := currentResource.Proto()
+	switch resourceTypeParam {
```
sdreyer, line 1266: Feel like this could be condensed with an interface. Ok for now but may want to rework if theres time

metadata/dashboard/dashboard_metadata.go
RedLeader16: just a reference to duck typing, I wasn't sure what to call the interface to also denote it's sole purpose is to just be a common result for all variants. maybe `VariantResult`?

metadata/dashboard/dashboard_metadata.go
RedLeader16: 100% yeah looking for a clean way to handle this. The issue is that interfaces in go only define _behavior_ and not properties. In this case I'd need common access to `*variant.Tags` which is a property in all the variant structs, so a quick interface won't work. 

Alternatively...I could write a method on all the variants, `setTags([]string)` and then create a single method interface, which gets rid of the switch but adds more boilerplate to metadata.go :(

Open to any ideas!

metadata/dashboard/dashboard_metadata.go
RedLeader16: gotcha, reworking

metadata/dashboard/dashboard_metadata.go
RedLeader16: swapped to `VariantResult`

metadata/client.go
```diff
@@ -1536,6 +1536,22 @@ type User struct {
 	fetchPropertiesFn
 }
 
+func (e Entity) Variant() string {
```
aolfat, line 1539: shouldn't this functions be ordered along with their superclass? So shouldn't the e entity variant be with th Entity class?

RedLeader16: yeah good call out. I'll place them closer to their parent.

RedLeader16: done.

metadata/metadata.go
```diff
@@ -337,34 +337,34 @@ func (lookup localResourceLookup) SetSchedule(id ResourceID, schedule string) er
...
+	return make(LocalResourceLookup), nil
 }
 
-func (resource *sourceResource) Proto() proto.Message {
+func (resource *SourceResource) Proto() proto.Message {
 	return resource.serialized
 }
 
-func (this *sourceResource) Notify(lookup ResourceLookup, op operation, that Resource) error {
+func (this *SourceResource) Notify(lookup ResourceLookup, op operation, that Resource) error {
```
aolfat, line 367: super nit: i know you didn't write this but since you're refactoring, golang best practices is to use the same receiver name everywhere. This one's kinda special because it uses "this" and "that" to differentiate the two. I'd still probably keep it consistent for readability but super minor just wanted to point it out

RedLeader16: I noticed that warning too, but wasn't sure if changing it would wreck something. Looks like it's safe to change. Swapping now since it's quick...I think. 
image

RedLeader16: ok renamed the `that` param to `resourceParam` and renamed the `this` receiver to `resource`.

metadata/dashboard/dashboard_metadata.go
```diff
@@ -1074,16 +1112,219 @@ func (m *MetadataServer) getSourceDataIterator(name, variant string, limit int64
...
+	}
+	foundResource, err := m.lookup.Lookup(objID)
+
+	if err != nil {
+		fetchError := m.GetTagError(400, err, c, "PostTags - Error finding the resource with resourceID")
+		c.JSON(fetchError.StatusCode, fetchError.Error())
+		return
+	}
+
+	replaceTags(resourceTypeParam, foundResource, &pb.Tags{Tag: requestBody.Tags})
```
aolfat, line 1244: I agree with sterlz comment about the interface. I imagine this could jst look like. 

`foundResource.Tags.replace(requestBody.Tags)`

aolfat: how that's done I'm not sure yet lol

metadata/metadata.go
RedLeader16: I triple checked but give it a scan when you can.

sdreyer (approved):

metadata/dashboard/dashboard_metadata.go
RedLeader16: Agreed. I'll add it to my list to fix next week. I'll resolve for now.

PR #947, opened 2023-08-05
Cloud Storage Pathing (Azure Blob Storage)

epps:
# Description

This PR addresses issues around how we handle paths in cloud storage (e.g. Azure Blob Storage) and begins to extend the use of the `Filepath` interface to address these issues. There remains much work to do to fully harden the code base to pathing error, for example:

* refactor store methods that construct paths based on IDs (e.g. `PathWithPrefix`)
* document and cover all cases that could result in pathing issues (i.e. for S3 and GCS)
* begin to standardize how the code base handles paths (e.g. `root_path1` param in the client)

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/client.py
+++ b/client/src/featureform/client.py
@@ -54,6 +54,7 @@ class Client(ResourceClient, ServingClient):
         source: Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str],
         variant: Union[str, None] = None,
         limit=NO_RECORD_LIMIT,
+        asynchronous=False,
     ):
         """
         Compute a dataframe from a registered source or transformation
@@ -62,6 +63,7 @@ class Client(ResourceClient, ServingClient):
             source (Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str]): The source or transformation to compute the dataframe from
             variant (str): The source variant; defaults to a Docker-style random name and is ignored if source argument is not a string
             limit (int): The maximum number of records to return; defaults to NO_RECORD_LIMIT
+            asynchronous (bool): Whether to wait for the apply to finish; defaults to False to ensure sources are ready
 
         **Example:**
         ```py title="definitions.py"
@@ -70,7 +72,7 @@ class Client(ResourceClient, ServingClient):
         avg_user_transaction_df = transactions_df.groupby("CustomerID")["TransactionAmount"].mean()
         ```
         """
-        self.apply()
+        self.apply(asynchronous=asynchronous)
         if isinstance(
             source, (SourceRegistrar, LocalSource, SubscriptableTransformation)
         ):
--- a/client/src/featureform/client.py
+++ b/client/src/featureform/client.py
@@ -63,7 +63,7 @@ class Client(ResourceClient, ServingClient):
             source (Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str]): The source or transformation to compute the dataframe from
             variant (str): The source variant; defaults to a Docker-style random name and is ignored if source argument is not a string
             limit (int): The maximum number of records to return; defaults to NO_RECORD_LIMIT
-            asynchronous (bool): Whether to wait for the apply to finish; defaults to False to ensure sources are ready
+            asynchronous (bool): @param asynchronous: Flag to determine whether the client should wait for resources to be in either a READY or FAILED state before returning. Defaults to False to ensure that newly registered resources are in a READY state prior to serving them as dataframes.
 
         **Example:**
         ```py title="definitions.py"
--- a/coordinator/Dockerfile.old
+++ b/coordinator/Dockerfile.old
@@ -67,15 +67,15 @@ RUN curl https://pyenv.run | bash
 
 ## Install Python versions
 ARG TESTING
-# RUN if [ "$TESTING" = "True" ]; then \
-#   pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
-#   else \
-#   pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-#   pyenv install 3.8.16 && pyenv global 3.8.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-#   pyenv install 3.9.16 && pyenv global 3.9.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-#   pyenv install 3.10.10 && pyenv global 3.10.10 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-#   pyenv install 3.11.2 && pyenv global 3.11.2 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
-#   fi
+RUN if [ "$TESTING" = "True" ]; then \
+  pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
+  else \
+  pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+  pyenv install 3.8.16 && pyenv global 3.8.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+  pyenv install 3.9.16 && pyenv global 3.9.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+  pyenv install 3.10.10 && pyenv global 3.10.10 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+  pyenv install 3.11.2 && pyenv global 3.11.2 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
+  fi
 
 ENV SPARK_SCRIPT_PATH="/app/provider/scripts/spark/offline_store_spark_runner.py"
 ENV PYTHON_INIT_PATH="/app/provider/scripts/spark/python_packages.sh"
--- a/coordinator/Dockerfile.old
+++ b/coordinator/Dockerfile.old
@@ -67,15 +67,15 @@ RUN curl https://pyenv.run | bash
 
 ## Install Python versions
 ARG TESTING
-RUN if [ "$TESTING" = "True" ]; then \
-  pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
-  else \
-  pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-  pyenv install 3.8.16 && pyenv global 3.8.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-  pyenv install 3.9.16 && pyenv global 3.9.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-  pyenv install 3.10.10 && pyenv global 3.10.10 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
-  pyenv install 3.11.2 && pyenv global 3.11.2 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
-  fi
+# RUN if [ "$TESTING" = "True" ]; then \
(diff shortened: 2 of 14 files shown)


serving/serving.go
```diff
@@ -153,10 +153,14 @@ func (serv *FeatureServer) getTrainingSetIterator(name, variant string) (provide
...
 	serv.Logger.Infow("Getting Source Variant Iterator", "name", name, "variant", variant)
-	sv, err := serv.Metadata.GetSourceVariant(ctx, metadata.NameVariant{name, variant})
+	sv, err := serv.Metadata.GetSourceVariant(ctx, metadata.NameVariant{Name: name, Variant: variant})
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get source variant")
 	}
+	// TODO: Determine if we want to add a backoff here to wait for the source
+	if sv.Status() != metadata.READY {
+		return nil, fmt.Errorf("source variant is not ready")
+	}
```
epps, line 163: @sdreyer, we should talk about how we want to handle the resource status. I added this here and changed code in `client.py` to ensure I wasn't trying to fetch a resource before it was ready, which is especially important in the case of transformation sources.

For your consideration, here is the code I changed in `client.py`:

```py
def dataframe(
    self,
    source: Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str],
    variant: Union[str, None] = None,
    limit=NO_RECORD_LIMIT,
    synchronous=True,
):
    # ... method docstring ...
    self.apply(synchronous=synchronous)
    # ... rest of method logic ...
    return self.impl._get_source_as_df(name, variant, limit)
```

epps: You'll note that I changed the param name in `apply` to synchronous because it was somewhat confusing based on the existing naming and documentation. We can chat more about whether we want to include this change as is, but we'll need to handle the resource state somewhere.

provider/k8s.go
```diff
@@ -1309,7 +1354,13 @@ func (k8s *K8sOfflineStore) getResourceInformationFromFilePath(path string) (str
 func (k8s *K8sOfflineStore) GetTransformationTable(id ResourceID) (TransformationTable, error) {
 	transformationPath := k8s.store.PathWithPrefix(fileStoreResourcePath(id), false)
 	k8s.logger.Debugw("Retrieved transformation source", "ResourceId", id, "transformationPath", transformationPath)
-	return &FileStorePrimaryTable{k8s.store, transformationPath, true, id}, nil
+	filePath, err := NewFilepath(k8s.store.FilestoreType(), "", "", transformationPath)
```
sdreyer, line 1357: Why do we do this? Should always require a bucket

provider/spark.go
```diff
@@ -1463,7 +1463,14 @@ func (d *DatabricksExecutor) GetDFArgs(outputURI string, code string, sources []
 func (spark *SparkOfflineStore) GetTransformationTable(id ResourceID) (TransformationTable, error) {
 	transformationPath := spark.Store.PathWithPrefix(fileStoreResourcePath(id), false)
 	spark.Logger.Debugw("Retrieved transformation source", "ResourceID", id, "transformationPath", transformationPath)
-	return &FileStorePrimaryTable{spark.Store, transformationPath, true, id}, nil
+	// Unlike a primary source, which comes across with the fully specified storage URI, transformations are relative paths
+	// stored in the feature/ dir
+	filePath, err := NewFilepath(spark.Store.FilestoreType(), "", "", transformationPath)
```
sdreyer, line 1468: We may want to consider a different funciton for this usage

provider/filepath.go
```diff
@@ -133,19 +145,13 @@ func (azure *AzureFilepath) FullPathWithBucket() string {
...
 func (azure *AzureFilepath) ParseFullPath(fullPath string) error {
 	abfssRegex := regexp.MustCompile(`abfss://(.+?)@(.+?)\.dfs.core.windows.net/(.+)`)
-	matches := abfssRegex.FindStringSubmatch(fullPath)
-
-	// If the regex matches all parts of the ABFS path, then we can parse the
-	// bucket, storage account, and path components. Otherwise, we can just set
-	// the path for standard Azure Blob Storage paths.
-	if len(matches) == 4 {
+	if matches := abfssRegex.FindStringSubmatch(fullPath); len(matches) != 4 {
+		return fmt.Errorf("could not parse full path '%s'", fullPath)
```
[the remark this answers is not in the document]
epps: This is great feedback. I'm not sure about the trailing slash, so I'll check and update the regex. I need to add tests around this, but I wanted to arrive at a working solution given the urgency.

provider/k8s.go
```diff
@@ -966,16 +987,30 @@ func (tbl *FileStorePrimaryTable) Write(GenericRecord) error {
...
+	// The length of keyParts is 1 if the key is a paht to a directory. This case is invalid
+	// in the case of a primary table; however, we expect this case in the case of a transformation.
+	if len(keyParts) == 1 && !tbl.isTransformation {
 		return nil, fmt.Errorf("expected a file but got a directory: %s", keyParts[0])
+	} else {
+		// The file structure in cloud storage for transformations is /featureform/Transformation/<NAME>/<VARIANT>
+		// but there is an additional directory that's named using a timestamp that contains the transformation file
+		// we need to access. NewestFileOfType will recursively search for the newest file of the given type (i.e.
+		// parquet) given a path (i.e. `key`).
+		filename, err := tbl.store.NewestFileOfType(key, Parquet) // TODO: determine if we can actually assume Parquet here
```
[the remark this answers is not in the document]
epps: No, we can read other types (e.g. CSV); however, I'm not certain how we can "know" the file extension at this point. Let's discuss when you have the time.

provider/filepath.go
```diff
@@ -133,19 +145,13 @@ func (azure *AzureFilepath) FullPathWithBucket() string {
...
+	} else {
 		azure.filePath.bucket = matches[1]
```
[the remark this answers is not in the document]
epps: Good catch. Just added some tests to cover this case.

client/src/featureform/client.py
```diff
@@ -62,6 +63,7 @@ def dataframe(
             source (Union[SourceRegistrar, LocalSource, SubscriptableTransformation, str]): The source or transformation to compute the dataframe from
             variant (str): The source variant; defaults to a Docker-style random name and is ignored if source argument is not a string
             limit (int): The maximum number of records to return; defaults to NO_RECORD_LIMIT
+            asynchronous (bool): Whether to wait for the apply to finish; defaults to False to ensure sources are ready
```
ahmadnazeri, line 66: this message can be confusing. I would assume if I provide it True then it would wait.

provider/spark.go
```diff
@@ -1461,9 +1461,18 @@ func (d *DatabricksExecutor) GetDFArgs(outputURI string, code string, sources []
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create empty filepath due to error %w (store type: %s; path: %s)", err, spark.Store.FilestoreType(), transformationPath)
+	}
+	err = filePath.ParseFullPath(transformationPath)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse path due to error %w (store type: %s; path: %s)", err, spark.Store.FilestoreType(), transformationPath)
+	}
+	spark.Logger.Debugw("Retrieved transformation source", "id", id, "filePath", filePath)
+	return &FileStorePrimaryTable{spark.Store, filePath, true, id}, nil
 }
```
epps, line 1476: @aolfat, let's discuss this tomorrow.

ahmadnazeri (approved):

provider/filepath_test.go
```diff
@@ -201,6 +207,20 @@ func TestParseFullPath(t *testing.T) {
...
 			ExpectedError:   nil,
 		},
+		"AzureBlobStoreFilePathWithTrailingSlash": {
+			StoreType: Azure,
+			FullPath:  "abfss://container@account.dfs.core.windows.net/elasticmapreduce/path/to/file/",
+			ExpectedPath: &AzureFilepath{
+				storageAccount: "account",
+				filePath: filePath{
+					bucket: "container",
+					prefix: "",
```
sdreyer, line 217: this test and the one above have different prefixes

provider/filepath_test.go
```diff
@@ -261,3 +281,57 @@ func TestParseFullPath(t *testing.T) {
...
+
+	tests := map[string]TestCase{
+		"AzureBlobStoreFilePathWithTrailingSlash": {
+			StoreType: Azure,
+			FullPath:  "abfss://container@account.dfs.core.windows.net/elasticmapreduce/path/to/file/",
+			ExpectedPath: &AzureFilepath{
+				storageAccount: "account",
+				filePath: filePath{
+					bucket: "container",
+					prefix: "",
```
sdreyer, line 302: id the prefix just abfss://?

provider/k8s.go
```diff
@@ -496,11 +497,28 @@ func (store *genericFileStore) DeleteAll(dir string) error {
...
-	err := store.bucket.WriteAll(context.TODO(), key, data, nil)
+	ctx := context.TODO()
+	err := store.bucket.WriteAll(ctx, key, data, nil)
 	if err != nil {
 		return err
 	}
-	return nil
+	err = re.Do(
+		func() error {
+			blob, errRetr := store.bucket.ReadAll(ctx, key)
```
sdreyer, line 507: This may be a lot if the file is large, but well see when we get there

provider/k8s.go
```diff
@@ -967,16 +986,33 @@ func (tbl *FileStorePrimaryTable) Write(GenericRecord) error {
...
 func (tbl *FileStorePrimaryTable) GetName() string {
-	return tbl.sourcePath
+	return tbl.source.FullPathWithBucket()
 }
 
 func (tbl *FileStorePrimaryTable) IterateSegment(n int64) (GenericTableIterator, error) {
-	path := tbl.sourcePath
-	keyParts := strings.Split(path, ".")
+	key := tbl.source.Path()
+	keyParts := strings.Split(key, ".")
```
sdreyer, line 994: Can we add this logic to the filepath?

provider/k8s.go
```diff
@@ -967,16 +986,33 @@ func (tbl *FileStorePrimaryTable) Write(GenericRecord) error {
...
+		// but there is an additional directory that's named using a timestamp that contains the transformation file
+		// we need to access. NewestFileOfType will recursively search for the newest file of the given type (i.e.
+		// parquet) given a path (i.e. `key`).
+		filename, err := tbl.store.NewestFileOfType(key, Parquet)
+		if err != nil {
+			return nil, fmt.Errorf("could not find newest file of type %s: %w", Parquet, err)
+		}
+		// We need to update key and keyParts with the result of NewestFileOfType for the Read below to succeed.
+		key = filename
+		keyParts = strings.Split(key, ".")
```
sdreyer, line 1012: lets try to do any splitting in the filepath object

provider/k8s.go
epps: Agreed. I can check the docs again to see if there's another method that might give us the same assurance without reading in the entire contents. One more thought: I wonder if this is _as_ important given the client-side changes I put in place to ensure the call to `apply` in `client.dataframe` is synchronous and waits for the resource to either be in a Ready or Failed state. Thouhts?

client/src/featureform/client.py
epps: I agree. I was mirroring the naming and messaging in `apply`, but I can improve both to make things clearer.

coordinator/Dockerfile.old
```diff
@@ -67,15 +67,15 @@ RUN curl https://pyenv.run | bash
...
-  fi
+# RUN if [ "$TESTING" = "True" ]; then \
+#   pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
+#   else \
+#   pyenv install 3.7.16 && pyenv global 3.7.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#   pyenv install 3.8.16 && pyenv global 3.8.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#   pyenv install 3.9.16 && pyenv global 3.9.16 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#   pyenv install 3.10.10 && pyenv global 3.10.10 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt && \
+#   pyenv install 3.11.2 && pyenv global 3.11.2 && pyenv exec pip install --upgrade pip && pyenv exec pip install -r /app/provider/scripts/spark/requirements.txt ; \
+#   fi
```
epps, line 78: Restore prior to merge.

coordinator/coordinator.go
```diff
@@ -533,9 +533,11 @@ func (c *Coordinator) runDFTransformationJob(transformSource *metadata.SourceVar
 		return fmt.Errorf("map name: %v sources: %v", err, sources)
 	}
 
-	sourceMapping := []provider.SourceMapping{}
-	for nameVariantClient, transformationTableName := range sourceMap {
-		sourceMapping = append(sourceMapping, provider.SourceMapping{Template: nameVariantClient, Source: transformationTableName})
+	sourceMapping := make([]provider.SourceMapping, len(sources))
```
sdreyer, line 536: Can you split this into a function and add a test for it?

provider/spark.go
```diff
@@ -1595,26 +1616,42 @@ func sparkTrainingSet(def TrainingSetDef, spark *SparkOfflineStore, isUpdate boo
...
+		// prior to adding it to the list of source paths
+		err = filepath.ParseFullPath(featureSourcePath)
 		if err != nil {
-			spark.Logger.Errorw("Could not get latest feature file", "feature", feature, "error", err)
-			return fmt.Errorf("could not get latest feature file: %v", err)
+			featureSourcePath = spark.Store.PathWithPrefix(featureSchema.SourceTable, true)
 		}
-		featurePath := spark.Store.PathWithPrefix(latestFeaturePath, true)
-		sourcePaths = append(sourcePaths, featurePath)
+		spark.Logger.Debugw("================>>>>> FEATURE SOURCE TABLE", "featureSourcePath", featureSourcePath)
```
sdreyer, line 1653: This need to be here?

sdreyer (approved):

PR #956, opened 2023-08-12
Check if source exists in k8

aolfat:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/provider/filestore.go
+++ b/provider/filestore.go
@@ -675,3 +675,22 @@ func (fs *HDFSFileStore) FilestoreType() pc.FileStoreType {
 func (fs *HDFSFileStore) AddEnvVars(envVars map[string]string) map[string]string {
 	panic("HDFS Filestore is not supported for K8s at the moment.")
 }
+
+// ExistsByUrl checks if a source url i.e "abfss://test@testingstoragegen.dfs.core.windows.net/featureform/tests/ice_cream.parquet" exists in the file store.
+func ExistsByUrl(store FileStore, sourceUrl string) (bool, error) {
+	filePathParts, err := NewEmptyFilepath(store.FilestoreType())
+	if err != nil {
+		return true, fmt.Errorf("could not create empty file path: %w", err)
+	}
+	if err := filePathParts.ParseFullPath(sourceUrl); err != nil {
+		return true, fmt.Errorf("could not parse full path: %w", err)
+	}
+	sourceExists, err := store.Exists(filePathParts.Path())
+	if err != nil {
+		return true, fmt.Errorf("error checking if source exists in store: %w", err)
+	}
+	if !sourceExists {
+		return false, nil
+	}
+	return true, nil
+}
--- a/provider/filestore_test.go
+++ b/provider/filestore_test.go
@@ -8,15 +8,11 @@
 package provider
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"github.com/google/uuid"
+	"github.com/joho/godotenv"
 	"os"
 	"testing"
-
-	"github.com/joho/godotenv"
-
-	pc "github.com/featureform/provider/provider_config"
 )
 
 func TestFileStore(t *testing.T) {
@@ -25,7 +21,7 @@ func TestFileStore(t *testing.T) {
 		fmt.Println(err)
 	}
 
-	fileStores, err := getFileStores()
+	fileStores := getFileStores(t)
 	if err != nil {
 		t.Fatalf("could not get filestores: %s", err)
 	}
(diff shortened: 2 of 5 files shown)


provider/k8s_test.go
```diff
@@ -977,3 +1025,67 @@ func TestParquetIterator_vector32(t *testing.T) {
...
+	testLogger := zaptest.NewLogger(t).Sugar()
+
+	tests := []struct {
+		name                 string
+		store                FileStore
+		nonExistantSourceUrl string
+		expectedErrorMessage string
+	}{
+		{"Azure", createAzureFileStore(t), "abfss://test@testingstoragegen.dfs.core.windows.net/featureform/tests/this_doesnt_exist.parquet", "source does not exist at path: featureform/tests/this_doesnt_exist.parquet"},
+		//{"S3", createS3FileStore(t), "s3://featureform-spark-testing/featureform/tests/this_doesnt_exist.parquet", "source does not exist at path: featureform/tests/this_doesnt_exist.parquet"},
```
aolfat, line 1041: could not get this to properly work, need someone to look at this with me tomorrow

provider/k8s_test.go
```diff
@@ -977,3 +1025,37 @@ func TestParquetIterator_vector32(t *testing.T) {
...
+	tests := []struct {
+		name                 string
+		store                FileStore
+		workingUrl           string
+		nonExistantSourceUrl string
+	}{
+		{"Azure", createAzureFileStore(t), "abfss://test@testingstoragegen.dfs.core.windows.net/featureform/tests/ice_cream.parquet", "abfss://test@testingstoragegen.dfs.core.windows.net/featureform/tests/this_doesnt_exist.parquet"},
+		{"S3", createS3FileStore(t), "s3://featureform-spark-testing/featureform/tests/this_doesnt_exist.parquet", "source does not exist at path: featureform/tests/this_doesnt_exist.parquet"},
+	}
+	for _, tt := range tests {
```
sdreyer, line 1039: Add ability to upload a file

sdreyer (approved):

provider/filestore.go
```diff
@@ -663,3 +663,24 @@ func (fs *HDFSFileStore) FilestoreType() pc.FileStoreType {
 func (fs *HDFSFileStore) AddEnvVars(envVars map[string]string) map[string]string {
 	panic("HDFS Filestore is not supported for K8s at the moment.")
 }
+
+/**
+ * Checks if the source url i.e "abfss://test@testingstoragegen.dfs.core.windows.net/featureform/tests/ice_cream.parquet" exists in the file store.
+ */
+func doesSourceUrlExist(store FileStore, sourceUrl string) (bool, error) {
```
ahmadnazeri, line 670: Is there a reason why this is a standalone function instead of part of the FileStore?

provider/filestore_test.go
```diff
@@ -168,147 +185,60 @@ func testPathPrefix(t *testing.T, store FileStore, storeName string) {
...
-	serializedConfig, err := hdfsConfig.Serialize()
-	if err != nil {
-		return nil, fmt.Errorf("could not serialized hdfs config: %s", err)
-	}
-	fileStore, err := NewHDFSFileStore(serializedConfig)
+func TestDoesSourceUrlExist(t *testing.T) {
+	err := godotenv.Load("../.env")
 	if err != nil {
-		return nil, fmt.Errorf("could not initialize hdfs filestore: %s", err)
+		t.Fatalf("could not load env: %s", err)
```
ahmadnazeri, line 191: I wouldn't do this because the CI/CD doesn't have a `.env` file so it will fail.

provider/k8s_test.go
```diff
@@ -122,56 +122,13 @@ func TestBlobInterfaces(t *testing.T) {
...
-		Host:     "localhost",
-		Port:     "9000",
-		Username: "hduser",
-	}
-
-	serializedHDFSConfig, err := hdfsConfig.Serialize()
-	if err != nil {
-		t.Fatalf("failed to create serialize hdfs blob store: %v", err)
+	if err := godotenv.Load("../.env"); err != nil {
+		t.Fatalf("could not load env file: %v", err)
```
ahmadnazeri, line 126: would this fail if `.env` file doesn't exist? if yes, then it will fail in CI/CD.

provider/filestore.go
aolfat: It's a good question. I chose not to do for because I like the core interface functions on the interface. I don't want to bloat the interface with a ExistsByUrl since this feels more like a utility function and also since getting the file parts by url works only for azure, gcs, and s3 right now. I could be convinced otherwise but that was my current reasoning.

aolfat: i'm really 50/50 on it right now the more I think about it. Because everything is stored as a url.

simba-git: If this code would be exactly the same for each of them (which seems to be the case) then it should be a utility as you have here. My only nitpick on it is that the function name should be capitalized assuming its also meant to be used outside of this package.

provider/k8s.go
```diff
@@ -1048,6 +1049,15 @@ func (k8s *K8sOfflineStore) RegisterPrimaryFromSourceTable(id ResourceID, source
 }
 
 func blobRegisterPrimary(id ResourceID, sourceName string, logger *zap.SugaredLogger, store FileStore) (PrimaryTable, error) {
+	sourceUrlExists, err := doesSourceUrlExist(store, sourceName)
```
simba-git, line 1052: This is amazing, can this be done upstream somewhere so it catches everything that uses a file store? Maybe in the coordinator?

provider/k8s_test.go
aolfat: good point

provider/filestore_test.go
aolfat: so will this work in ci/cd?

provider/k8s.go
aolfat: It can I think. Let me do think about how iI would do that in a separate PR

PR #978, opened 2023-09-01
Filepath Refactor

epps:
# Description

This PR has a twofold purpose:

* Enhance the `Filepath` interface and extend its use to all relevant areas of the code base (e.g. file store providers, such as Azure Blob Storage)
* Integrate Spark into the offline tests to validate the changes

### Select type(s) of change

- [ ] Documentation update

# Checklist:


filestore/filepath.go
```diff
@@ -0,0 +1,399 @@
...
+func (s3 *S3Filepath) Validate() error {
+	if s3.scheme != "s3://" && s3.scheme != "s3a://" {
+		return fmt.Errorf("invalid scheme '%s', must be 's3:// or 's3a://'", s3.scheme)
+	}
+	if s3.bucket == "" {
+		return fmt.Errorf("bucket cannot be empty")
+	} else {
+		s3.bucket = strings.Trim(s3.bucket, "/")
+	}
+	if s3.key == "" {
```
ahmadnazeri, line 283: would we have cases where `key == "/"`?

filestore/filepath.go
```diff
@@ -0,0 +1,399 @@
...
+	u, err := url.Parse(fullPath)
+	if err != nil {
+		return fmt.Errorf("could not parse full path '%s': %v", fullPath, err)
+	}
+	// Our scheme is the protocol + "://", so we need to suffix the scheme with "://"
+	// to ensure the comparison works.
+	err = azure.FilePath.checkSchemes(fmt.Sprintf("%s://", u.Scheme))
+	if err != nil {
+		return err
+	} else {
```
ahmadnazeri, line 318: Is the `else` required? the next line, you do the same thing

filestore/filepath.go
```diff
@@ -0,0 +1,399 @@
...
+		azure.FilePath.scheme = u.Scheme
+	}
+	azure.FilePath.scheme = u.Scheme
+	azure.FilePath.bucket = u.User.String()              // The container will be in the User field due to the format <scheme>://<container>@<storage_account>
+	azure.StorageAccount = strings.Split(u.Host, ".")[0] // The host will be in the format <storage_account>.dfs.core.windows.net
+	azure.FilePath.key = strings.TrimPrefix(u.Path, "/")
+	azure.FilePath.isDir = false
```
ahmadnazeri, line 325: Is there a reason why this is false? can you pass a folder as key?

metadata/dashboard/dashboard_metadata_test.go
```diff
@@ -126,35 +125,35 @@ func MockGetSourceGet(c *gin.Context, params gin.Params, u url.Values) {
...
-	var data SourceDataResponse
-	rowValues := []string{"row value", "row value"}
-	expectedRows := [][]string{rowValues}
-
-	json.Unmarshal(mockRecorder.Body.Bytes(), &data)
-	assert.Equal(t, http.StatusOK, mockRecorder.Code)
-	assert.Equal(t, iterator.Columns(), data.Columns)
-	assert.Equal(t, expectedRows, data.Rows)
-}
+//func TestGetSourceDataReturnsData(t *testing.T) {
```
ahmadnazeri, line 128: Any reason why this is commented?

provider/k8s.go
```diff
@@ -913,126 +709,312 @@ func parquetIteratorFromBytes(b []byte) (Iterator, error) {
...
+	for {
+		val, err := iter.Next()
+		if err != nil {
+			return nil, fmt.Errorf("could not iterate over file: %w", err)
+		}
+		if val == nil {
+			break
+		}
+		record := GenericRecord{}
+		for _, col := range tbl.schema.Columns {
```
sdreyer, line 939: Should split the nested loops apart into different functions

provider/k8s.go
```diff
@@ -913,126 +709,312 @@ func parquetIteratorFromBytes(b []byte) (Iterator, error) {
...
 func (tbl *FileStorePrimaryTable) NumRows() (int64, error) {
-	return tbl.store.NumRows(tbl.source.Path())
+	src, err := tbl.GetSource()
+	if err != nil {
+		return 0, err
+	}
+	return tbl.store.NumRows(src)
+}
+
+func (tbl *FileStorePrimaryTable) GetSource() (filestore.Filepath, error) {
```
sdreyer, line 1011: Does it make sense to return a reference to the actual table? Then you can do
```
source = tbl.GetSource()
source.path()
```

sdreyer: Feel like itd be more clear

provider/k8s.go
```diff
@@ -1642,10 +1675,14 @@ func (k8s *K8sOfflineStore) materialization(id ResourceID, isUpdate bool) (Mater
...
-	materializationNewestFile, err := k8s.store.NewestFileOfType(destinationPath, Parquet)
+	destinationPath, err := k8s.store.CreateFilePath(featureResourcePath)
+	if err != nil {
+		k8s.logger.Errorw("Could not create file path", "error", err, "featureResourcePath", featureResourcePath)
+		return nil, err
+	}
+	materializationNewestFile, err := k8s.store.NewestFileOfType(destinationPath, filestore.Parquet)
 	k8s.logger.Debugw("Running Materialization", "id", id, "destinationPath", destinationPath, "materializationNewestFile", materializationNewestFile)
-	materializationExists := materializationNewestFile != ""
+	materializationExists := materializationNewestFile.Key() != ""
```
sdreyer, line 1685: Hate this logic. Replace with a `store.exists()` call

provider/k8s.go
```diff
@@ -1728,15 +1768,18 @@ func (k8s *K8sOfflineStore) trainingSet(def TrainingSetDef, isUpdate bool) error
 	}
 	sourcePaths := make([]string, 0)
 	featureSchemas := make([]ResourceSchema, 0)
-	destinationPath := k8s.store.PathWithPrefix(fileStoreResourcePath(def.ID), false)
-	trainingSetExactPath, err := k8s.store.NewestFileOfType(destinationPath, Parquet)
+	destinationPath, err := k8s.store.CreateFilePath(fileStoreResourcePath(def.ID))
+	if err != nil {
+		return fmt.Errorf("could not create file path: %w", err)
+	}
+	trainingSetExactPath, err := k8s.store.NewestFileOfType(destinationPath, filestore.Parquet)
```
sdreyer, line 1775: This isnt used anywhere except to check if the training set exists. replace with an exists call

provider/k8s.go
```diff
@@ -1799,26 +1850,35 @@ func fileStoreGetTrainingSet(id ResourceID, store FileStore, logger *zap.Sugared
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if training set exists: %w", err)
+	}
+	trainingSetExactPath, err := store.NewestFileOfType(filepath, filestore.Parquet)
 	if err != nil {
 		return nil, fmt.Errorf("could not get training set: %v", err)
 	}
-	if trainingSetExactPath == "" {
-		return nil, fmt.Errorf("the training set (%v at resource prefix: %s) does not exist", id, resourceKeyPrefix)
+	if trainingSetExactPath.Key() == "" {
```
sdreyer, line 1869: Don't use this logic to check if exists. Use an exists() call

provider/spark.go
```diff
@@ -1230,44 +1278,59 @@ func (spark *SparkOfflineStore) sqlTransformation(config TransformationConfig, i
...
+	}
+
+	transformationFilepath, err := spark.Store.NewestFileOfType(transformationDirPath, filestore.Parquet)
 	if err != nil {
 		return fmt.Errorf("error checking if transformation file exists")
 	}
-	spark.Logger.Infow("Transformation file", "dest", transformationFile)
-	transformationExists := transformationFile != ""
+	spark.Logger.Infow("Transformation file", "dest", transformationFilepath.PathWithBucket())
+	transformationExists := transformationFilepath.Key() != ""
```
sdreyer, line 1319: bad logic, just check if the file actually exists

provider/spark.go
```diff
@@ -1230,44 +1278,59 @@ func (spark *SparkOfflineStore) sqlTransformation(config TransformationConfig, i
...
+		return fmt.Errorf("transformation %v doesn't exist at %s and you are trying to update", config.TargetTableID, transformationDestination.PathWithBucket())
 	}
 
-	transformationFilePath := spark.Store.PathWithPrefix(GetTransformationFileLocation(config.TargetTableID), false)
-	spark.Logger.Infow("Transformation file path", "dest", transformationFilePath)
-	fileName := "transformation.pkl"
-	transformationFileLocation := fmt.Sprintf("%s/%s", transformationFilePath, fileName)
-	spark.Logger.Infow("Uploading Transformation File", "location", transformationFileLocation)
-	if err := spark.Store.Write(transformationFileLocation, config.Code); err != nil {
+	pklFilepath, err := spark.Store.CreateFilePath(fmt.Sprintf("featureform/DFTransformations/%s/%s/transformation.pkl", config.TargetTableID.Name, config.TargetTableID.Variant))
```
sdreyer, line 1328: this should be in a function somewhere

provider/spark.go
```diff
@@ -1343,42 +1408,62 @@ func (spark *SparkOfflineStore) getSourcePath(path string) (string, error) {
...
-		filePath = spark.Store.PathWithPrefix(transformationPath[:strings.LastIndex(transformationPath, "/")], true)
-		return filePath, nil
+		return transformationPath.PathWithBucket(), nil
 	} else {
 		return filePath, fmt.Errorf("could not find path for %s; fileType: %s, fileName: %s, fileVariant: %s", path, fileType, fileName, fileVariant)
 	}
 }
 
+// TODO: delete this method in favor of Filepath methods
 func (spark *SparkOfflineStore) getResourceInformationFromFilePath(path string) (string, string, string) {
```
sdreyer, line 1440: pls get rid of this

provider/spark.go
```diff
@@ -1513,12 +1677,16 @@ func blobSparkMaterialization(id ResourceID, spark *SparkOfflineStore, isUpdate
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create file path due to error %w (store type: %s; path: %s)", err, spark.Store.FilestoreType(), materializationID.ToFilestorePath())
+	}
+	// TODO: deprecate NewestFileOfType and use List instead
+	materializationNewestFile, err := spark.Store.NewestFileOfType(destinationPath, filestore.Parquet)
 	if err != nil {
 		return nil, fmt.Errorf("could not get newest materialization file: %v", err)
 	}
-	materializationExists := materializationNewestFile != ""
+	materializationExists := materializationNewestFile.Key() != ""
```
sdreyer, line 1689: bad logic, use exists()

filestore/filepath.go
epps: We shouldn't but, I could add that to the conditional check to be sure.

filestore/filepath.go
epps: The idea here is that you're creating a `FilePath` instance that points to a file; otherwise, you'd use `CreateDirPath`.

metadata/dashboard/dashboard_metadata_test.go
epps: No clue. Might've been a bad rebase.

filestore/filepath.go
```diff
@@ -0,0 +1,399 @@
...
+
+func GetFileType(file string) FileType {
+	// check to see if its any of the constants
+	for _, fileType := range []FileType{Parquet, CSV, DB} {
+		if fileType.Matches(file) {
+			return fileType
+		}
+	}
+	// defaults to parquet
+	return Parquet
```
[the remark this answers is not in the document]
epps: Actually, this isn't used, so I'm going to remove it.

provider/filestore.go
```diff
@@ -658,20 +638,313 @@ func (fs *HDFSFileStore) NumRows(key string) (int64, error) {
...
+	)
+	return err
+}
+
+func (store *genericFileStore) Read(path filestore.Filepath) ([]byte, error) {
+	data, err := store.bucket.ReadAll(context.TODO(), path.Key())
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("Read (%d) bytes of object with key (%s)\n", len(data), path.Key())
```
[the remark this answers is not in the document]
epps: I'm not sure I follow you here: are you suggesting I add a logger to the `genericFileStore` struct and use that?

provider/filestore.go
```diff
@@ -658,20 +638,313 @@ func (fs *HDFSFileStore) NumRows(key string) (int64, error) {
...
+
+func (store *genericFileStore) Serve(path filestore.Filepath) (Iterator, error) {
+	if path.IsDir() {
+		return store.ServeDirectory(path)
+	} else {
+		return store.ServeFile(path)
+	}
+}
+
+func (store *genericFileStore) NumRows(path filestore.Filepath) (int64, error) {
```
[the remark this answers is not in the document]
epps: This is used in 2 locations:
* `FileStorePrimaryTable.NumRows`
* `FileStoreMaterialization.NumRows`

In both cases, it'll throw an exception if the file extension isn't CSV.

Am I correct in assuming that in the case of materialization, we wouldn't want/expect the extension to be CSV, but we of course want to support this in the case of a primary table?

provider/k8s.go
epps: It's actually used a few lines down to serve the training set, but I'm converting the key comparison to a call to `Validate`.

backup/save/main.go
```diff
@@ -35,32 +35,32 @@ func main() {
 		panic(err)
 	}
 
-	p := pc.FileStoreType(help.GetEnv("CLOUD_PROVIDER", string(pc.FileSystem)))
+	p := filestore.FileStoreType(help.GetEnv("CLOUD_PROVIDER", string(filestore.FileSystem)))
 
 	var backupProvider backup.Provider
 	switch p {
-	case pc.Azure:
+	case filestore.Azure:
```
simba-git, line 42: Should we consolidate both of these logics into a package of some kind

simba-git: Also is the error handling good on these in case someone forgets to set one?

coordinator/Dockerfile.old
```diff
@@ -18,6 +18,7 @@ ENV PATH /go/bin:$PATH
 RUN protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative ./metadata/proto/metadata.proto
 
 # Copying source files
+COPY ./filestore/ ./filestore/
```
simba-git, line 21: Why is this called Dockerfile.old

filestore/filepath.go
```diff
@@ -0,0 +1,520 @@
...
+	S3         FileStoreType = "S3"
+	GCS        FileStoreType = "GCS"
+	HDFS       FileStoreType = "HDFS"
+)
+
+const (
+	Parquet FileType = "parquet"
+	CSV     FileType = "csv"
+	DB      FileType = "db"
+)
```
simba-git, line 34: right underneath here I'd do:
var AllFileTypes []string = []string{Parquet, CSV, DB}

filestore/filepath.go
```diff
@@ -0,0 +1,520 @@
...
+	GSPrefix, S3Prefix, S3APrefix, AzureBlobPrefix, HDFSPrefix, FileSystemPrefix,
+}
+
+func (ft FileType) Matches(file string) bool {
+	ext := GetFileExtension(file)
+	return FileType(ext) == ft
+}
+
+func IsValidFileType(file string) bool {
+	for _, fileType := range []FileType{Parquet, CSV, DB} {
```
simba-git, line 55: Use AllFileTypes from comment above, these sort of things tend to get out of sync

filestore/filepath.go
```diff
@@ -0,0 +1,520 @@
...
+
+func (hdfs *HDFSFilepath) Validate() error {
+	return fmt.Errorf("not implemented")
+}
+
+type LocalFilepath struct {
+	FilePath
+}
+
+type FilePathGroupingType string
```
simba-git, line 435: Can we comment what this is?

filestore/filepath.go
```diff
@@ -0,0 +1,520 @@
...
+
+const (
+	DateTimeDirectoryGrouping FilePathGroupingType = "DATETIME_DIRECTORY"
+)
+
+type FilePathGroup struct {
```
simba-git, line 441: Same, not clear what this is just via its name

filestore/filepath.go
simba-git: Don't love this abstraction, but need to understand what this does to have a better proposal

provider/blob_online.go
```diff
@@ -69,16 +78,28 @@ func (store OnlineFileStore) readTableValue(feature, variant string) (ValueType,
...
-	if err := store.Delete(tableKey); err != nil {
+	filepath := filestore.AzureFilepath{}
+	if err := filepath.ParseFilePath(tableKey); err != nil {
+		return err
+	}
+	if err := store.Delete(&filepath); err != nil {
 		return fmt.Errorf("could not delete table index key %s: %v", tableKey, err)
 	}
-	if err := store.DeleteAll(entityDirectory); err != nil {
+	filepath = filestore.AzureFilepath{}
```
simba-git, line 98: Typically I would have a filestore.ParseAzureFilePath top level function for each of these rather than instantiating in this way then calling Parse

provider/filestore.go
```diff
@@ -26,62 +31,32 @@ import (
...
+	Delete(key filestore.Filepath) error
+	DeleteAll(dir filestore.Filepath) error
+	NewestFileOfType(prefix filestore.Filepath, fileType filestore.FileType) (filestore.Filepath, error)
+	List(dirPath filestore.Filepath, fileType filestore.FileType) ([]filestore.Filepath, error)
+	NumRows(key filestore.Filepath) (int64, error)
+	Close() error
+	Upload(sourcePath filestore.Filepath, destPath filestore.Filepath) error
+	Download(sourcePath filestore.Filepath, destPath filestore.Filepath) error
+	FilestoreType() filestore.FileStoreType
+	AddEnvVars(envVars map[string]string) map[string]string
```
simba-git, line 48: This method needs a comment, kind of strange to have this in FileStore

provider/filestore.go
```diff
@@ -26,62 +31,32 @@ import (
...
+	// CreateFilePath creates a new filepath object with the bucket and scheme from a Key
+	CreateFilePath(key string) (filestore.Filepath, error)
+	CreateDirPath(key string) (filestore.Filepath, error)
```
simba-git, line 51: Can you even create directories? You can't in S3 for example When would we create directories?

provider/spark.go
```diff
@@ -58,7 +59,7 @@ type SparkFileStore interface {
 	SparkConfig() []string
 	CredentialsConfig() []string
 	Packages() []string
-	Type() string
+	Type() string // s3, azure_blob_store, google_cloud_storage, hdfs, local
```
simba-git, line 62: should this be a type variable? Or use one of the filestore consts?

provider/spark.go
```diff
@@ -1461,42 +1571,118 @@ func (d *DatabricksExecutor) GetDFArgs(outputURI string, code string, sources []
...
-	return &FileStorePrimaryTable{spark.Store, filePath, true, id}, nil
+	spark.Logger.Debugw("Retrieved transformation source", "id", id, "filePath", transformationPath.ToURI())
+	return &FileStorePrimaryTable{spark.Store, transformationPath, TableSchema{}, true, id}, nil
 }
 
 func (spark *SparkOfflineStore) UpdateTransformation(config TransformationConfig) error {
 	return spark.transformation(config, true)
 }
 
+// TODO: add a comment akin to the one explaining the logic for CreateResourceTable
```
simba-git, line 1586: Think you did the TODO yeah?

runner/copy_test.go
```diff
@@ -791,6 +791,10 @@ func (m MockOfflineTable) Write(provider.ResourceRecord) error {
 	return nil
 }
 
+func (m MockOfflineTable) WriteBatch([]provider.ResourceRecord) error {
+	return nil
```
simba-git, line 795: Have we just continued to not implement thing and return nil? Does this thing get used?

provider/spark.go
```diff
@@ -362,22 +345,23 @@ type DatabricksExecutor struct {
 	errorMessageClient *dbClient.DatabricksClient
 }
 
-func (e *EMRExecutor) PythonFileURI(store SparkFileStore) string {
-	return ""
+func (e *EMRExecutor) PythonFileURI(store SparkFileStore) (filestore.Filepath, error) {
```
simba-git, line 348: What is this doing?

provider/spark.go
```diff
@@ -541,32 +546,67 @@ type defaultPythonOfflineQueries struct{}
...
+	return fmt.Sprintf("`%s__%s__%s`", id.Type, id.Name, id.Variant)
 }
 
 func (q defaultPythonOfflineQueries) trainingSetCreate(def TrainingSetDef, featureSchemas []ResourceSchema, labelSchema ResourceSchema) string {
 	columns := make([]string, 0)
 	joinQueries := make([]string, 0)
 	feature_timestamps := make([]string, 0)
 	for i, feature := range def.Features {
-		featureColumnName := featureColumnName(feature)
+		fmt.Printf("%d feature: %v\n", i, featureSchemas[i].Entity)
```
simba-git, line 603: Did you mean to leave this here?

provider/spark.go
```diff
@@ -800,13 +853,13 @@ func (s *SparkGenericExecutor) RunSparkJob(args []string, store SparkFileStore)
 	return nil
 }
 
-func (s *SparkGenericExecutor) PythonFileURI(store SparkFileStore) string {
+func (s *SparkGenericExecutor) PythonFileURI(store SparkFileStore) (filestore.Filepath, error) {
 	// not used for Spark Generic Executor
-	return ""
+	return nil, nil
```
simba-git, line 858: If this is only used by one executor, why does the interface have it at all?

provider/spark.go
```diff
@@ -1343,46 +1412,84 @@ func (spark *SparkOfflineStore) getSourcePath(path string) (string, error) {
...
+		if err != nil {
+			spark.Logger.Errorf("could not construct ResourceID for Azure Blob Storage path %s due to %v", path, err)
 			return "", "", ""
 		}
-		fileType, fileName, fileVariant = strings.ToLower(filePaths[2]), filePaths[3], filePaths[4]
-	} else if strings.HasPrefix(path, s3aPrefix) {
-		filePaths := strings.Split(path[len(s3aPrefix):], "/")
+		fileType, fileName, fileVariant = strings.ToLower(id.Type.String()), id.Name, id.Variant
+	} else if strings.HasPrefix(path, filestore.S3Prefix) {
+		filePaths := strings.Split(path[len(filestore.S3Prefix):], "/")
```
simba-git, line 1466: All this logic should be encapsulated somewhere too imo

provider/spark.go
simba-git: Agreed we're still creating paths with Sprintf

backup/providers.go
```diff
@@ -6,16 +6,22 @@ import (
...
-	Download(src, dest string) error
-	LatestBackupName(prefix string) (string, error)
+	// **NOTE:** Upload accepts strings instead of filestore.Filepath because
+	// the store, which could create a filestore-specific filepath, is not exposed
+	// by the Provider interface, and it seems unnecessary to expose it at this time.
+	// In contrast, Download accepts a filestore.Filepath because the LatestBackupName
+	// returns a filestore.Filepath and dest is an instance of filestore.LocalFilepath.
+	Upload(src, dest string) error
+	Download(src, dest filestore.Filepath) error
+	LatestBackupName(filenamePrefix string) (filestore.Filepath, error)
```
aolfat, line 24: I know there wasn't a comment before but it might be useful to know what this is

provider/iterators.go
```diff
@@ -0,0 +1,451 @@
...
+
+	"github.com/parquet-go/parquet-go"
+
+	_ "gocloud.dev/blob/fileblob"
+	_ "gocloud.dev/blob/memblob"
+
+	filestore "github.com/featureform/filestore"
+)
+
+// PARQUET
```
aolfat, line 22: nit: I don't think we need this comment

provider/k8s.go
```diff
@@ -913,126 +709,312 @@ func parquetIteratorFromBytes(b []byte) (Iterator, error) {
...
 	schema ResourceSchema
 }
 
+// TODO: implement
 func (tbl *BlobOfflineTable) Write(ResourceRecord) error {
 	return fmt.Errorf("not yet implemented")
 }
 
+func (tbl *BlobOfflineTable) WriteBatch(records []ResourceRecord) error {
+	destination, err := filestore.NewEmptyFilepath(tbl.store.FilestoreType())
```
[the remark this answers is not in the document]
aolfat: i fully agree with this. They're tied together.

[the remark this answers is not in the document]
aolfat: like you should be able to do filestore.ParseFilePath()

provider/filestore.go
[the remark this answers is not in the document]
aolfat: yeahhh you could just do that ^

filestore/filepath.go
```diff
@@ -0,0 +1,520 @@
...
+
+	// Returns the full path to the object, including the scheme and bucket/container
+	ToURI() string
+	// Consumes a URI (e.g. abfss://<container>@<storage_account>/path/to/file) and parses it into
+	// the specific parts that the implementation expects.
+	ParseFilePath(path string) error
+	ParseDirPath(path string) error
+
+	Validate() error
+	IsValid() bool
```
aolfat, line 105: Why are we storing whether it's valid or not? I imagine we shouldn't even successfully create a file path if it's not valid?

aolfat: like why should we store a filepath that's not valid?

filestore/filepath.go
```diff
@@ -0,0 +1,544 @@
...
+		return nil, fmt.Errorf("unknown store type '%s'", storeType)
+	}
+}
+
+type FilePath struct {
+	scheme  string
+	bucket  string
+	key     string
+	isDir   bool
+	isValid bool
```
aolfat, line 156: not convinced we need this

aolfat: something feels off and this design and I don't have a good solution yet. All I know is that it's a bit confusing.

correct me if i'm wrong but the way it will always work is
- you always create a filepath like this -> filestore.NewEmptyFilepath(store.FilestoreType()).
- and you do this so that you can eventually parse some path with the fileStore's schema

so I'm thinking if this is the use case, then the way things should be done is, store.ParseFilePathFromUrl() or store.ParseFilePathFromKey().

ahmadnazeri (approved):

PR #993, opened 2023-09-25
Bugfix: Returning meaningful error message when Dill doesn't work

ahmadnazeri:
Adding a check to make sure that we return a more meaningful error message when the dill throws:
```
TypeError: code() takes at most 16 arguments (19 given)
```

I was also reading through more about it and came across this Issue on [dill's GitHub repo](https://github.com/uqfoundation/dill/issues/254). This [comment](https://github.com/uqfoundation/dill/issues/254#issuecomment-363625245) on the issue can potentially be really useful for us. Basically, they mention that **marshal** might be an alternative option that would avoid these issues. 

# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/provider/scripts/spark/offline_store_spark_runner.py
+++ b/provider/scripts/spark/offline_store_spark_runner.py
@@ -199,6 +199,7 @@ def get_code_from_file(file_path, store_type=None, credentials=None):
     print(f"Retrieving transformation code from {file_path} in {store_type}.")
 
     code = None
+    transformation_pkl = None
     if store_type == "s3":
         # S3 paths are the following path: 's3a://{bucket}/key/to/file'.
         # the split below separates the bucket name and the key that is
@@ -226,7 +227,7 @@ def get_code_from_file(file_path, store_type=None, credentials=None):
             s3_object.download_fileobj(f)
 
             f.seek(0)
-            code = dill.loads(f.read())
+            transformation_pkl = f.read()
 
     elif store_type == "hdfs":
         # S3 paths are the following path: 's3://{bucket}/key/to/file'.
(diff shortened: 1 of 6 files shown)


provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -362,6 +363,15 @@ def delete_file(file_path):
         print(f"{file_path} does not exist.")
 
 
+def dill_exception(exception):
+    if "TypeError: code() takes at most" in str(exception):
```
aolfat, line 367: This feels like we're just putting another bandaid on the issue.

Are we confident that the only thing that can produce this is a dill issue?

The best way to solve this is to somehow catch if there are different python versions. Maybe we expose an API that outputs it.

But if we ARE going to bandaid this up we should at least change the error message to be something more like: "This error is most likley caused by ..., check to see if..."

simba-git: I would at the very least verify this error is happening when we are using Dill to atleast drop the chances of ti being something else

Also rename function to is_dill_exception

ahmadnazeri: So this PR is focused on just showing a better error message when users are faced with this known [Dill issue](https://github.com/uqfoundation/dill/issues/254). 

How does this error message sound? 
```
This error is most likely caused by different Python versions between the client and Spark provider. Check to see if you are running Python version '3.10.2' on the client. 
```

aolfat: yeah I like that message better

aolfat: ideally it should be impossible to even get into this state

sdreyer (approved):
lgtm

PR #1036, opened 2023-09-29
Bugfix: etcd tests

aolfat:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/metadata/etcd_test.go
+++ b/metadata/etcd_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	"log"
 	"reflect"
 	"testing"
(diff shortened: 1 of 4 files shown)


metadata/etcd_test.go
```diff
@@ -689,46 +690,76 @@ func TestEtcdConfig_Put(t *testing.T) {
...
+	if err == nil {
+		t.Errorf("Get() should have failed with invalid key")
+	}
+	// check to see if it's an etcderror
+	if _, ok := err.(*KeyNotFoundError); !ok {
+		t.Errorf("Get() should have failed with KeyNotFoundError")
+	}
+}
+
+func TestEtcdConfig_ValidKey(t *testing.T) {
```
ahmadnazeri, line 744: LGTM but we can change the tests into test cases instead of separate tests. For now, we are good.

ahmadnazeri (approved):

metadata/etcd_test.go
aolfat: why split it up?

aolfat: why do test cases?

aolfat: I actually tried to it at first. They're slightly different enough where it actually makes the one bigger test harder to read.

aolfat: let me try one more time

aolfat: alright i think i made it better

sdreyer (approved):

PR #1041, opened 2023-09-30
Bugfix: Same variants being added to parent resource, no deduping

aolfat:
# Description
https://github.com/featureform/featureform-enterprise/issues/221

Issue has to do with how we update. When you apply a resource twice it attempts to update it, after the update (which only updates tags and properties btw) it goes to propagate the change and part of the propagate goes to update the parent resource (a sourcevariant resource goes to update the source resource) and part of the propagation has a notify that adds the source variant to the source's variant list. Since the variant list is a list (and not a set) it allows the same variant to be added twice. All I've added is a check to ensure it doesn't exist

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/metadata/dashboard/dashboard_metadata.go
+++ b/metadata/dashboard/dashboard_metadata.go
@@ -1,5 +1,4 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
-// This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
--- a/metadata/dashboard/dashboard_metadata.go
+++ b/metadata/dashboard/dashboard_metadata.go
@@ -1,4 +1,5 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
+// This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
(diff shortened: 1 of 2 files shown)


metadata/dashboard/dashboard_metadata.go
```diff
@@ -1,4 +1,5 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
+// This Source Code Form is subject to the terms of the Mozilla Public
```
sdreyer, line 2: duplicate

metadata/metadata.go
```diff
@@ -770,6 +781,9 @@ func (this *trainingSetResource) Notify(lookup ResourceLookup, op operation, tha
 	if !isVariant {
 		return nil
 	}
+	if slices.Contains(this.serialized.Variants, otherId.Variant) {
```
sdreyer, line 784: Whats the purpose of this PR?

metadata/dashboard/dashboard_metadata.go
aolfat: whoops

metadata/metadata.go
```diff
@@ -1533,11 +1547,15 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	}
 	parentId, hasParent := id.Parent()
 	if hasParent {
-		if parentExists, err := serv.lookup.Has(parentId); err != nil {
+		parentExists, err := serv.lookup.Has(parentId)
```
aolfat, line 1550: this was just so hard to read so we make it easier

metadata/dashboard/dashboard_metadata.go
aolfat, line 2:
```suggestion
```

metadata/metadata.go
```diff
@@ -641,6 +649,9 @@ func (this *labelResource) Notify(lookup ResourceLookup, op operation, that Reso
 	if !isVariant {
 		return nil
 	}
+	if slices.Contains(this.serialized.Variants, otherId.Variant) {
```
ahmadnazeri, line 652: should we add a debug log here?

ahmadnazeri (approved):

PR #1045, opened 2023-10-01
Databricks End to end tests

sdreyer:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/status_display.py
+++ b/client/src/featureform/status_display.py
@@ -156,6 +156,11 @@ class StatusDisplayer:
                 live.refresh()
 
                 if finished_running:
+                    # This block is used for testing
+                    # Tests check for both stderr and an exception
+                    # If we don't throw an exception, then tests will pass even when things fail to register
+                    # We also print all the error messages because the table does not get saved when
+                    # capturing stdout/stderr
                     if self.did_error:
                         statuses = ""
                         for _, status in self.resource_to_status_list:
--- a/client/src/featureform/status_display.py
+++ b/client/src/featureform/status_display.py
@@ -25,8 +25,8 @@ MAX_NUM_RUNNING_DOTS = 10
 SECONDS_BETWEEN_STATUS_CHECKS = 2
 
 
-def display_statuses(stub: ApiStub, resources: List[Resource]):
-    StatusDisplayer(stub, resources).display()
+def display_statuses(stub: ApiStub, resources: List[Resource], verbose=False):
+    StatusDisplayer(stub, resources, verbose=verbose).display()
 
 
 @dataclass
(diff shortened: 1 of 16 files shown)


client/src/featureform/status_display.py
```diff
@@ -151,6 +155,9 @@ def display(self):
                 live.refresh()
 
                 if finished_running:
+                    if self.did_error:
+                        sys.tracebacklimit = 0
+                        raise Exception("Some resources failed to create")
```
aolfat, line 160: what is this for?

sdreyer: Right now we just break when things are done, rather than exit(1). However, if you exit(1) it kills the parent process aka the tests. This format gives a failed exit without a massive stack trace

aolfat: so this is purely for tests?

provider/filestore.go
```diff
@@ -244,9 +244,9 @@ func NewS3FileStore(config Config) (FileStore, error) {
 		return nil, fmt.Errorf("could not deserialize s3 store config: %v", err)
 	}
 
-	s3StoreConfig.BucketPath = strings.TrimPrefix("s3://", strings.TrimPrefix(s3StoreConfig.BucketPath, "s3a://"))
+	trimmedBucket := strings.TrimPrefix(strings.TrimPrefix(s3StoreConfig.BucketPath, "s3a://"), "s3://")
```
aolfat, line 247: ooooof that's a bad error lol. We should have had a test for this

aolfat: also I would've maybe split this into two lines, it took me a sec to figure out what was going on

sdreyer: fair

sdreyer: This isn't really where I want it, but theres 2 other tasks to fix this for other filestores where we can put it in a better place

provider/spark.go
```diff
@@ -1722,31 +1722,40 @@ func blobSparkMaterialization(id ResourceID, spark *SparkOfflineStore, isUpdate
...
-		return nil, fmt.Errorf("could not get datetime directory grouping for source files: %v", err)
-	}
-	newest, err := groups.GetFirst()
-	if err != nil {
-		return nil, fmt.Errorf("could not get newest source file: %v", err)
-	}
-	sourceUris := make([]string, len(newest))
-	for i, sourceFile := range newest {
-		sourceUris[i] = sourceFile.ToURI()
+	var sourceURIs []string
```
aolfat, line 1725: yah this whole thing should be it's own method. GetSourceUrisFromPath or something

I'm not familiar with a lot of these functions, like NewFilePathGroup or even Groups. trusting ya

client/src/featureform/status_display.py
aolfat: if it's just for tests I would leave a comment or something.

sdreyer: and any usage where apply would be synchronous and fail. Otherwise it fails silently

sdreyer: like in .dataframe

aolfat (approved):

client/src/featureform/status_display.py
aolfat: I'd still add a comment on what's going on. I know I'll look at this in a few months about be like wtf is going on

PR #1054, opened 2023-10-01
Fix Broken SQL Check

sdreyer:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -1100,9 +1100,9 @@ class SQLTransformationDecorator:
     def __set_query(self, query: str):
         if query == "":
             raise ValueError("Query cannot be an empty string")
-        # The regex here is wrong, needs to be fixed
-        # if not self._is_valid_sql_query(query):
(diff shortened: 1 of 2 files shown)


client/src/featureform/register.py
```diff
@@ -1100,8 +1100,9 @@ def __call__(self, fn: Callable[[], str]):
     def __set_query(self, query: str):
         if query == "":
             raise ValueError("Query cannot be an empty string")
-        if not self._is_valid_sql_query(query):
-            raise InvalidSQLQuery(query)
+        # The regex here is wrong, needs to be fixed
+        # if not self._is_valid_sql_query(query):
```
aolfat, line 1104: Was there no test for this? If there was, add the case where you found it to be broken into the test.

aolfat (approved):

client/src/featureform/register.py
crfeliz: I updated the regex and tests. Please re-review.

PR #1079, opened 2023-10-12
Materialization Copy Performance Improvement

epps:
# Description

Based on profiling conducted using `pprof`, we determined that the materialization's copy operation spent far too much time waiting on calls to `Table.Set`; introducing concurrency resulted in a ~72% decrease in the time required to copy ~1M records to the inference store.

### Select type(s) of change

- [ ] Documentation update
- [ ] Performance enhancement

# Checklist:

### DIFF
--- a/runner/copy.go
+++ b/runner/copy.go
@@ -75,13 +75,14 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
 			jobWatcher.EndWatch(fmt.Errorf("failed to create iterator: %w", err))
 			return
 		}
-		// Buffer the records channel to ensure readers won't block on the writer loop below
-		ch := make(chan provider.ResourceRecord, 100_000)
+		// Buffer the records channel to ensure readers won't block the writer loop below
+		ch := make(chan provider.ResourceRecord, 1_000_000)
 		var wg sync.WaitGroup
-		wg.Add(100)
+		waitGroupSize := 1000
+		wg.Add(waitGroupSize)
 		// Create a "pool" of 100 goroutines to write to the table to better utilize CPU resources
 		// and speed up the write operations to the inference table.
-		for idx := 0; idx < 100; idx++ {
+		for idx := 0; idx < waitGroupSize; idx++ {
 			go func() {
 				defer wg.Done()
 				for record := range ch {
--- a/runner/copy.go
+++ b/runner/copy.go
@@ -75,17 +75,31 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
 			jobWatcher.EndWatch(fmt.Errorf("failed to create iterator: %w", err))
 			return
 		}
-		i := 0
-		for it.Next() {
-			i += 1
-			value := it.Value().Value
-			entity := it.Value().Entity
-			err := m.Table.Set(entity, value)
-			if err != nil {
-				jobWatcher.EndWatch(fmt.Errorf("could not set table: %w", err))
-				return
-			}
+		// Buffer the records channel to ensure readers won't block on the writer loop below
+		ch := make(chan provider.ResourceRecord, 100_000)
+		var wg sync.WaitGroup
+		wg.Add(100)
+		// Create a "pool" of 100 goroutines to write to the table to better utilize CPU resources
+		// and speed up the write operations to the inference table.
+		for idx := 0; idx < 100; idx++ {
+			go func() {
+				defer wg.Done()
+				for record := range ch {
+					err := m.Table.Set(record.Entity, record.Value)
+					if err != nil {
+						close(ch)
+						jobWatcher.EndWatch(fmt.Errorf("could not set table: %w", err))
+						return
+					}
+				}
+			}()
 		}
+		// Completely consume the iterator and write all records to the channel
+		for it.Next() {
+			ch <- it.Value()
+		}
+		close(ch)
+		wg.Wait()
 		if err = it.Err(); err != nil {
 			jobWatcher.EndWatch(fmt.Errorf("iteration failed with error: %w", err))
 			return
@@ -93,10 +107,12 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
 		err = it.Close()
 		if err != nil {
 			jobWatcher.EndWatch(fmt.Errorf("failed to close iterator: %w", err))
+			return
 		}
 		err = m.Store.Close()
 		if err != nil {
 			jobWatcher.EndWatch(fmt.Errorf("failed to close Online Store: %w", err))
+			return
 		}
 		jobWatcher.EndWatch(nil)
 	}()
--- a/runner/copy.go
+++ b/runner/copy.go
@@ -17,8 +17,15 @@ import (
 	"go.uber.org/zap"
 )
 
+// We buffer the channel responsible for passing records to the goroutines that will write
+// to the inference store avoid blocking the writer loop below; after some initial testing,
+// 1M records seems to be a good buffer size
 const resourceRecordBufferSize = 1_000_000
-const waitGroupSize = 1000
+
+// We create a pool of goroutines per materialization chunk runner to make Set requests to
+// the inference store asynchronously; after some initial testing, 1000 workers appears to
+// offer the best results
+const workerPoolSize = 1000
 
 type IndexRunner interface {
 	types.Runner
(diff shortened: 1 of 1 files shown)


runner/copy.go
```diff
@@ -75,28 +75,44 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
 			jobWatcher.EndWatch(fmt.Errorf("failed to create iterator: %w", err))
 			return
 		}
-		i := 0
+		// Buffer the records channel to ensure readers won't block on the writer loop below
+		ch := make(chan provider.ResourceRecord, 100_000)
+		var wg sync.WaitGroup
+		wg.Add(100)
```
aolfat, line 81: 100 should probz be a var if it's going to be used on line 84 too

runner/copy.go
```diff
@@ -75,28 +75,44 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+					if err != nil {
+						close(ch)
+						jobWatcher.EndWatch(fmt.Errorf("could not set table: %w", err))
+						return
+					}
+				}
+			}()
+		}
+		// Completely consume the iterator and write all records to the channel
 		for it.Next() {
```
aolfat, line 98: i'm no goroutines expert, maybe this is standard, but it seems counterintuitive to start consuming from the channels first and produce to the channels below

epps: The logic is here:
* create a channel that has a buffer of 1M records, which means that sending isn't blocked by receiving
* spawn 1K go routines first so that once you start sending messages into the channel, receive and process them, theoretically keeping the actual buffer usage < 1M records
* Consume the iterator last because it's much faster than the Set operations performed in the go routines.

aolfat: I'm not sure I'm following and I'm asking because I'm trying to understand. This is what the code looks like to me

`
for all goroutines:
    read record from channels

for all records in iterator:
   put record in channel
`

it just feels more intuitive to do it the other way around? Is there some sort of performance benefit we get from doing it this way?

epps: If you send messages to the channel before you had go routines (i.e. workers to pull from it) you would synchronously fill the buffer, and exceed it in the case of files with 1M+ rows, before anything was initialized to process them.

The current pattern allows those go routines (workers) to begin working through the buffer concurrently as it's being filled by the iterator.

aolfat: ok simba just explained it to me

runner/copy.go
```diff
@@ -75,28 +75,45 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+		wg.Add(waitGroupSize)
+		// Create a "pool" of 100 goroutines to write to the table to better utilize CPU resources
+		// and speed up the write operations to the inference table.
+		for idx := 0; idx < waitGroupSize; idx++ {
+			go func() {
+				defer wg.Done()
+				for record := range ch {
+					err := m.Table.Set(record.Entity, record.Value)
+					if err != nil {
+						close(ch)
```
aolfat, line 91: There's a bug in here (that chatgpt helped me find) (and I brought it up to simba too, he said he'll help you with it)

By closing the channel here; if there was an error, it closes all the other channels. Which is fine. BUT, now that the channels are closed, if you try to write to the channel below, you'd get a shitty panic. There's a trick about using an error channel with buffer size of 1 to capture those which allows you to more gracefully handle the error.

runner/copy.go
aolfat: Sorry, I just saw your explanation, that's exactly it!

runner/copy.go
epps: Addressed and tested.

runner/copy.go
```diff
@@ -17,6 +17,9 @@ import (
 	"go.uber.org/zap"
 )
 
+const resourceRecordBufferSize = 1_000_000
+const waitGroupSize = 1000
```
simba-git, line 21: I'd add a comment and make this workerSize

Make it clear that this is the number of go-rountines PER materialization chunk runner to make redis requests

runner/copy.go
```diff
@@ -17,6 +17,9 @@ import (
 	"go.uber.org/zap"
 )
 
+const resourceRecordBufferSize = 1_000_000
```
simba-git, line 20: Needs a comment for clarity on what this is

sdreyer (approved):

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+					if err := m.Table.Set(record.Entity, record.Value); err != nil {
+						select {
+						case errCh <- fmt.Errorf("could not set value to table: %w", err):
+						default:
+						}
+					}
+				}
+			}()
+		}
+	iterationLoop:
```
aolfat, line 104: woah i didn't know you could do this

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
-		i := 0
-		for it.Next() {
-			i += 1
-			value := it.Value().Value
-			entity := it.Value().Entity
-			err := m.Table.Set(entity, value)
-			if err != nil {
-				jobWatcher.EndWatch(fmt.Errorf("could not set table: %w", err))
-				return
+		// Buffer the records channel to ensure readers won't block the writer loop below
```
simba-git, line 81: I'd add a comment above all of this to explain the general algorithm:

One reader thread reads as fast as it can and writers to a buffered channel. workerSize go routines run and each reads from the buffered channel and writes to Redis, waiting for a response.

aolfat (approved):

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+		ch := make(chan provider.ResourceRecord, resourceRecordBufferSize)
+		// Using a buffered error channel to capture any errors that may occur while trying to
+		// set values; buffering the error channel prevents the goroutines from blocking when
+		// trying to write to the channel.
+		errCh := make(chan error, 1)
+		var wg sync.WaitGroup
+		wg.Add(waitGroupSize)
+		// Create a "pool" of 100 goroutines to write to the table to better utilize CPU resources
```
simba-git, line 89: itsa not 100, and I'd frame it as make a set of go-routines so we can wait for Redis to respond asynchronously.

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
-				jobWatcher.EndWatch(fmt.Errorf("could not set table: %w", err))
-				return
+		// Buffer the records channel to ensure readers won't block the writer loop below
+		ch := make(chan provider.ResourceRecord, resourceRecordBufferSize)
+		// Using a buffered error channel to capture any errors that may occur while trying to
+		// set values; buffering the error channel prevents the goroutines from blocking when
+		// trying to write to the channel.
+		errCh := make(chan error, 1)
+		var wg sync.WaitGroup
+		wg.Add(waitGroupSize)
```
simba-git, line 88: wg is created to make sure that all the workers are done before the main go-routine continues

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+		// Create a "pool" of 100 goroutines to write to the table to better utilize CPU resources
+		// and speed up the write operations to the inference table.
+		for idx := 0; idx < waitGroupSize; idx++ {
+			go func() {
+				defer wg.Done()
+				for record := range ch {
+					if err := m.Table.Set(record.Entity, record.Value); err != nil {
+						select {
+						case errCh <- fmt.Errorf("could not set value to table: %w", err):
```
simba-git, line 97: because errCh has a buffer of 1, this case will only happen if no other errors are waiting to be processed. If another go-routine has an error, we will ignore this current error.

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+						default:
+						}
+					}
+				}
+			}()
+		}
+	iterationLoop:
+		for {
```
simba-git, line 105: for it.Next() {
  select {
  case err = <-errCh:
    break
  case ch <- it.Value():
  }
}

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+				// Completely consume the iterator and write all records to the channel
+				if it.Next() {
+					ch <- it.Value()
+				} else {
+					break iterationLoop
+				}
 			}
 		}
+		close(ch)
+		wg.Wait()
```
simba-git, line 121: close errChan after this

runner/copy.go
```diff
@@ -75,28 +78,64 @@ func (m *MaterializedChunkRunner) Run() (types.CompletionWatcher, error) {
...
+		if err, ok := <-errCh; ok {
```
simba-git, line 122: you already read from errChan above

runner/copy.go
epps: This is OK, right? We want to fail on the first error, correct? If even one record fails to materialize, then we should consider the materialization a failure, right?

PR #1085, opened 2023-10-16
Provider Health Checks

epps:
# Description

This PR introduces a `Check` method on the `Provider` interface, with implementations for:

* Redis
* DynamoDB
* Snowflake
* Spark
* PostgreSQL

Additionally, the `health` package is introduced to handle processing health checks by provider name. In the near feature, this package will become a standalone service to facilitate runner provider health checks on a schedule.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/api/main.go
+++ b/api/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/featureform/metadata"
 	pb "github.com/featureform/metadata/proto"
 	srv "github.com/featureform/proto"
+	pt "github.com/featureform/provider/provider_type"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
@@ -590,8 +591,12 @@ func (serv *MetadataServer) CreateProvider(ctx context.Context, provider *pb.Pro
 		serv.Logger.Errorw("Failed to create provider", "error", err)
 		return nil, err
 	}
+	if !serv.health.IsSupportedProvider(pt.Type(provider.Type)) {
+		serv.Logger.Infow("Provider type is currently not supported for health check", "type", provider.Type)
+		return &pb.Empty{}, nil
+	}
 	var status *pb.ResourceStatus
-	isHealthy, err := serv.health.Check(provider.Name)
+	isHealthy, err := serv.health.CheckProvider(provider.Name)
 	if err != nil || !isHealthy {
 		serv.Logger.Errorw("Provider health check failed", "error", err)
 		status = &pb.ResourceStatus{
--- a/api/main.go
+++ b/api/main.go
@@ -21,6 +21,7 @@ import (
 
 	"google.golang.org/grpc/credentials/insecure"
 
+	health "github.com/featureform/health"
 	help "github.com/featureform/helpers"
 	"github.com/featureform/metadata"
 	pb "github.com/featureform/metadata/proto"
@@ -44,6 +45,7 @@ type MetadataServer struct {
 	meta    pb.MetadataClient
 	client  *metadata.Client
 	pb.UnimplementedApiServer
+	health *health.Health
 }
 
 type OnlineServer struct {
@@ -583,7 +585,40 @@ func (serv *MetadataServer) ListProviders(in *pb.Empty, stream pb.Api_ListProvid
 
 func (serv *MetadataServer) CreateProvider(ctx context.Context, provider *pb.Provider) (*pb.Empty, error) {
 	serv.Logger.Infow("Creating Provider", "name", provider.Name)
-	return serv.meta.CreateProvider(ctx, provider)
+	_, err := serv.meta.CreateProvider(ctx, provider)
+	if err != nil {
+		serv.Logger.Errorw("Failed to create provider", "error", err)
+		return nil, err
+	}
+	var status *pb.ResourceStatus
+	isHealthy, err := serv.health.Check(provider.Name)
+	if err != nil || !isHealthy {
+		serv.Logger.Errorw("Provider health check failed", "error", err)
+		status = &pb.ResourceStatus{
+			Status:       pb.ResourceStatus_FAILED,
+			ErrorMessage: err.Error(),
+		}
+	} else {
+		serv.Logger.Infow("Provider health check passed", "name", provider.Name)
+		status = &pb.ResourceStatus{
+			Status: pb.ResourceStatus_READY,
+		}
+	}
+	statusReq := &pb.SetStatusRequest{
+		ResourceId: &pb.ResourceID{
+			Resource: &pb.NameVariant{
+				Name: provider.Name,
+			},
+			ResourceType: pb.ResourceType_PROVIDER,
+		},
+		Status: status,
+	}
+	_, statusErr := serv.meta.SetResourceStatus(ctx, statusReq)
+	if statusErr != nil {
+		serv.Logger.Errorw("Failed to set provider status", "error", statusErr, "health check error", err)
+		return nil, statusErr
+	}
+	return &pb.Empty{}, err
 }
 
 func (serv *MetadataServer) CreateSourceVariant(ctx context.Context, source *pb.SourceVariant) (*pb.Empty, error) {
@@ -762,6 +797,7 @@ func (serv *ApiServer) Serve() error {
 	}
 	serv.metadata.client = client
 	serv.online.client = srv.NewFeatureClient(servConn)
+	serv.metadata.health = health.NewHealth(client)
 	return serv.ServeOnListener(lis)
 }
 
--- a/health/health.go
+++ b/health/health.go
@@ -1,0 +1,45 @@
+package health
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/featureform/metadata"
+	"github.com/featureform/provider"
+	pc "github.com/featureform/provider/provider_config"
+	pt "github.com/featureform/provider/provider_type"
+)
+
+type Health struct {
+	metadata *metadata.Client
+}
+
+func NewHealth(client *metadata.Client) *Health {
+	return &Health{
+		metadata: client,
+	}
+}
+
+func (h *Health) Check(providerName string) (bool, error) {
+	rec, err := h.metadata.GetProvider(context.Background(), providerName)
+	if err != nil {
+		return false, err
+	}
+	if !h.isSupportedProvider(pt.Type(rec.Type())) {
+		return false, fmt.Errorf("unsupported provider type: %s", rec.Type())
+	}
+	p, err := provider.Get(pt.Type(rec.Type()), pc.SerializedConfig(rec.SerializedConfig()))
+	if err != nil {
+		return false, err
+	}
+	return p.Check()
+}
+
+func (h *Health) isSupportedProvider(t pt.Type) bool {
+	switch t {
+	case pt.RedisOnline, pt.DynamoDBOnline, pt.PostgresOffline, pt.SnowflakeOffline, pt.SparkOffline:
+		return true
+	default:
+		return false
+	}
+}
--- a/health/health.go
+++ b/health/health.go
@@ -2,13 +2,18 @@ package health
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/featureform/metadata"
 	"github.com/featureform/provider"
 	pc "github.com/featureform/provider/provider_config"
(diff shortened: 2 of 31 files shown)


health/health.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"fmt"
+
+	"github.com/featureform/metadata"
+	"github.com/featureform/provider"
+	pc "github.com/featureform/provider/provider_config"
+	pt "github.com/featureform/provider/provider_type"
+)
+
+type Health struct {
```
aolfat, line 13: what about HealthChecker?

health/health.go
```diff
@@ -0,0 +1,45 @@
...
+	metadata *metadata.Client
+}
+
+func NewHealth(client *metadata.Client) *Health {
+	return &Health{
+		metadata: client,
+	}
+}
+
+func (h *Health) Check(providerName string) (bool, error) {
```
aolfat, line 23: nit: what do you think about CheckProviderHealth

aolfat: who's going to call this? We have to expose protos for this right?

provider/provider.go
```diff
@@ -45,6 +45,7 @@ type Provider interface {
 	AsOfflineStore() (OfflineStore, error)
 	Type() pt.Type
 	Config() pc.SerializedConfig
+	Check() (bool, error)
```
aolfat, line 48: nit: what do you think about just calling it ping?

health/health_test.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+var providerType = flag.String("provider", "", "provider type under test")
+
+type testMember struct {
+	providerDef metadata.ProviderDef
+}
+
+func checkEnv(envVar string) string {
+	value, has := os.LookupEnv(envVar)
```
aolfat, line 25: @ahmadnazeri I remember ahmad saying we need to not fail for tests because end2end tests don't have an env or something. I can't quite remember

provider/provider.go
epps: I'm not opposed in theory; however, when it comes to Spark, this "ping" is going to be a lot more involved, in which case `Check` feels more appropriate.

health/health.go
epps: Implementing this as a service with its own API is out of scope for this initial phase, but in the near future, we'll need protos for this, yes.

health/health.go
```diff
@@ -0,0 +1,45 @@
...
+	rec, err := h.metadata.GetProvider(context.Background(), providerName)
+	if err != nil {
```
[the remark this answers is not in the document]
epps: I'll be looking into how we want to handle error messages and logs once I have all supported providers working, but we should definitely probably have an additional method for known error messages returned from providers.

provider/provider.go
epps: `CheckHealth` on the Provider might be better now that I think about it again because the top-level struct doesn't provide context for the operation like it does with `Health.Check(providerName)`.

api/main.go
```diff
@@ -582,7 +585,40 @@ func (serv *MetadataServer) ListProviders(in *pb.Empty, stream pb.Api_ListProvid
...
+			ResourceType: pb.ResourceType_PROVIDER,
+		},
+		Status: status,
+	}
+	_, statusErr := serv.meta.SetResourceStatus(ctx, statusReq)
+	if statusErr != nil {
+		serv.Logger.Errorw("Failed to set provider status", "error", statusErr, "health check error", err)
+		return nil, statusErr
+	}
+	return &pb.Empty{}, err
```
epps, line 621: This logic will now set providers to either a Failed or Ready status based on the results of the health check.

Ready Example (dashboard view):
Screen Shot 2023-10-18 at 9 23 03 PM

Failed Example (dashboard view):
Screen Shot 2023-10-18 at 9 22 00 PM

Thoughts, @sdreyer?

health/health.go
```diff
@@ -0,0 +1,82 @@
...
+func (h *Health) isSupportedProvider(t pt.Type) bool {
+	switch t {
+	case pt.RedisOnline, pt.DynamoDBOnline, pt.PostgresOffline, pt.SnowflakeOffline, pt.SparkOffline:
+		return true
+	default:
+		return false
+	}
+}
+
+func (h *Health) handleError(err error) string {
```
ahmadnazeri, line 56: I am wondering if it makes sense to pass in the provider type and run the general checks and only checks for that provider? Then we can move the individual provider error checks into separate functions. If you agree, we can add it as a TODO, doesn't have to be done in this PR.

health/health.go
```diff
@@ -0,0 +1,82 @@
...
+		return fmt.Sprintf("%s '%s' Check that the password is correct.", baseInitializationErrorMessage, err.Error())
+	}
+	if strings.Contains(err.Error(), "NOAUTH") {
+		return fmt.Sprintf("%s '%s' The server has been configured with the 'requirepass' directive, which requires a password for authentication.", baseInitializationErrorMessage, "NOAUTH HELLO")
+	}
+	// POSTGRES ERRORS
+	if match, matchErr := regexp.MatchString("pq: database .* does not exist", err.Error()); match && matchErr == nil {
+		return fmt.Sprintf("%s '%s' Check that the database name is correct.", baseInitializationErrorMessage, err.Error())
+	}
+	if match, matchErr := regexp.MatchString("pq: password authentication failed for user .*", err.Error()); match && matchErr == nil {
```
ahmadnazeri, line 78: do we need Spark error message handling as well?

health/health_test.go
ahmadnazeri: I am honestly not sure what this is referring to

health/health_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+		fileStoreConfig = &pc.GCSFileStoreConfig{
+			BucketName: os.Getenv("GCS_BUCKET_NAME"),
+			BucketPath: "",
+			Credentials: pc.GCPCredentials{
+				ProjectId: os.Getenv("GCP_PROJECT_ID"),
+				JSON:      creds,
+			},
+		}
+	case fs.Azure:
```
ahmadnazeri, line 96: should we add HDFS?

health/health_test.go
```diff
@@ -0,0 +1,352 @@
...
+				Software:         "snowflake",
+				Tags:             metadata.Tags{},
+				Properties:       metadata.Properties{},
+			},
+		})
+	}
+
+	server, addr := initMetadataServer(t)
+
+	fmt.Printf("Metadata server listening on %s\n", addr)
```
ahmadnazeri, line 296: Is this needed?

provider/dynamodb.go
```diff
@@ -259,6 +259,25 @@ func (store *dynamodbOnlineStore) DeleteTable(feature, variant string) error {
...
+	}
+	if len(listOutput.TableNames) == 0 {
+		return false, fmt.Errorf("(DynamoDB Health Check) no tables found")
+	}
+	scanInput := &dynamodb.ScanInput{TableName: listOutput.TableNames[0], Limit: aws.Int64(1)}
+	scanOutput, err := store.client.Scan(scanInput)
+	if err != nil {
+		return false, err
+	}
+	if len(scanOutput.Items) == 0 {
```
ahmadnazeri, line 275: what happens if the table is empty?

provider/redis.go
```diff
@@ -184,6 +184,18 @@ func (store *redisOnlineStore) DeleteTable(feature, variant string) error {
 	return nil
 }
 
+func (store *redisOnlineStore) Check() (bool, error) {
```
ahmadnazeri, line 187: I think it should be out of scope but we should add more robust checks later on to make sure we have read and write access to the providers.

ahmadnazeri (commented):
Overall, it looks good to me. I left some comments; let me know if there is anything that you want to discuss. 

Also, I was thinking about the Filestores and how we register them as providers. Should they have some sort of checks as well? Or should we propagate the Spark provider health check to the file stores too?

Another question: what should we display on the dashboard if a provider doesn't have a health check? Would it be ready? If yes, should we specify that health checks haven't been performed yet?

api/main.go
```diff
@@ -582,7 +585,40 @@ func (serv *MetadataServer) ListProviders(in *pb.Empty, stream pb.Api_ListProvid
 
 func (serv *MetadataServer) CreateProvider(ctx context.Context, provider *pb.Provider) (*pb.Empty, error) {
 	serv.Logger.Infow("Creating Provider", "name", provider.Name)
-	return serv.meta.CreateProvider(ctx, provider)
+	_, err := serv.meta.CreateProvider(ctx, provider)
+	if err != nil {
+		serv.Logger.Errorw("Failed to create provider", "error", err)
+		return nil, err
+	}
+	var status *pb.ResourceStatus
```
aolfat, line 593: we have a setResourceStatus function that I think we should leverage.

health/health.go
aolfat: I mostly agree with the above. We can have a general error handler but I think its cleaner if each provider is responsible for handling its own errors. The below can get really messy

api/main.go
aolfat: What's your thinking on where to add this health check. We have CreateProvider in api/main.go, metadata/client.go and metadata.go?

health/health_test.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+func TestHealth_Check(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	err := godotenv.Load("../.env")
+	if err != nil {
+		t.Fatal(err)
```
ahmadnazeri, line 34: This might cause the test to fail in CI/CD because the `.env` file wouldn't exist.

provider/unit_test_store.go
```diff
@@ -26,6 +28,10 @@ func (u UnitTestProvider) Config() pc.SerializedConfig {
 	return u.ProviderConfig
 }
 
+func (u UnitTestProvider) Check() (bool, error) {
```
ahmadnazeri, line 31: Should this be `CheckHealth()` now?

runner/copy_test.go
```diff
@@ -748,6 +750,10 @@ func (m MockOnlineStoreTable) Get(entity string) (interface{}, error) {
 	return nil, nil
 }
 
+func (m MockOnlineStoreTable) Check() (bool, error) {
```
ahmadnazeri, line 753: `CheckHealth()`?

ahmadnazeri (approved):
LGTM; just three minor comments

health/health_test.go
epps: @sdreyer? I can add it if you think we need to support it in this first phase. We can also add support later as needed.

provider/dynamodb.go
epps: That's a good point. I was looking for the `SELECT 1;` equivalent in DynamoDB and a couple of SO comments pointed to `Scan`, but perhaps I should only fail the health check if the `Scan` returns an error. Thanks for the attention to detail!

health/health_test.go
anthonylasso: +1, for unit testing you can instead use the testing lib's `.Setenv(_ , _)` mock.
unless there's a specific reason we need to load a `.env` file

epps: Agreed, but many of our other tests are currently already relying on this pattern, right?

provider/unit_test_store.go
epps: That's odd. I thought I globally renamed this method. I'm not sure why it didn't catch these. Thank you!

health/health.go
epps: I originally wrote this pattern until I started to see enough overlap to second guess myself. I've now added @sdreyer's recommendation for typed errors and am switching on these for Spark. We can add others if necessary.

epps:
> Overall, it looks good to me. I left some comments; let me know if there is anything that you want to discuss.
> 
> Also, I was thinking about the Filestores and how we register them as providers. Should they have some sort of checks as well? Or should we propagate the Spark provider health check to the file stores too?
> 
> Another question: what should we display on the dashboard if a provider doesn't have a health check? Would it be ready? If yes, should we specify that health checks haven't been performed yet?

I think we should (eventually) move the health check to the file store providers when the time comes. In that case, we'll have to break out the `CheckHealth` method into a separate interfaces (`HealthChecker`?) or refactor so that the file stores implement the `Provider` interface.

PR #1102, opened 2023-10-19
Databricks Chained DF Transformation Tests

sdreyer:
# Description

Fixing an issue where the root_path would be repeated twice. This was due to the full filestore path being returned when looking for the newest file, instead of just the key after the path. This implements a check that looks to see if the prefix is already included in the key that the filepath object is being created with

### Select type(s) of change

- [ ] Documentation update

# Checklist:


tests/end_to_end/features/steps/spark_registeration.py
```diff
@@ -136,17 +136,31 @@ def step_impl(context):
...
+)
+def step_impl(context, transformation_type, name, sources):
+    source_list = sources.split(",")
+    if transformation_type == "DF":
+
+        @context.spark.df_transformation(
+            name=name,
+            inputs=source_list,
+        )
+        def some_transformation(df):
```
epps, line 150: I'm not seeing this, so apologies if this is obvious, but will this be tested with more than one input?

ahmadnazeri: For these tests, it looks like single inputs. You can see them in the [transformations.feature](https://github.com/featureform/featureform/blob/563dce3107ef9e39760faf317afc9d1e49809bb1/tests/end_to_end/features/transformations.feature#L11-L12) file

ahmadnazeri (approved):

provider/filestore.go
```diff
@@ -141,10 +141,10 @@ func (store *AzureFileStore) CreateFilePath(key string) (filestore.Filepath, err
 	}
 	fp.SetBucket(store.ContainerName)
 	var err error
-	if store.Path != "" {
+	if store.Path != "" && !strings.HasPrefix(key, store.Path) {
```
sdreyer, line 144: Change to remove everytime instead of checking

PR #1109, opened 2023-10-19
Feature: Create timestamp variants for each run

aolfat:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -19,7 +19,6 @@ from .get import *
 from .get_local import *
 from .list import *
 from .list_local import *
-from .names_generator import get_random_name
 from .parse import *
 from .proto import metadata_pb2_grpc as ff_grpc
 from .resources import (
@@ -74,6 +73,8 @@ from .search_local import search_local
 from .sqlite_metadata import SQLiteMetadata
 from .status_display import display_statuses
 from .tls import insecure_channel, secure_channel
+from .variant_names_generator import get_current_timestamp_variant
+from .variant_names_generator import get_random_name
 
 NameVariant = Tuple[str, str]
 
@@ -1185,7 +1186,7 @@ class DFTransformationDecorator:
         variant: str = "",
         name: str = "",
         description: str = "",
-        inputs: list = [],
+        inputs: list = [],  # TODO Make common type for NameVariant
         args: Union[K8sArgs, None] = None,
         source_text: str = "",
     ):
@@ -1213,13 +1214,19 @@ class DFTransformationDecorator:
             raise ValueError(
                 f"Transformation function has more parameters than inputs. \n"
                 f"Make sure each function parameter has a corresponding input in the decorator."
+                f"Make sure each function parameter has a corresponding input in the decorator."
             )
 
         if not isinstance(self.inputs, list):
             raise ValueError("Dataframe transformation inputs must be a list")
 
+        # check that input isn't self referencing
         for nv in self.inputs:
-            if self.name is nv[0] and self.variant is nv[1]:
+            if isinstance(nv, tuple):
+                n, v = nv
+            else:
+                n, v = nv.name_variant()
+            if self.name is n and self.variant is v:
                 raise ValueError(
                     f"Transformation cannot be input for itself: {self.name} {self.variant}"
                 )
@@ -1660,7 +1667,11 @@ class Registrar:
         self.__state = ResourceState()
         self.__resources = []
         self.__default_owner = ""
-        self.__run = get_random_name()
+        self.__run_prefix = ""
+        auto_variant = get_random_name()
+        if os.getenv("FF_TIMESTAMP_VARIANT") is not None:
+            auto_variant = get_current_timestamp_variant(self.__run_prefix)
+        self.__run = auto_variant
 
     def add_resource(self, resource):
         self.__resources.append(resource)
@@ -1699,6 +1710,14 @@ class Registrar:
         if owner == "":
             raise ValueError("Owner must be set or a default owner must be specified.")
         return owner
+
+    def set_run_prefix(self, run_prefix: str = ""):
+        """Set run prefix.
+
+        Args:
+            run_prefix (str): Run prefix to be set.
+        """
+        self.__run_prefix = run_prefix
 
     def set_run(self, run: str = ""):
         """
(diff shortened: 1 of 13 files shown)


metadata/proto/metadata.proto
```diff
@@ -12,36 +12,45 @@ option go_package = "github.com/featureform/metadata/proto";
 package featureform.serving.metadata.proto;
 
 service Metadata {
-    rpc ListFeatures(Empty) returns (stream Feature);
+    rpc CreateUser(User) returns (Empty);
```
aolfat, line 15: I organized these so they're the same as API

metadata/proto/metadata.proto
```diff
@@ -152,6 +168,15 @@ enum ComputationMode {
     CLIENT_COMPUTED = 1;
 }
 
+message ResourceVariant {
```
aolfat, line 171: I debated not having this but I think its something we need.

metadata/metadata.go
```diff
@@ -1663,6 +1768,43 @@ func (serv *MetadataServer) genericList(t ResourceType, send sendFn) error {
...
+		resourceType = FEATURE_VARIANT
+	case *pb.ResourceVariant_LabelVariant:
+		currentResource = &labelVariantResource{resourceVariant.GetLabelVariant()}
+		resourceType = LABEL_VARIANT
+	case *pb.ResourceVariant_TrainingSetVariant:
+		currentResource = &trainingSetVariantResource{resourceVariant.GetTrainingSetVariant()}
+		resourceType = TRAINING_SET_VARIANT
+	}
+
+	resources, err := serv.lookup.ListForType(resourceType)
```
aolfat, line 1790: @crfeliz or @anthonylasso is there a way to have auth around this call (or maybe some other call that I can use) so that we only get all resources given the users auth creds

client/src/featureform/register.py
```diff
@@ -1185,7 +1186,7 @@ def __init__(
         variant: str = "",
         name: str = "",
         description: str = "",
-        inputs: list = [],
+        inputs: list = [],  # TODO Make common type for NameVariant
```
anthonylasso, line 1189: implement or no?

client/src/featureform/register.py
```diff
@@ -1213,13 +1214,19 @@ def __call__(self, fn):
             raise ValueError(
                 f"Transformation function has more parameters than inputs. \n"
                 f"Make sure each function parameter has a corresponding input in the decorator."
+                f"Make sure each function parameter has a corresponding input in the decorator."
```
anthonylasso, line 1217: need this string twice?

metadata/metadata.go
```diff
@@ -482,6 +488,27 @@ func (resource *sourceVariantResource) Update(lookup ResourceLookup, updateRes R
 	return nil
 }
 
+func (resource *sourceVariantResource) IsEquivalent(other ResourceVariant) bool {
+	otherVariant, ok := other.(*sourceVariantResource)
```
anthonylasso, line 492: +1 appreciate the "other" identifier

metadata/metadata.go
anthonylasso: yeah in `main.go` where this is invoked, we can use `authorizer.GetUserResources()` to list them first.

aolfat: It looks like that function should probably have a filter by resource type

anthonylasso: yup

metadata/metadata.go
```diff
@@ -622,6 +649,30 @@ func (resource *featureVariantResource) Update(lookup ResourceLookup, updateRes
...
+	otherProto := otherVariant.serialized
+
+	if thisProto.GetName() == otherProto.GetName() &&
+		proto.Equal(thisProto.GetSource(), otherProto.GetSource()) &&
+		(proto.Equal(thisProto.GetFunction(), otherProto.GetFunction()) ||
+			proto.Equal(thisProto.GetColumns(), otherProto.GetColumns()) &&
+				thisProto.GetProvider() == otherProto.GetProvider()) {
+
+		return true
+	}
```
epps, line 668: Given this boolean logic, it would be nice to have a comment that explains the ways two feature variants could be considered equivalent. That way, if someone else come to this to extend it, it'll be easier to reason about the requirements IMHO.

metadata/metadata.go
```diff
@@ -1663,6 +1768,43 @@ func (serv *MetadataServer) genericList(t ResourceType, send sendFn) error {
...
+	case *pb.ResourceVariant_FeatureVariant:
+		currentResource = &featureVariantResource{resourceVariant.GetFeatureVariant()}
+		resourceType = FEATURE_VARIANT
+	case *pb.ResourceVariant_LabelVariant:
+		currentResource = &labelVariantResource{resourceVariant.GetLabelVariant()}
+		resourceType = LABEL_VARIANT
+	case *pb.ResourceVariant_TrainingSetVariant:
+		currentResource = &trainingSetVariantResource{resourceVariant.GetTrainingSetVariant()}
+		resourceType = TRAINING_SET_VARIANT
+	}
```
epps, line 1788: `default` case with an error for an invalid resource type?

client/src/featureform/register.py
```diff
@@ -3637,8 +3658,8 @@ def df_transformation(
                         f"DF transformation {transformation_message} requires correct inputs "
                         f" '{nv}' is not a valid tuple: {e}"
                     )
-            if inputs[i][1] == "":
-                inputs[i] = (inputs[i][0], self.__run)
+            # if inputs[i][1] == "":
+            #     inputs[i] = (inputs[i][0], self.__run)
```
epps, line 3662: Did you meant to remove this or should you uncomment it?

metadata/metadata.go
aolfat: yeahhh agreed. I was even considering having a set of fields on Resource Variants that dictate the set of equivalent fields

client/src/featureform/register.py
aolfat: whoops

client/src/featureform/register.py
aolfat: i uncommented this because I wanted sterlings thoughts on it hahah.

PR #1121, opened 2023-10-23
Feature: ability to get spark data frames for training sets

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


client/src/featureform/serving.py
```diff
@@ -21,6 +21,8 @@
 from featureform.providers import get_provider, Scalar, VectorType
 from pandas.core.generic import NDFrame
 from pandasql import sqldf
+from pyspark.sql import DataFrame
```
aolfat, line 24: nit: can we give this a alias so the typing isn't as confusing?

client/src/featureform/serving.py
```diff
@@ -1020,6 +1035,45 @@ def dataframe(self) -> pd.DataFrame:
...
+
+        location = (
+            location.replace("s3://", "s3a://")
+            if location.startswith("s3://")
+            else location
+        )
+
+        return file_format, location
+
+    def get_spark_dataframe(
```
aolfat, line 1055: are we exposing this function to the API?

client/src/featureform/serving.py
```diff
@@ -1020,6 +1035,45 @@ def dataframe(self) -> pd.DataFrame:
             self._dataframe.rename(columns={cols.label: "label"}, inplace=True)
             return self._dataframe
 
+    def get_file_format_and_location(self, location):
+        if location.endswith(".csv"):
```
aolfat, line 1039: we have a whole `FileFormat` and `FilePrefix` class for a lot of this, can we use those here?

client/src/featureform/serving.py
```diff
@@ -1020,6 +1035,45 @@ def dataframe(self) -> pd.DataFrame:
...
+            file_format = "csv"
+        else:
+            file_format = "parquet"
+
+        if location.split("/")[-1].startswith("part-"):
```
aolfat, line 1044: lol wutttttt is this. a comment or a separate function por favor

client/src/featureform/serving.py
```diff
@@ -1020,6 +1035,45 @@ def dataframe(self) -> pd.DataFrame:
...
+            if location.startswith("s3://")
+            else location
+        )
+
+        return file_format, location
+
+    def get_spark_dataframe(
+        self, spark: SparkSession, file_format: str, location: str
+    ) -> DataFrame:
+        if file_format not in ["csv", "parquet"]:
```
aolfat, line 1058: this is why we have a fileformat class

client/src/featureform/serving.py
```diff
@@ -1006,6 +1010,17 @@ def dataframe(self) -> pd.DataFrame:
...
         if self._dataframe is not None:
             return self._dataframe
+        elif spark_session is not None:
+            req = serving_pb2.TrainingDataRequest()
+            req.id.name = self._stream.name
+            req.id.version = self._stream.version
+            resp = self._stream._stub.ResourceLocation(req)
+
+            file_format, location = self.get_file_format_and_location(resp.location)
+            self._dataframe = self.get_spark_dataframe(
```
aolfat, line 1020: do you have a python file that shows what this'll look like? Is there a way to do this so you don't have to pass in a spark session?

ahmadnazeri:
```
df = client.training_set("name", "variant").dataframe(spark_session=spark)
```
Unfortunately, you can't because the spark session sets the credentials needed to access the file path.

aolfat: but don't we have all of that stored on the host?

ahmadnazeri: we don't want to retrieve credentials from host. It can lead to more problems.

client/src/featureform/serving.py
```diff
@@ -1020,6 +1035,51 @@ def dataframe(self) -> pd.DataFrame:
             self._dataframe.rename(columns={cols.label: "label"}, inplace=True)
             return self._dataframe
 
+    def __get_file_format_and_location(self, location: str) -> str:
```
sdreyer, line 1038: do you need (str, str) here?

client/src/featureform/serving.py
```diff
@@ -1020,6 +1035,51 @@ def dataframe(self) -> pd.DataFrame:
...
+        try:
+            file_format = FileFormat.get_format(location)
+        except Exception:
+            file_format = "parquet"
+
+        is_individual_part_file = location.split("/")[-1].startswith("part-")
```
sdreyer, line 1044: Can you split these parsing bits into their own functions with comments on what its doing?

client/src/featureform/serving.py
```diff
@@ -1002,10 +1005,22 @@ def dataframe(self) -> pd.DataFrame:
...
+
         if self._dataframe is not None:
             return self._dataframe
+        elif spark_session is not None:
+            req = serving_pb2.TrainingDataRequest()
+            req.id.name = self._stream.name
+            req.id.version = self._stream.version
+            resp = self._stream._stub.ResourceLocation(req)
+
+            file_format, location = self.__get_file_format_and_location(resp.location)
```
sdreyer, line 1019: Is there a way to test these with local files and spark sessions?

client/src/featureform/serving.py
```diff
@@ -1020,6 +1036,60 @@ def dataframe(self) -> pd.DataFrame:
             self._dataframe.rename(columns={cols.label: "label"}, inplace=True)
             return self._dataframe
 
+    def _get_file_format(self, location: str) -> str:
```
aolfat, line 1039: I would argue that you can put this function in the FileFormat class where you can pass in it a default

client/src/featureform/serving.py
```diff
@@ -1002,10 +1005,23 @@ def dataframe(self) -> pd.DataFrame:
...
+            df: Union[pd.DataFrame, pyspark.sql.DataFrame] A DataFrame containing the training set.
         """
+
         if self._dataframe is not None:
             return self._dataframe
+        elif spark_session is not None:
+            req = serving_pb2.TrainingDataRequest()
+            req.id.name = self._stream.name
+            req.id.version = self._stream.version
+            resp = self._stream._stub.ResourceLocation(req)
```
aolfat, line 1017: nit: this should probably be it's own method somewhere so that anyone can use this API

client/src/featureform/enums.py
```diff
@@ -142,13 +142,15 @@ def is_supported(cls, file_path: str) -> bool:
...
     @classmethod
-    def get_format(cls, file_path: str) -> str:
+    def get_format(cls, file_path: str, default: str = "") -> str:
         file_name = path.basename(file_path)
 
         for file_format in cls:
             if fnmatch(file_name, f"*.{file_format.value}"):
                 return file_format.value
 
+        if default != "":
```
sdreyer, line 152: Whats this default logic doing?

serving/serving.go
```diff
@@ -381,3 +382,57 @@ func (serv *FeatureServer) getVectorTable(ctx context.Context, fv *metadata.Feat
...
+	id := req.GetId()
+	name, variant := id.GetName(), id.GetVersion()
+	serv.Logger.Infow("Getting the Resource Location:", "Name", name, "Variant", variant)
+
+	tv, err := serv.Metadata.GetTrainingSetVariant(ctx, metadata.NameVariant{Name: name, Variant: variant})
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get training set variant")
+	}
+
+	if tv.Status() != metadata.READY {
```
sdreyer, line 400: Do we want this logic? We're not pulling the training set here, just getting its location.

ahmadnazeri: we do need it because we need to pull the latest file for it.

client/src/featureform/enums.py
ahmadnazeri: The idea is that `get_format` would fail if you don't provide it a default format; thus, using the empty string. And if you have a default value then it would default to that value if it doesn't pick up a certain format; ex. directories

serving/serving.go
sdreyer: But this doesn't pull a file, it returns a filepath. Its fine but feel like the logic should be somewhere else

client/src/featureform/enums.py
sdreyer: I guess do we even need it if we're just returning the value we give it?

ahmadnazeri: I had it separately then Ali wanted it within this class. So moved it here.

sdreyer (approved):

PR #1130, opened 2023-10-30
Feature (Timestamp Variants) Add GetEquivalent API

aolfat:
# Description

Adds an API to determine if two resource variants are equivalent. Will be used by the client to check if an equivalent variant exists so we could leverage it. Also uses the check in the event that the same name variant exists but with different key components.

i.e. You register a transformation with the same name variant but you change the definition

### Select type(s) of change

- [ ] Documentation update

# Checklist:


metadata/client.go
```diff
@@ -268,7 +268,7 @@ func (def FeatureDef) ResourceType() ResourceType {
 	return FEATURE_VARIANT
 }
 
-func (client *Client) CreateFeatureVariant(ctx context.Context, def FeatureDef) error {
+func (def FeatureDef) Serialize() (*pb.FeatureVariant, error) {
```
aolfat, line 271: I extracted these functions out for testing

metadata/metadata.go
```diff
@@ -1572,6 +1832,20 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	return &pb.Empty{}, nil
 }
 
+func isEquivalent(this Resource, other Resource) bool {
```
ahmadnazeri, line 1835: Is the idea that we will expand it to other Resources too?

metadata/metadata.go
```diff
@@ -1511,6 +1721,47 @@ func (serv *MetadataServer) GetModels(stream pb.Metadata_GetModelsServer) error
...
+		return nil, fmt.Errorf("unknown resource variant type: %v", req.Resource)
+	}
+
+	resources, err := serv.lookup.ListForType(resourceType)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, res := range resources {
+		other, ok := res.(ResourceVariant)
```
ahmadnazeri, line 1753: wouldn't these always be ResourceVariants given the cases above?

metadata/metadata.go
```diff
@@ -1511,6 +1721,47 @@ func (serv *MetadataServer) GetModels(stream pb.Metadata_GetModelsServer) error
...
+		if !ok {
+			continue
+		}
+		if currentResource.IsEquivalent(other) {
```
ahmadnazeri, line 1757: can we remove the above conditional statement and just do 
```if ok && currentResource.IsEquivalent(other) {```

File: metadata/metadata.go [hunk 5]
```diff
@@ -886,6 +1056,46 @@ func (resource *trainingSetVariantResource) Update(lookup ResourceLookup, update
...
+	*/
+	otherVariant, ok := other.(*trainingSetVariantResource)
+	if !ok {
+		return false
+	}
+
+	thisProto := resource.serialized
+	otherProto := otherVariant.serialized
+
+	if otherProto.Status.Status == pb.ResourceStatus_FAILED {
```
ahmadnazeri, line 1076: what should happen if a resource status is PENDING?

lib/proto_utils.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// EqualProtoContents compares two slices of proto messages to check that the contents are the same.
+// It excludes the order and dedupes.
+func EqualProtoContents[T proto.Message](a, b []T) bool {
+	// We marshal the proto messages to strings so that we can compare them in a set
+	marshaledA := fp.Map[T, string](func(x T) string {
+		marshal, err := proto.Marshal(x)
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
anthonylasso: have the same question, if it's an equals check, shouldn't the caller handle the true/false state?

metadata/metadata.go
```diff
@@ -482,6 +502,70 @@ func (resource *sourceVariantResource) Update(lookup ResourceLookup, updateRes R
 	return nil
 }
 
+func (resource *sourceVariantResource) IsEquivalent(other ResourceVariant) bool {
```
anthonylasso, line 505: good identifier choice "other"

metadata/metadata_test.go
```diff
@@ -2027,3 +2028,247 @@ func TestSourceShallowMapOK(t *testing.T) {
 	assertEqual(t, slices.Contains(sourceVariantResource.Tags, "test.inactive"), true)
 	assertEqual(t, sv.Properties()["test.map.key"], sourceVariantResource.Properties["test.map.key"])
 }
+
+// TODO split these up into better tests
```
anthonylasso, line 2032: probably split by resource type. easier to digest/diagnose

metadata/metadata_test.go
```diff
@@ -2027,3 +2028,247 @@ func TestSourceShallowMapOK(t *testing.T) {
...
+func Test_GetEquivalent(t *testing.T) {
+	serv, addr := startServNoPanic(t)
+	client := client(t, addr)
+	context := context.Background()
+
+	redisConfig := pc.RedisConfig{
```
anthonylasso, line 2038: this could all go into a setup function that the split tests each call

lib/proto_utils.go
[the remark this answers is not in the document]
aolfat: yeahh, good call; should return an error

metadata/metadata.go
aolfat: yeahh, I should throw an error instead

metadata/metadata.go
aolfat: if should never be !ok though. Otherwise something we're doing is really wrong

metadata/metadata.go
aolfat: Eventually we could do it; I think it will really make sense for providers

aolfat: But right now the equivalnce check if for finding equivalent variants

metadata/metadata_test.go
aolfat: yeahhh, there's 100000x better ways to do this lol

metadata/metadata.go
aolfat: we just create another variant right now

metadata/client.go
```diff
@@ -523,8 +539,8 @@ func (def TrainingSetDef) ResourceType() ResourceType {
 	return TRAINING_SET_VARIANT
 }
 
-func (client *Client) CreateTrainingSetVariant(ctx context.Context, def TrainingSetDef) error {
-	serialized := &pb.TrainingSetVariant{
+func (def TrainingSetDef) Serialize() *pb.TrainingSetVariant {
```
ahmadnazeri, line 542: Are protos serialized objects or just objects?

metadata/metadata.go
```diff
@@ -482,6 +511,74 @@ func (resource *sourceVariantResource) Update(lookup ResourceLookup, updateRes R
...
+	}
+
+	thisProto := resource.serialized
+	otherProto := otherVariant.serialized
+
+	isDefinitionEqual := false
+	var err error
+	switch thisDef := thisProto.Definition.(type) {
+	case *pb.SourceVariant_Transformation:
+		if otherDef, ok := otherProto.Definition.(*pb.SourceVariant_Transformation); ok {
```
ahmadnazeri, line 534: we don't check for `!ok` here. Is the reason because `isDefinitionEqual` will be false so we will return `false, nil`?

metadata/metadata.go
```diff
@@ -482,6 +511,74 @@ func (resource *sourceVariantResource) Update(lookup ResourceLookup, updateRes R
...
+			isDefinitionEqual, err = isProtoDefinitionEqual(thisDef, otherDef)
+			if err != nil {
+				return false, fmt.Errorf("error comparing proto messages: %v", err)
```
ahmadnazeri, line 537: from the user's perspective, would the error message make sense? would they know what is `photo messages`?

metadata/metadata.go
```diff
@@ -482,6 +511,74 @@ func (resource *sourceVariantResource) Update(lookup ResourceLookup, updateRes R
...
+		thisProto.GetOwner() == otherProto.GetOwner() &&
+		thisProto.GetProvider() == otherProto.GetProvider() &&
+		isDefinitionEqual {
+
+		return true, nil
+	}
+	return false, nil
+}
+
+func isProtoDefinitionEqual(thisDef, otherDef *pb.SourceVariant_Transformation) (bool, error) {
```
ahmadnazeri, line 557: This is for transformations only?

metadata/metadata.go
```diff
@@ -622,6 +727,46 @@ func (resource *featureVariantResource) Update(lookup ResourceLookup, updateRes
...
+	- Function
+	- Provider
+	- Entity
+	- Owner
+	- Type
+	*/
+
+	otherVariant, ok := other.(*featureVariantResource)
+	if !ok {
+		return false, nil
```
ahmadnazeri, line 744: does it make sense to return an error that you are comparing two different types of resources?

metadata/metadata.go
aolfat: yeah, I can rename it

metadata/metadata.go
aolfat: yeah I could do that, my initial thnking was, "well if they're different they're not equal, but that shouldn't even happen"

metadata/client.go
aolfat: they're the serialization of objects

metadata/metadata.go
aolfat: yeah, it'll just default to false

metadata/metadata.go
aolfat: no, but what else could you say?

metadata/metadata.go
aolfat: actually no, because i'm allowing comparisons between all resource variants

metadata/metadata.go
ahmadnazeri: `error comparing two resources: %v`?

metadata/metadata.go
ahmadnazeri: wait so a feature will be compared to a primary source?

aolfat: I'm saying any resource variant can be compared to another resource variant

metadata/metadata.go
```diff
@@ -219,12 +220,32 @@ func (err *ResourceExists) GRPCStatus() *status.Status {
 	return status.New(codes.AlreadyExists, err.Error())
 }
 
+type ResourceChanged struct {
```
sdreyer, line 223: Should be ResourceChangedError

metadata/metadata.go
```diff
@@ -219,12 +220,32 @@ func (err *ResourceExists) GRPCStatus() *status.Status {
...
+	ID ResourceID
+}
+
+func (err *ResourceChanged) Error() string {
+	id := err.ID
+	errMsg := fmt.Sprintf("%s changed. Use a new variant.", id.String())
```
sdreyer, line 229: What does it mean for a resource to change and why use a new variant?

metadata/metadata.go
sdreyer, line 534: Do we need to check the sources as well as the definition?

metadata/metadata.go
sdreyer: bump

metadata/metadata.go
sdreyer: nm see it below

metadata/metadata.go
```diff
@@ -1511,6 +1759,66 @@ func (serv *MetadataServer) GetModels(stream pb.Metadata_GetModelsServer) error
...
+		return nil, fmt.Errorf("unknown resource variant type: %v", req.Resource)
+	}
+
+	resources, err := serv.lookup.ListForType(resourceType)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, res := range resources {
+		resourceStatus := res.GetStatus()
```
sdreyer, line 1800: Can you put the contents of this loop into its own func for readability?

aolfat: gladly

metadata/metadata.go
aolfat: I might just move this part to another PR. The idea is if you try and register a variant that already exists, if the key attributes on it have changed (for example for a feature, the type has changed) it should say; "look dude, you changed some important shit on this variant; you should register it as a new variant."

lib/proto_utils_test.go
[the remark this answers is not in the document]
aolfat: might need to point me to that

metadata/metadata.go
aolfat: None of the errors above do that lol. ResourceExists and ResourceNotFound don't have the Error suffix. Just following simba's code

lib/proto_utils_test.go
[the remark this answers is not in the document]
sdreyer: In the testing.yml change:
`run: go test -v -coverpkg=./... -coverprofile coverage/cover.out.tmp ./metadata/`
to
`run: go test -v -coverpkg=./... -coverprofile coverage/cover.out.tmp ./metadata/ ./lib/` 
(or wherever that directory is)

metadata/metadata.go
sdreyer: Its best practice https://gobyexample.com/errors 

Or move it into an errors.go file, or both

metadata/metadata.go
sdreyer: Is this if you manually set a variant then change it?

aolfat: yisshhh

metadata/metadata.go
aolfat: I was just hoping my PR wouldn't be a big refactor of other pieces too. If I change mine I'll have to change the other ones as well. Usually consistency is more important than best practice (if you're already not following best practice). I'm more than happy to do it in another PR (you know how much I love refactoring for better practices).

metadata/metadata.go
aolfat: let me try

ahmadnazeri (approved):

PR #1133, opened 2023-10-31
Feature (Timestamp variants): defer fetching of name and variant until create

aolfat:
# Description

For part of the timestamp variant work we want to defer the retrieval of the name variant of a dependency until at create in case the variant changes.

Let's say you register

tf1 
tf2
tf3

If, while registering tf2, we find an equivalent tf1 and use it's variant. We should use that new variant to tf2

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/register.py
+++ b/client/src/featureform/register.py
@@ -11,7 +11,6 @@ from typing import Dict, Tuple, Callable, List, Union, Optional
 import dill
 import pandas as pd
 from typeguard import typechecked
-
 from .enums import FileFormat
 from .exceptions import InvalidSQLQuery
 from .file_utils import absolute_file_paths
@@ -1024,6 +1023,7 @@ class SubscriptableTransformation:
         provider,
         decorator_register_resources_method,
         decorator_name_variant_method,
+        transformation
     ):
         # if not self.__has_return_statement(fn):
         #     raise Exception(
@@ -1042,7 +1042,10 @@ class SubscriptableTransformation:
         # so the code has been reverted to the original syntax, which simply passes a reference to
         # the decorator methods to the SubscriptableTransformation class.
         self.register_resources = decorator_register_resources_method
-        self.name_variant = decorator_name_variant_method
+        self.transformation = transformation
+
+    def name_variant(self):
+        return self.transformation.name_variant()
 
     def __getitem__(self, columns: List[str]):
         col_len = len(columns)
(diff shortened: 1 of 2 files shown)


client/src/featureform/register.py
```diff
@@ -3835,7 +3841,12 @@ def register_entity(
 
     def register_column_resources(
         self,
-        source: Union[NameVariant, SourceRegistrar, SQLTransformationDecorator],
+        source: Union[
+            NameVariant,
+            SourceRegistrar,
+            SQLTransformationDecorator,
```
aolfat, line 3847: I plan on making a new class for these two decorators called SourceVariantProvider

client/src/featureform/register.py
```diff
@@ -4089,25 +4096,22 @@ def register_training_set(
             elif resource_label != ():
                 raise ValueError("A training set can only have one label")
 
-        if isinstance(label, LabelColumnResource):
-            label = label.name_variant()
```
aolfat, line 4093: this is a good example to understand what's going on, by removing this we're not locked into the label's name variant until we finally go to create

client/src/featureform/register.py
```diff
@@ -3637,8 +3643,8 @@ def df_transformation(
                         f"DF transformation {transformation_message} requires correct inputs "
                         f" '{nv}' is not a valid tuple: {e}"
                     )
-            if inputs[i][1] == "":
-                inputs[i] = (inputs[i][0], self.__run)
+                if inputs[i][1] == "":
```
ahmadnazeri, line 3646: any reason why this is moved within the elif statement?

client/src/featureform/resources.py
```diff
@@ -1353,7 +1357,10 @@ def get(self, stub) -> "FeatureVariant":
             error=feature.status.error_message,
         )
 
-    def _create(self, stub) -> None:
+    def _create(self, stub) -> Optional[str]:
```
ahmadnazeri, line 1360: it is never returning a string?

client/src/featureform/resources.py
```diff
@@ -1639,7 +1646,10 @@ def get(self, stub) -> "LabelVariant":
             error=label.status.error_message,
         )
 
-    def _create(self, stub) -> None:
+    def _create(self, stub) -> Optional[str]:
```
ahmadnazeri, line 1649: never returning a string?

client/src/featureform/register.py
```diff
@@ -1218,8 +1222,15 @@ def __call__(self, fn):
         if not isinstance(self.inputs, list):
             raise ValueError("Dataframe transformation inputs must be a list")
 
+        # check that input isn't self referencing
         for nv in self.inputs:
-            if self.name is nv[0] and self.variant is nv[1]:
+            if isinstance(
+                nv, tuple
+            ):  # TODO all that should be called here is name_variant()
```
anthonylasso, line 1229: with the above changes, do we still need to call `ininstance()`, can we just rely on `nv.name_variant()`?

PR #1161, opened 2023-11-16
Materialization via S3 Import to DynamoDB

epps:
# Description

DynamoDB offers table creation via a direct import from S3 (see [docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/S3DataImport.HowItWorks.html) for details). Given some users already use this feature to save on DynamoDB costs, supporting this as a materialization option is valuable.

### Select type(s) of change

- [x] Documentation update

# Checklist:


client/src/featureform/register.py
```diff
@@ -2810,9 +2810,9 @@ def register_cassandra(
     def register_dynamodb(
         self,
         name: str,
-        access_key: str,
-        secret_key: str,
+        credentials: AWSCredentials,
         region: str,
+        import_from_s3: bool = False,
```
ahmadnazeri, line 2815: should this be `use_import_from_s3`? I know it's longer

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -329,9 +332,18 @@ def execute_sql_query(job_type, output_uri, sql_query, spark_configs, source_lis
...
         output_uri_with_timestamp = f"{output_uri.rstrip('/')}/{safe_datetime}"
 
-        output_dataframe.write.option("header", "true").mode("overwrite").parquet(
-            output_uri_with_timestamp
-        )
+        if output_format == "parquet":
+            output_dataframe.write.option("header", "true").mode("overwrite").parquet(
+                output_uri_with_timestamp
+            )
+        elif output_format == "csv":
```
ahmadnazeri, line 339: I am assuming that we don't include the headers because of the s3 import. I wonder if it makes sense to have another flag for it because in the future we might want to write as csv. Or I guess once you add your comment, it should make sense.

provider/spark.go
```diff
@@ -1829,6 +1868,9 @@ func blobSparkMaterialization(id ResourceID, spark *SparkOfflineStore, isUpdate
 		spark.Logger.Errorw("Problem creating spark submit arguments", "error", err)
 		return nil, fmt.Errorf("error with getting spark submit arguments %v", sparkArgs)
 	}
+	if outputFormat == filestore.CSV {
```
ahmadnazeri, line 1871: can we just always append the output format regardless of it being CSV or not?

runner/s3_import_dynamodb.go
```diff
@@ -0,0 +1,247 @@
...
+		r.Logger.Errorf("failed to list files in source dir path %s: %v", sourceDirPath, err)
+		return nil, err
+	}
+
+	if len(files) == 0 {
+		r.Logger.Errorf("no files found in source dir path %s", sourceDirPath)
+		return nil, fmt.Errorf("no files found in source dir path %s", sourceDirPath)
+	}
+
+	sourceFile := files[0]
```
ahmadnazeri, line 103: will there always be only one file?

provider/online.go
```diff
@@ -28,6 +29,19 @@ type OnlineStore interface {
...
+	Status() string
+	ErrorMessage() string
+}
+
+type ImportableOnlineStore interface {
+	OnlineStore
+	ImportTable(feature, variant string, valueType ValueType, source fs.Filepath) (ImportID, error)
+	GetImport(id ImportID) (Import, error)
+}
+
```
epps, line 44: @sdreyer, this was my solution to not making the DynamoDB provider struct public in the `provider` package. Let me know if you'd prefer that instead of this abstraction.

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -329,9 +332,18 @@ def execute_sql_query(job_type, output_uri, sql_query, spark_configs, source_lis
         # remove the '/' at the end of output_uri in order to avoid double slashes in the output file path.
         output_uri_with_timestamp = f"{output_uri.rstrip('/')}/{safe_datetime}"
 
-        output_dataframe.write.option("header", "true").mode("overwrite").parquet(
-            output_uri_with_timestamp
-        )
+        if output_format == "parquet":
```
anthonylasso, line 335: nit, should we enum these?

provider/spark.go
```diff
@@ -590,20 +605,46 @@ func (q defaultPythonOfflineQueries) materializationCreate(schema ResourceSchema
...
 			ORDER BY
 				maxr.max_row DESC`, schema.Entity, schema.Value, schema.Entity)
 	}
-	return fmt.Sprintf(
-		"SELECT entity, value, ts, ROW_NUMBER() over (ORDER BY (SELECT NULL)) AS row_number, rn2 FROM "+
-			"(SELECT entity, value, ts, ROW_NUMBER() OVER (PARTITION BY entity ORDER BY ts DESC) AS rn2 FROM "+
-			"(SELECT entity, value, ts, rn FROM (SELECT %s AS entity, %s AS value, %s AS ts, "+
-			"ROW_NUMBER() OVER (ORDER BY (SELECT NULL)) AS rn FROM %s) t ORDER BY rn DESC) t2 ) t3 WHERE rn2=1",
-		schema.Entity, schema.Value, timestampColumn, "source_0")
+	return fmt.Sprintf(`SELECT
```
anthonylasso, line 616: did the formatting bork out?

aolfat: yeahh i prefer the previous formatting lol; or maybe have a file for sql string constants

provider/scripts/spark/offline_store_spark_runner.py
aolfat: probz

client/src/featureform/register.py
epps: We could. Are you suggesting the change to make the param more boolean sounding? If so, I think we should go with `should_import_from_s3`, yeah?

provider/scripts/spark/offline_store_spark_runner.py
epps: I actually think I might be able to add the headers back in after changing the `SELECT` statement in the materialization queries. Let me give that a shot so this flag is responsible for one thing, which would of course be better. Stay tuned.

runner/s3_import_dynamodb.go
epps: There should be _at least_ one file and all I need it for is to have a fully-qualified key prefix for the part files given that the import tool doesn't attempt to look at file extensions.

client/src/featureform/register.py
aolfat: should this always be done by default?

aolfat: in fact when would you even want it to be off?

coordinator/coordinator.go
```diff
@@ -169,7 +169,7 @@ func (c *Coordinator) AwaitPendingLabel(labelNameVariant metadata.NameVariant) (
 }
 
 type JobSpawner interface {
-	GetJobRunner(jobName string, config runner.Config, resourceId metadata.ResourceID) (types.Runner, error)
+	GetJobRunner(jobName runner.RunnerName, config runner.Config, resourceId metadata.ResourceID) (types.Runner, error)
```
aolfat, line 172: good refactor!

provider/dynamodb.go
```diff
@@ -275,6 +276,108 @@ func (store *dynamodbOnlineStore) CheckHealth() (bool, error) {
 	return true, nil
 }
 
+func (store *dynamodbOnlineStore) ImportTable(feature, variant string, valueType ValueType, source filestore.Filepath) (ImportID, error) {
+	fmt.Printf("Checking metadata table for existing table %s\n", GetTablename(store.prefix, feature, variant))
```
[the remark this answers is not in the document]
epps: The linter provider by the LSP highlights capitalized error strings, but other than that, I've always capitalized print/log statements. But I think we _could_ benefit from standardizing around one or the other.

provider/dynamodb.go
```diff
@@ -275,6 +276,108 @@ func (store *dynamodbOnlineStore) CheckHealth() (bool, error) {
...
+	if err == nil {
+		return "", &TableAlreadyExists{feature, variant}
+	}
+	fmt.Printf("Updating metadata table %s\n", GetTablename(store.prefix, feature, variant))
+	err = store.UpdateMetadataTable(GetTablename(store.prefix, feature, variant), valueType)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Printf("Building import table input for %s\n", GetTablename(store.prefix, feature, variant))
```
aolfat, line 291: out of the scope for this but we should just have loggers and avoid the print statements haha

provider/dynamodb.go
```diff
@@ -275,6 +276,108 @@ func (store *dynamodbOnlineStore) CheckHealth() (bool, error) {
...
+		// https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.47.7/service/dynamodb#S3BucketSource
+		S3BucketSource: &dynamodb.S3BucketSource{
+			S3Bucket: aws.String(source.Bucket()),
+			// To avoid importing Spark's _committed/_SUCCESS files, we use a prefix that contains the beginning of the
+			// part-file naming conventions (e.g. `part-`). This ensures we only import the actual data files.
+			S3KeyPrefix: aws.String(fmt.Sprintf("%s/part-", source.KeyPrefix())),
+		},
+
+		// https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.47.7/service/dynamodb#TableCreationParameters
+		TableCreationParameters: &dynamodb.TableCreationParameters{
```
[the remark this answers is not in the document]
epps: @anthonylasso, you mean in Featureform? In AWS you can see all S3 imports, but I agree it might be good to see it in our UI, too. CC @sdreyer

provider/spark.go
epps: I'd prefer a separate SQL file. Inlined SQL is often hard to read for subqueries, etc. I'll give this a shot.

provider/online.go
```diff
@@ -28,6 +29,19 @@ type OnlineStore interface {
...
 }
 
+type ImportID string
+
+type Import interface {
+	Status() string
+	ErrorMessage() string
+}
+
+type ImportableOnlineStore interface {
```
aolfat, line 39: Can we add a comment on what an ImportableOnlineStore is?

provider/dynamodb.go
epps: I agree. For some things (e.g. `Coordinator`) we do, and I like using them; however, for other components (e.g. providers) we haven't taken that step.

provider/dynamodb.go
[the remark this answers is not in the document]
anthonylasso: sorry should have clarified, yeah on our end (mainly for debugging etc)

coordinator/coordinator.go
```diff
@@ -823,6 +827,36 @@ func (c *Coordinator) runFeatureMaterializeJob(resID metadata.ResourceID, schedu
...
 	c.Logger.Debugw("Resource Table Created", "id", featID, "schema", schema)
+
+	// TODO: refactor this into a separate method
+	if featureProvider.Type() == string(pt.DynamoDBOnline) {
+		c.Logger.Debugw("Feature provider is DynamoDB", "id", featID)
+		config := pc.DynamodbConfig{}
+		if err := config.Deserialize(featureProvider.SerializedConfig()); err != nil {
+			return fmt.Errorf("could not deserialize DynamoDB config due to error: %v", err)
+		}
+		if config.ImportFromS3 {
```
ahmadnazeri, line 838: should we also check that the offline store is `s3`?

provider/spark.go
epps: So, I _can_ do this, but I set the default value for this arg to parquet. Given this, do you still want to append parquet to all arg lists or simple add a comment noting this?

coordinator/coordinator.go
epps: At this point in the materialization, I can't effectively do that; however, I can add a check in the runner factory to ensure we throw an exception if the file store isn't S3.

provider/dynamodb.go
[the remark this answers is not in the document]
epps: I'm going to leave this up to @sdreyer to determine if we want to include this in the original scope. We have several more items associated with DynamoDB, so it's very possible we take care of this at a later stage.

provider/scripts/spark/offline_store_spark_runner.py
epps: OK, I spoke too soon: we're using slightly different naming conventions for keys in DynamoDB, so I will need to provide a list of headers at the time of import. I'll look into adding a `--headers` flag to the runner script.

epps: `--headers include/exclude` arg has been added to direct the runner script whether to write to the output file. I've added an exception if used for parquet purely given our iterators rely on the headers.

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -329,9 +332,18 @@ def execute_sql_query(job_type, output_uri, sql_query, spark_configs, source_lis
...
+            output_dataframe.write.option("header", "true").mode("overwrite").parquet(
+                output_uri_with_timestamp
+            )
+        elif output_format == "csv":
+            # TODO: explain why we're not adding the header option here
```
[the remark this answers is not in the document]
epps: Added.

runner/s3_import_dynamodb.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		r.Logger.Errorf("failed to create source dir path for resource %s: %v", r.ID.ToFilestorePath(), err)
+		return nil, err
+	}
+
+	files, err := sparkOffline.Store.List(sourceDirPath, filestore.CSV)
+	if err != nil {
+		r.Logger.Errorf("failed to list files in source dir path %s: %v", sourceDirPath, err)
+		return nil, err
+	}
```
epps, line 98: @sdreyer, I really wish I didn't have to break our abstractions to do this, but I need a file path from the newly created materialization. I _could_ change the `Materialization` interface to include a new method so that I don't have to rely on this type assertion. Thoughts?

client/src/featureform/register.py
```diff
@@ -2825,17 +2825,16 @@ def register_dynamodb(
...
         )
         ```
 
         Args:
             name (str): (Immutable) Name of DynamoDB provider to be registered
             region (str): (Immutable) Region to create dynamo tables
-            access_key (str): (Mutable) An AWS Access Key with permissions to create DynamoDB tables
-            secret_key (str): (Mutable) An AWS Secret Key with permissions to create DynamoDB tables
+            credentials (AWSCredentials): (Mutable) AWS credentials with permissions to create DynamoDB tables
+            should_import_from_s3 (bool): (Mutable) Determines whether feature materialization will occur via a direct import of data from S3 to new table
```
sdreyer, line 2837: Should we include a link here to the amazon docs?

coordinator/coordinator.go
```diff
@@ -193,7 +207,7 @@ func (k *KubernetesJobSpawner) GetJobRunner(jobName string, config runner.Config
 	fmt.Println("GETJOBRUNNERID:", resourceId)
 	kubeConfig := kubernetes.KubernetesRunnerConfig{
 		EnvVars: map[string]string{
-			"NAME":             jobName,
+			"NAME":             string(jobName),
```
sdreyer, line 210: Nit: Adding a jobName.String() method is cleaner

coordinator/coordinator.go
```diff
@@ -399,20 +413,10 @@ func (c *Coordinator) runTransformationJob(transformationConfig provider.Transfo
...
-		}
-	}
-	if status == metadata.FAILED {
-		return ResourceAlreadyFailedError{
-			resourceID: resID,
-		}
-	}
-	if err := c.Metadata.SetStatus(context.Background(), resID, metadata.PENDING, ""); err != nil {
-		return fmt.Errorf("set pending status for transformation job: %v", err)
+	err = c.setPending(resID, transformation.Status())
```
sdreyer, line 417: Nice

coordinator/coordinator.go
```diff
@@ -823,54 +792,106 @@ func (c *Coordinator) runFeatureMaterializeJob(resID metadata.ResourceID, schedu
...
+		OnlineConfig:  featureProvider.SerializedConfig(),
+		OfflineConfig: sourceProvider.SerializedConfig(),
+		ResourceID:    provider.ResourceID{Name: resID.Name, Variant: resID.Variant, Type: provider.Feature},
+		VType:         provider.ValueTypeJSONWrapper{ValueType: vType},
+		Cloud:         runner.LocalMaterializeRunner,
+		IsUpdate:      false,
+	}
+
+	isImportToS3Enabled := false
+	if featureProvider.Type() == string(pt.DynamoDBOnline) {
```
sdreyer, line 808: could probably pull this out into a function:
`isImportToS3Enabled, err := c.s3ImportCheck(featureProvider)`

coordinator/coordinator.go
```diff
@@ -823,54 +792,106 @@ func (c *Coordinator) runFeatureMaterializeJob(resID metadata.ResourceID, schedu
...
+	if !isCronRunner {
+		return fmt.Errorf("kubernetes runner does not implement schedule")
+	}
+	if err := cronRunner.ScheduleJob(kubernetes.CronSchedule(schedule)); err != nil {
+		return fmt.Errorf("schedule materialize job in kubernetes: %v", err)
+	}
+	return nil
+}
+
+func (c *Coordinator) materializeFeatureViaS3Import(id metadata.ResourceID, config runner.MaterializedRunnerConfig) error {
```
sdreyer, line 877: Should have a check in here that makes sure that the source filestore is actually s3. Was considering doing it later but feel like it should get checked before we attempt to spun up a worker

provider/dynamodb.go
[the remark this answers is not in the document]
sdreyer: It would be nice to see but I feel like we could leave it as a separate enhancement

runner/s3_import_dynamodb.go
sdreyer: Discussed, not a great way to do this without a large refactor

sdreyer (approved):

Issue #1155, opened 2023-11-10
[Feature Request]: Searchable pre-made tags


ihkap11:
### Description

In the search bar in FF UI does not search through the tags.
Plan is to make tags searchable. 

Potential code file that needs changing:
Code is in metadata/search/search.go

TODO: add screenshot for current behavior vs desired behavior

PR #1167, opened 2023-11-21
I 1155/feature/searchable premade tags

ihkap11:
# Description
Make pre-made tags searchable.

### Task 1: Premade tags should be searchable (`metadata`)
- `ResourceDoc`  is the structure that can expose Tags as a searchable parameter for Meilisearch. 
- Test strategy: Using personal tokens for running Meilisearch locally, and testing through `search_test.go`
image

### Task 2:  New tags added to resource should be searchable (from` metadata dashboard`)
- PostTags allows updating variant resource's tags.
- Updating resources in the Meilisearch resource index makes the newly introduced tags appear in search results.
- Test Strategy: 

https://github.com/featureform/featureform/assets/31574850/b2f384a0-0129-4fe8-8c9e-93176677cd39

## Type of change
Feature

https://github.com/featureform/featureform/issues/1155

### Select type(s) of change

- [ ]  Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

# Checklist:


metadata/metadata.go
```diff
@@ -272,6 +273,7 @@ func (wrapper SearchWrapper) Set(id ResourceID, res Resource) error {
 		Name:    id.Name,
 		Type:    id.Type.String(),
 		Variant: id.Variant,
+		Tags: 	 id.Tags,
```
ihkap11, line 276: Hey @ahmadnazeri,
`Set` method is getting its content from `ResourceID` struct. But adding `Tags` to it (which is of `[]string` type) breaks the backward compatibility:

1. using maps with `ResrourceID` which fails since `map` can't compare `[]string` types.
2. will have to pass Tags whereever `ResourceID` is called. 
3. for functions like [ResourceLocation](https://github.com/featureform/featureform/blob/1b6004810e9fb7861dc2317309c5d2dfe7800c37/serving/serving.go#L386), will have to use GetTags() that will go into [ResourceID](https://github.com/featureform/featureform/blob/1b6004810e9fb7861dc2317309c5d2dfe7800c37/serving/serving.go#L424) 

So one way I can think of making it backward compatible is:
- For point 1,  create a custom mapping function that handles comparing []string types
- Point 2 and 3 will change lots of files so test coverage will be essential to know if everything is working.

Let me know if this seems the right way to handle this or creating another struct with tags is a better way to  implement this.

sdreyer: Is it the schema in the search index thats breaking or just the change to the struct that would break something?

ihkap11: Line 281 (just below in this page) has a map function:
`type LocalResourceLookup map[ResourceID]Resource` 

Since `id.Tags` has type `[]String` (line 268, code block shown below), `ResourceID` has the same tags type which will trigger an error on line 281 because maps can't compare []string types inherently. So to handle such issue I should create a custom map, I guess. 
```
func (wrapper SearchWrapper) Set(id ResourceID, res Resource) error {
	if err := wrapper.ResourceLookup.Set(id, res); err != nil {
		return err
	}
	doc := search.ResourceDoc{
		Name:    id.Name,
		Type:    id.Type.String(),
		Variant: id.Variant,
		Tags: 	 id.Tags,
	}
	return wrapper.Searcher.Upsert(doc)
}
```

metadata/metadata.go
```diff
@@ -229,11 +229,30 @@ func (wrapper SearchWrapper) Set(id ResourceID, res Resource) error {
...
+        allTags = append(allTags, featureVariant.serialized.Tags.Tag...)
+    }
+
+	if labelVariant, ok := res.(*labelVariantResource); ok {
+        allTags = append(allTags, labelVariant.serialized.Tags.Tag...)
+    }
+
+	if trainingSetVariant, ok := res.(*trainingSetVariantResource); ok {
+        allTags = append(allTags, trainingSetVariant.serialized.Tags.Tag...)
+    }
```
epps, line 248: This logic seems like it would be better expressed as a switch statement there will only ever be one match here:

```go
switch res.(type) {
    case *sourceVariantResource:
        allTags = sourceVariant.serialized.Tags.Tag
```

metadata/search/search.go
```diff
@@ -149,11 +152,21 @@ func (s Search) RunSearch(q string) ([]ResourceDoc, error) {
...
+				doc := hit.(map[string]interface{})
+
+		var tags []string
+		if tagSlice, ok := doc["Tags"].([]interface{}); ok {
+			for _, tag := range tagSlice {
+				if strTag, ok := tag.(string); ok {
+					tags = append(tags, strTag)
+				}
+			}
+		}
```
epps, line 164: I'm not certain but it looks like you might have some formatting issues in this file.

metadata/metadata.go
```diff
@@ -229,11 +229,30 @@ func (wrapper SearchWrapper) Set(id ResourceID, res Resource) error {
...
+	
+    doc := search.ResourceDoc{
+        Name:    id.Name,
+        Type:    id.Type.String(),
+		Tags:    allTags,
```
anthonylasso, line 253: looks like a format issue

anthonylasso:
This is great, thanks for the effort! Resolve the comments at will and we'll approve.

metadata/metadata.go
ihkap11: fixed here - commit c41c396943af6a33a6475dde4455410c1b4b9fb1

metadata/metadata.go
ihkap11: fixed here - commit c25645fa992f010c65d955ea66799b006099d574

metadata/search/search.go
ihkap11: Tried this instead- https://github.com/featureform/featureform/blob/c25645fa992f010c65d955ea66799b006099d574/metadata/search/search.go#L161

anthonylasso (approved):

anthonylasso:
Passing CI/CD in a separate branch. Will merge commit the unit test fix into this one. 

Screenshot 2024-01-12 at 2 05 47 PM

PR #1178, opened 2023-11-29
Task :Optional inference store for features

aolfat:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


coordinator/coordinator.go
```diff
@@ -744,6 +744,11 @@ func (c *Coordinator) runFeatureMaterializeJob(resID metadata.ResourceID, schedu
 	}(sourceStore)
 
 	featureProvider, err := feature.FetchProvider(c.Metadata, context.Background())
+	// remove this
```
ahmadnazeri, line 747: // remove this

ahmadnazeri (approved):

client/src/featureform/resources.py
```diff
@@ -1389,6 +1389,10 @@ def get(self, stub) -> "FeatureVariant":
     def _create(self, stub) -> Optional[str]:
         if hasattr(self.source, "name_variant"):
             self.source = self.source.name_variant()
+
+        # TODO: remove this hack
+        if self.provider == "local-mode":
```
ahmadnazeri, line 1394: can we not make it optional for local-mode?

coordinator/coordinator.go
```diff
@@ -743,7 +743,14 @@ func (c *Coordinator) runFeatureMaterializeJob(resID metadata.ResourceID, schedu
...
-	featureProvider, err := feature.FetchProvider(c.Metadata, context.Background())
+	var featureProvider *metadata.Provider
+	if feature.Provider() != "" {
+		featureProvider, err = feature.FetchProvider(c.Metadata, context.Background())
+		if err != nil {
+			return fmt.Errorf("could not fetch online provider: %v", err)
+		}
+	}
+
 	if err != nil {
```
ahmadnazeri, line 754: do we need to remove this check?

provider/provider_type/provider_type.go
```diff
@@ -31,6 +31,8 @@ const (
 	HDFS             Type = "HDFS"
 	AZURE            Type = "AZURE"
 	UNIT_TEST        Type = "UNIT_TEST"
+
```
ahmadnazeri, line 34: curious what is the reasoning behind the extra space?

aolfat: just to separate it all from the rest

coordinator/coordinator.go
aolfat: why would we need to remove it?

client/src/featureform/resources.py
aolfat: we have a really annoying line of code in register that says 

```
        if len(features) > 0 and inference_store == "":
            inference_store = "local-mode"
```

i got rid of that and it broke all local dash tests

ahmadnazeri: so would `inference_store == ""` for hosted if they don't want to use inference_store?

coordinator/coordinator.go
ahmadnazeri: you are performing the same check in the if conditional above?

client/src/featureform/resources.py
aolfat: it would, but how do you tell we're local or not? do we have a is_local mode flag?

PR #1181, opened 2023-11-30
Bug: Batch Feature Data Types

epps:
# Description

This feature introduces fixes to ensure batch features served by the client are the correct data types in the client. Additionally, it adds a materialization grouping by datetime to ensure the most recent data is served.

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/spark.go
```diff
@@ -716,7 +716,15 @@ func (store *SparkOfflineStore) GetBatchFeatures(ids []ResourceID) (BatchFeature
 	if err != nil {
 		return nil, fmt.Errorf("could not get output files: %v", err)
 	}
-	iterator, err := store.Store.Serve(outputFiles)
+	groups, err := filestore.NewFilePathGroup(outputFiles, filestore.DateTimeDirectoryGrouping)
```
sdreyer, line 719: Do we need this?

sdreyer: Since we only have one version

epps: We do need this given the UUID is deterministic given the name-variants and their order, so if we serve the same features in the same order, a new datetime directory will be written to `/featureform/BatchFeatures/<NAME>/<VARIANT>/`, so this grouping mechanism is key to avoid reading all parquet files from all materializations.

sdreyer (approved):

Issue #1160, opened 2023-11-16
Implement ClickHouse as an Offline Store[Feature Request]: 


kerekylytsiab:
### Description

Clickhouse is a fast open-source column-oriented database management system that allows generating analytical data reports in real-time using SQL queries. Implement Clickhouse as an Offline Store Provider in Featureform.

To learn more about how Offline Store providers in Featureform, visit [https://docs.featureform.com/providers/offline-store](https://docs.featureform.com/providers/offline-store)

### Desired Behavior

The ability to register ClickHouse as an Offline Store Provider in Featureform so that it can store primary data and function as a transformation engine.

### Benefits

With ClickHouse provider support, users can manage and version their transformations, features, and training sets with Featureform while leveraging ClickHouse.

### Possible Implementation

It should implement the Offline Store Interface: [https://github.com/featureform/featureform/blob/main/provider/offline.go#L253](https://github.com/featureform/featureform/blob/main/provider/offline.go#L253) 

Look at [https://github.com/featureform/featureform/blob/main/provider/postgres.go](https://github.com/featureform/featureform/blob/main/provider/postgres.go) for inspiration

PR #1224, opened 2023-12-15
Clickhouse offline store support

gingerwizard:
# Description

Support for ClickHouse as offline store. Closes https://github.com/featureform/featureform/issues/1160

## Type of change

This adds support for ClickHouse as an offline store. While CH could also be used as online store (and vector db) this is the main functional usecase.

I've made a few key design decisions/questions:

-  training sets, transformations and materializations are subject to update. We just rebuild these and use a standard table - materialized views in ClickHouse are a different concept. This means we can use a standard MergeTree table engine.
- For offline tables there appears to be a need to support updates i.e. `writeUpdate`, which according to sql.go base provider is always done by `entity` and `ts`. For this table we therefore use a ReplacingMergeTree - inserts with the same entity and ts will be de-duplicated (latest preserved) at insert and query time (final is set on all queries).
- For optimal query performance we have used the `entity, ts` as the ordering key on any offline tables. This key combination is also required to uniquely identify rows in our ReplacingMergeTree. The `ts` column can be null though - this means i need to use `allow_nullable_key` for the table - this is generally not recommended but probably ok for this usecase. 

Its not clear to me what a materialization is vs an offline table i.e. `OfflineTable`. The former i assume is when a feature is represented and "materialized" into the store. OfflineTable looks the same structurally - `entity`, `ts`, `value` columns. 

The above should be fine provided materializations cant recieve upserts like offline tables. If they are, we can make these tables a ReplacingMergeTree also.

A few other things:

- The methods `materializationCreate` and `transformationCreate` have been modified to return []string vs string. We can't do these in a single statement and don't support multiple statements in a single query. I've updated other offline stores and discussed with @ahmadnazeri 
 - the go client returns time.Time{} for the zero date. We coerce this to 1970. Unclear why the client does this but it seems intentional and i'm investigating.
 - We use DateTime64 with nanosecond precision. This might be unneccessary but its impossible to know what data we will get and complex to detect so we lean on the side of caution.
 - It might make sense to make entity LowCardinality(String) - without knowing common cardinalities its hard to say though and a premature optimization.
 - I've extended type support for uints, as well as int8 and int16. These are basic types and seem to have been ommitted but can significantly improve CH performance.

https://github.com/featureform/featureform/issues/1160

### Select type(s) of change

- [x] Documentation update

# Checklist:


client/mkdocs.yml
```diff
@@ -16,6 +16,10 @@ theme:
 extra_css:
   - stylesheets/extra.css
 
+extra:
+  version:
+    provider: mike
```
ahmadnazeri, line 21: any reason this is added?

health/health_test.go
```diff
@@ -175,6 +190,13 @@ func checkEnv(envVar string) string {
 	return value
 }
 
+func getEnv(key string, fallback string) string {
```
ahmadnazeri, line 193: You can leverage [GetEnv in helpers/env.go](https://github.com/featureform/featureform/blob/main/helpers/env.go#L11-L17)

health/health_test.go
```diff
@@ -284,6 +306,24 @@ func initProvider(t *testing.T, providerType pt.Type, executorType pc.SparkExecu
...
 		return postgresConfig.Serialize()
+	case pt.ClickHouseOffline:
+		db := checkEnv("CLICKHOUSE_DATABASE")
+		user := checkEnv("CLICKHOUSE_USER")
+		password := checkEnv("CLICKHOUSE_PASSWORD")
+		host := getEnv("CLICKHOUSE_HOST", "localhost")
+		port := uint64(9000)
+		port, _ = strconv.ParseUint(getEnv("CLICKHOUSE_PORT", "9000"), 10, 16)
+		ssl := false
+		ssl, _ = strconv.ParseBool(getEnv("CLICKHOUSE_SSL", "false"))
```
ahmadnazeri, line 317: There is a [GetEnvBool](https://github.com/featureform/featureform/blob/main/helpers/env.go#L47-L52) that you can leverage

health/health_test.go
```diff
@@ -284,6 +306,24 @@ func initProvider(t *testing.T, providerType pt.Type, executorType pc.SparkExecu
 			SSLMode:  "disable",
 		}
 		return postgresConfig.Serialize()
+	case pt.ClickHouseOffline:
+		db := checkEnv("CLICKHOUSE_DATABASE")
+		user := checkEnv("CLICKHOUSE_USER")
+		password := checkEnv("CLICKHOUSE_PASSWORD")
+		host := getEnv("CLICKHOUSE_HOST", "localhost")
+		port := uint64(9000)
+		port, _ = strconv.ParseUint(getEnv("CLICKHOUSE_PORT", "9000"), 10, 16)
```
ahmadnazeri, line 315: There is [GetEnvInt](https://github.com/featureform/featureform/blob/main/helpers/env.go#L33-L38) but we can add GetEnvUInt as well

provider/offline_test.go
```diff
@@ -60,6 +60,13 @@ func TestOfflineStores(t *testing.T) {
 		return value
 	}
 
+	getEnv := func(key, fallback string) string {
```
ahmadnazeri, line 63: This can be replaced by [GetEnv in helpers/env.go](https://github.com/featureform/featureform/blob/main/helpers/env.go#L11-L17)

provider/clickhouse.go
```diff
@@ -0,0 +1,1305 @@
...
+	}
+	db := clickhouse.OpenDB(&clickhouse.Options{
+		Addr: []string{fmt.Sprintf("%s:%d", cc.Host, cc.Port)},
+		Auth: clickhouse.Auth{
+			Database: cc.Database,
+			Username: cc.Username,
+			Password: cc.Password,
+		},
+		Settings: clickhouse.Settings{
+			"final": "1",
```
ahmadnazeri, line 233: would users want to make these configurable? also, any other settings that users might want to make configurable?

provider/clickhouse.go
```diff
@@ -0,0 +1,1305 @@
...
+		},
+		TLS: t,
+	})
+	queries := clickhouseSQLQueries{}
+	// numeric and should work
+	queries.setVariableBinding(PostgresBindingStyle)
```
ahmadnazeri, line 239: just want to make sure that this is the right value?

provider/clickhouse.go
```diff
@@ -0,0 +1,1305 @@
...
+		return nil, fmt.Errorf("could not return SQL offline table: %v", err)
+	}
+	return table, nil
+}
+
+func (store *clickHouseOfflineStore) GetResourceTable(id ResourceID) (OfflineTable, error) {
+	return store.getsqlResourceTable(id)
+}
+
+func (store *clickHouseOfflineStore) GetBatchFeatures(ids []ResourceID) (BatchFeatureIterator, error) {
```
ahmadnazeri, line 704: This can be implemented similar to [SQL GetBatchFeatures](https://github.com/featureform/featureform/blob/main/provider/sql.go#L581-L652)

provider/clickhouse.go
```diff
@@ -0,0 +1,1305 @@
...
+	query     OfflineTableQueries
+}
+
+func (mat *clickHouseMaterialization) ID() MaterializationID {
+	return mat.id
+}
+
+func (mat *clickHouseMaterialization) NumRows() (int64, error) {
+	var n interface{}
+	query := fmt.Sprintf("SELECT COUNT(*) FROM %s", sanitize(mat.tableName))
```
ahmadnazeri, line 1219: should this be `sanitizeCH()`?

provider/clickhouse.go
```diff
@@ -0,0 +1,1305 @@
...
+	intVar, err := mat.query.numRows(n)
+	if err != nil {
+		return 0, nil
+	}
+	return intVar, nil
+}
+
+func (mat *clickHouseMaterialization) IterateSegment(start, end int64) (FeatureIterator, error) {
+	query := mat.query.materializationIterateSegment(mat.tableName)
+	fmt.Println(query)
```
ahmadnazeri, line 1237: do we still need this?

ahmadnazeri (commented):
Looks good for the most part! Left some comments on the PR. There are two parts missing:
1. [End to End Behave Tests](https://github.com/featureform/featureform/tree/main/tests/end_to_end/features)
2. Having instant of Clickhouse available for CICD like we do for [Postgres](https://github.com/featureform/featureform/blob/main/.github/workflows/testing.yml#L265-273)

provider/clickhouse.go
gingerwizard: no this should be fixed. its for replacingmergetree and ignored for other engine types

provider/clickhouse.go
gingerwizard: Yes this works

provider/clickhouse.go
gingerwizard: id rather no expose other settings for now, lets see if they are needed

provider/clickhouse.go
gingerwizard: added

gingerwizard:
@ahmadnazeri i believe ive resolved all issues and added ci/cd tests + support for batch features. This should be ready to test.

anothserfile.txt
epps, line 1: Does this file serve a purpose I'm not seeing? If so, can we document this and/or change its name to reflect its purpose?

client/docs/register.md
```diff
@@ -93,6 +100,13 @@
         show_root_heading: false
         show_root_toc_entry: false
 
+### ClickHouse
+::: featureform.register.Registrar.register_clickhouse
+    handler: python
+    options:
+        show_root_heading: false
+        show_root_toc_entry: false
+
```
epps, line 109: Perhaps I'm missing something here, but I'm not seeing why there needs to be 2 entires for ClickHouse here.

anothserfile.txt
gingerwizard: no idea how that made it in, removing.

gingerwizard:
@epps addressed comments

ahmadnazeri (approved):

PR #1254, opened 2024-01-18
Removes all wrapped errors in filestore.go

epps:
# Description

Removes all wrapped errors in filestore.go.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/provider/filestore.go
+++ b/provider/filestore.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	s3v2 "github.com/aws/aws-sdk-go-v2/service/s3"
 	hdfs "github.com/colinmarc/hdfs/v2"
+	"github.com/featureform/fferr"
 	filestore "github.com/featureform/filestore"
 	pc "github.com/featureform/provider/provider_config"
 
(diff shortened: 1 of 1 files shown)


provider/filestore.go
```diff
@@ -581,23 +588,23 @@ func (fs *HDFSFileStore) ServeDirectory(files []filestore.Filepath) (Iterator, e
...
 	}
 	switch file.Ext() {
 	case filestore.Parquet:
 		return parquetIteratorFromBytes(b)
 	case filestore.CSV:
-		return nil, fmt.Errorf("could not find CSV reader")
+		return nil, fferr.NewInternalError(fmt.Errorf("CSV reader not implemented"))
 	default:
-		return nil, fmt.Errorf("unsupported file type")
+		return nil, fferr.NewDataTypeNotFoundError(string(file.Ext()), fmt.Errorf("unsupported file type"))
```
sdreyer, line 607: This should be something different since its not a datatype error

provider/filestore.go
```diff
@@ -606,25 +613,28 @@ func (fs *HDFSFileStore) Exists(path filestore.Filepath) (bool, error) {
 	if err != nil && strings.Contains(err.Error(), "file does not exist") {
 		return false, nil
 	} else if err != nil {
-		return false, err
+		return false, fferr.NewInternalError(err)
```
sdreyer, line 616: Thinking about this one. Like if its a connection error or something

provider/filestore.go
epps: Invalid arg or internal?

epps: On the one hand, this could be an invalid arg if a user supplies a file and then directly tries to server it; otherwise, it seems like an internal error if Spark generates something other than parquet.

sdreyer: Can this be both something that we set and the user set? We may want an InvalidFileFormat error

sdreyer (approved):

PR #1266, opened 2024-01-22
Removes all provider-specific errors

epps:
# Description

Removes all provider-specific errors.

### Select type(s) of change

- [ ] Documentation update

# Checklist:


health/health_test.go
```diff
@@ -233,7 +233,7 @@ func initSpark(t *testing.T, executorType pc.SparkExecutorType, storeType fs.Fil
 		}
 	case fs.GCS:
 		credsFile := os.Getenv("GCP_CREDENTIALS_FILE")
-		content, err := ioutil.ReadFile(credsFile)
+		content, err := os.ReadFile(credsFile)
```
sdreyer, line 236: Any particular reason for this?

provider/clickhouse.go
```diff
@@ -101,15 +102,15 @@ func (store *clickHouseOfflineStore) createTransformationName(id ResourceID) (st
 	case Transformation:
 		return GetPrimaryTableName(id)
 	case Label:
-		return "", TransformationTypeError{"Invalid Transformation Type: Label"}
+		return "", fferr.NewInvalidArgument(fmt.Errorf("invalid transformation type: Label"))
```
sdreyer, line 105: Don't we have an InvalidResourceType error?

provider/dynamodb.go
```diff
@@ -281,12 +281,16 @@ func (store *dynamodbOnlineStore) CheckHealth() (bool, error) {
 		return false, fferr.NewExecutionError(pt.DynamoDBOnline.String(), err)
 	}
 	if len(listOutput.TableNames) == 0 {
-		return false, NewProviderError(Connection, pt.DynamoDBOnline, Ping, "no tables found")
+		wrapped := fferr.NewConnectionError(pt.DynamoDBOnline.String(), fmt.Errorf("no tables found"))
```
sdreyer, line 284: Little confusing. Why does 0 tables mean it was a connection error?

provider/dynamodb.go
```diff
@@ -281,12 +281,16 @@ func (store *dynamodbOnlineStore) CheckHealth() (bool, error) {
...
+		wrapped.AddDetail("action", "ping")
+		return false, wrapped
 	}
 	scanInput := &dynamodb.ScanInput{TableName: listOutput.TableNames[0], Limit: aws.Int64(1)}
 	_, err = store.client.Scan(scanInput)
 	if err != nil {
-		return false, NewProviderError(Connection, pt.DynamoDBOnline, Ping, err.Error())
+		wrapped := fferr.NewConnectionError(pt.DynamoDBOnline.String(), fmt.Errorf("no tables found"))
```
sdreyer, line 291: Feel like this is also a misdirection. Shouldnt we just check to see if the query failed on the ListTables?

health/health_test.go
epps: My Golang linter pointed out that `ioutil.ReadFile` was deprecated in 1.16 and now just calls `os.ReadFile`, so I figured I'd remove instances as I encountered them.

provider/dynamodb.go
epps: As discussed, we'll change this check to see if the user has the correct permissions and remove the list tables check.

provider/clickhouse.go
epps: Good call. I should use this instead. Will correct this in `sql.go`, too.

sdreyer (approved):

PR #1331, opened 2024-02-13
merge taskrun with task

RiddhiBagadiaa:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


scheduling/task_run.go
```diff
@@ -0,0 +1,149 @@
...
+	triggerMap := make(map[string]interface{})
+	if err := json.Unmarshal(temp.Trigger, &triggerMap); err != nil {
+		return fmt.Errorf("failed to deserialize trigger data: %w", err)
+	}
+
+	if _, ok := triggerMap["triggerType"]; !ok {
+		return fmt.Errorf("target type is missing")
+	}
+
+	if triggerMap["triggerType"] == "OneOffTrigger" {
```
sdreyer, line 133: Something we should do that i realized from the other one is use the enum here instead of a single string and make it a switch rather than an if

If we can change this for tasks defs too thatd be good

PR #1335, opened 2024-02-14
add etcd provider, made new function to create new provider(memory an…

RiddhiBagadiaa:
…d etcd)

# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


scheduling/storage_provider.go
```diff
@@ -52,6 +56,56 @@ func (m *MemoryStorageProvider) Get(key string, prefix bool) ([]string, error) {
...
+func NewETCDStorageProvider(client *clientv3.Client, ctx context.Context) *ETCDStorageProvider {
+	return &ETCDStorageProvider{client: client, ctx: ctx}
+}
+
+func (etcd *ETCDStorageProvider) Set(key string, value string) error {
+	if key == "" {
+		return fmt.Errorf("key cannot be empty")
+	}
+	if value == "" {
+		return fmt.Errorf("value cannot be empty")
```
anthonylasso, line 73: small suggestion. might be useful to know which key was trying to set the blank value (the key should be something after passing the first check)  "value cannot be empty for key: <key>" or something.

scheduling/storage_providers/storage_provider.go
```diff
@@ -0,0 +1,18 @@
+package scheduling
+
+import (
+	"fmt"
+)
+
+type StorageProvider interface {
```
aolfat, line 7: does this need to be in scheduling?

scheduling/storage_provider.go
anthonylasso: so let's say that `set` is somehow invoked with ("myKey", "") 
the first if block will pass since `key` has a value, then the 2nd if block will return the error "value cannot be empty" but with no mention of which key. but if the logs already reference which key that error refers to then resolve this convo at will!

RiddhiBagadiaa: Makes sense, will fix that

scheduling/storage_providers/storage_provider.go
RiddhiBagadiaa: Apparently yes

scheduling/storage_providers/etcd_test.go
```diff
@@ -0,0 +1,49 @@
...
+	etcdConfig := clientv3.Config{
+		Endpoints:   []string{address},
+		DialTimeout: time.Second * 10,
+		Username:    etcdUsername,
+		Password:    etcdPassword,
+	}
+
+	client, err := clientv3.New(etcdConfig)
+	if err != nil {
+		panic(err)
```
ahmadnazeri, line 39: why panic?

ahmadnazeri (approved):

anthonylasso (approved):

PR #1359, opened 2024-02-28
Chore: Errors follow-up client clean up and refactor

aolfat:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/fferr/errors.go
+++ b/fferr/errors.go
@@ -76,6 +76,7 @@ type Error interface {
 	GRPCStatus() *status.Status
 	ToErr() error
 	AddDetail(key, value string)
+	AddDetails(keysAndValues ...interface{})
 	Error() string
 	Stack() JSONStackTrace
 }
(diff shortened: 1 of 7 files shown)


fferr/errors.go
```diff
@@ -76,6 +76,7 @@ type Error interface {
 	GRPCStatus() *status.Status
 	ToErr() error
 	AddDetail(key, value string)
+	AddDetails(keysAndValues ...interface{})
```
epps, line 79: Do we really want to allow values of any type here? Why not limit it to strings to keep things simple?

aolfat: It's just the interface method to do it. I modeled this after how logger.Infow works. The validation is done at the impl level

fferr/generic.go
```diff
@@ -44,6 +44,32 @@ func (e *GenericError) AddDetail(key, value string) {
...
+	}
+
+	for i := 0; i < n; i += 2 {
+		key, ok := keysAndValues[i].(string)
+		if !ok {
+			// Handle or log the error if the key is not a string.
+			fmt.Printf("Key at index %d is not a string\n", i)
+			continue
+		}
+		// Convert the value to a string, this assumes that the value has a meaningful string representation
```
epps, line 67: I don't think we should be making this assumption. I think we should force the consumer to use a string value so it's clear to them that without a meaningful string representation, they can't add details.

epps (approved):
👍

ahmadnazeri (approved):
LGTM, agree with Erik that we should expect it to be strings

PR #1363, opened 2024-03-01
Adding MemoryStorerImplementation 

ahmadnazeri:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


locker/locking.go
```diff
@@ -0,0 +1,73 @@
...
+
+type Key interface {
+	ID() string
+	Key() string
+	ExpirationTime() time.Time
+	SetExpirationTime(time.Time) error // CHECK: should this be here?
+}
+
+// A method used to close a channel only once
+func closeOnce(ch chan error) {
```
anthonylasso, line 68: nit: instead of the error type, maybe an empty struct{}. I feel like seeing "error" is throwing me off.

locker/memory_locker.go
```diff
@@ -0,0 +1,125 @@
...
+		Date: time.Now().UTC(),
+	}
+	m.lockedItems.Store(key, lock)
+
+	go m.updateLockTime(id, key, lockChannel)
+
+	return &lockKey, nil
+}
+
+func (m *MemoryLocker) updateLockTime(id string, key string, lockChannel chan error) {
```
anthonylasso, line 72: nit: use the "arrow" syntax to indicate this only reads from the channel

i.e. `lockChannel <-chan error`

locker/memory_locker.go
```diff
@@ -0,0 +1,125 @@
...
+	defer m.mutex.Unlock()
+
+	if lockInfo, ok := m.lockedItems.Load(key); ok {
+		keyLock := lockInfo.(LockInformation)
+		if time.Since(keyLock.Date) < ValidTimePeriod {
+			return &MemoryKey{}, fmt.Errorf("key '%s' is already locked by: %s", key, keyLock.ID)
+		}
+	}
+
+	lockChannel := make(chan error)
```
anthonylasso, line 57: looks like the `done` pattern, I'd change the name to `doneChannel` or just `done` for clarity.

locker/memory_locker.go
```diff
@@ -0,0 +1,125 @@
...
+	if key.Key() == "" {
+		return fmt.Errorf("key is empty")
+	}
+
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	lockInfo, ok := m.lockedItems.Load(key.Key())
+	if !ok {
+		return fmt.Errorf("key '%s' is not locked", key.Key())
```
anthonylasso, line 113: nice. should help with debugging. hopefully we won't need it 🤞

locker/locking_test.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	// Release the lock
+	err = locker.Unlock(lock)
+	if err != nil {
+		t.Fatalf("Unlock failed: %v", err)
+	}
+}
+
+func StressTestLockAndUnlock(t *testing.T, locker MultiLock) {
```
anthonylasso, line 121: good test btw

storage_storer/memory_storer.go
```diff
@@ -0,0 +1,60 @@
...
+type MemoryStorerImplementation struct {
+	storage map[string]string
+}
+
+func (m *MemoryStorerImplementation) Set(key string, value string) error {
+	if key == "" {
+		return fmt.Errorf("key is empty")
+	}
+
+	m.storage[key] = value
```
anthonylasso, line 17: do we care if the value is zeroed?

storage_storer/memory_storer.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (m *MemoryStorerImplementation) Get(key string) (string, error) {
+	if key == "" {
+		return "", fmt.Errorf("key is empty")
+	}
+
+	value, ok := m.storage[key]
+	if !ok {
+		return "", fmt.Errorf("key not found")
```
anthonylasso, line 29: nit: ref the key

storage_storer/memory_storer.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (m *MemoryStorerImplementation) Delete(key string) (string, error) {
+	if key == "" {
+		return "", fmt.Errorf("key is empty")
+	}
+
+	value, ok := m.storage[key]
+	if !ok {
+		return "", fmt.Errorf("key not found")
```
anthonylasso, line 54: nit: same

storage_storer/metadata_storer.go
```diff
@@ -0,0 +1,67 @@
...
+	newValue, err := updateFn(currentValue)
+	if err != nil {
+		return err
+	}
+
+	return s.Storer.Set(key, newValue)
+}
+
+func (s *MetadataStorer) List(prefix string) (map[string]string, error) {
+	// TODO: how do we lock a prefix?
```
anthonylasso, line 39: good question

anthonylasso: thinking out loud, we could create another map to track that state, but that feels like a lot of boiler plate.

locker/locking.go
```diff
@@ -0,0 +1,73 @@
...
+type MultiLock interface {
+	Lock(lock string) (Key, error)
+	Unlock(key Key) error
+}
+
+type Key interface {
+	ID() string
+	Key() string
+	ExpirationTime() time.Time
+	SetExpirationTime(time.Time) error // CHECK: should this be here?
```
simba-git, line 64: These are unused

locker/memory_locker.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	keyLock := lockInfo.(LockInformation)
+	if keyLock.ID != key.ID() {
+		return fmt.Errorf("key '%s' is locked by another id: locked by: %s, unlock  by: %s", key.Key(), keyLock.ID, key.ID())
+	}
+	m.LockedItems.Delete(key.Key())
+	memoryKey := key.(*MemoryKey)
+	closeOnce(*memoryKey.Channel)
```
simba-git, line 122: Don't understand why we need this close once. If we own the mutex, it should be impossible for it to happen twice?

locker/memory_locker.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (k *MemoryKey) ID() string {
+	return k.id
+}
+
+func (k *MemoryKey) Key() string {
+	return k.key
+}
+
+func (k *MemoryKey) ExpirationTime() time.Time {
```
simba-git, line 26: Dont think we need these, I just recommend making ExperationTime public

locker/memory_locker.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type MemoryKey struct {
```
simba-git, line 11: MemoryKey should be private: memoryKey

storage_storer/metadata_storer.go
simba-git: Could also make the locker interface prefix aware, or just lock the entire thing when listing

locker/locking.go
ahmadnazeri: removing it since it was from previous interface design

locker/locking.go
ahmadnazeri: removed

locker/memory_locker.go
ahmadnazeri: changed it to `close(mKey.Channel)`. it was just from previous implementation

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+		return TaskMetadata{}, fmt.Errorf("failed to create task metadata: %v", err)
+	}
+
+	runs := TaskRuns{
+		TaskID: metadata.ID,
+		Runs:   []TaskRunSimple{},
+	}
+	serializedRuns, err := runs.Marshal()
+	if err != nil {
+		return TaskMetadata{}, err
```
anthonylasso, line 105: should return an `fmt.Errorf` for clarity

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+
+func (m *TaskMetadataManager) CreateTaskRun(name string, taskID TaskID, trigger Trigger) (TaskRunMetadata, error) {
+	// ids will be generated by TM
+	taskRunKey := TaskRunKey{taskID: taskID}
+	taskMetadata, err := m.storer.Get(taskRunKey.String())
+	if err != nil {
+		return TaskRunMetadata{}, fmt.Errorf("failed to fetch task: %v", err)
+	}
+
+	// Not sold on this naming for this struct. Maybe like RunHistory or something?
```
anthonylasso, line 164: doesn't roll of the tongue but I understood the meaning.

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+	runs := TaskRuns{}
+	err = runs.Unmarshal([]byte(taskMetadata))
+	if err != nil {
+		return TaskRunMetadata{}, err
+	}
+
+	// This function could be a method of TaskRuns
```
anthonylasso, line 171: agree, should be method since it relates only to TRs

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+	serializedRuns, err := runs.Marshal()
+	if err != nil {
+		return TaskRunMetadata{}, err
+	}
+
+	serializedMetadata, err := metadata.Marshal()
+	if err != nil {
+		return TaskRunMetadata{}, fmt.Errorf("failed to marshal metadata: %v", err)
+	}
+	err = m.storer.Create(taskRunKey.String(), string(serializedRuns))
```
anthonylasso, line 200: do something with this error

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+
+	taskRunMetaKey := TaskRunMetadataKey{taskID: taskID, runID: metadata.ID, date: startTime}
+	err = m.storer.Create(taskRunMetaKey.String(), string(serializedMetadata))
```
anthonylasso, line 203: do something with this error x2

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+			return []TaskRunMetadata{}, err
+		}
+		runs = append(runs, dayRuns...)
+	}
+
+	return runs, nil
+}
+
+func (m *TaskMetadataManager) getRunsForDay(date time.Time, start time.Time, end time.Time) ([]TaskRunMetadata, error) {
+	// TODO: question, should this return an error if we can't marshal
```
anthonylasso, line 284: imo we should throw an error. a faulty record could cause chaos downstream.

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+
+		return string(serializedMetadata), nil
+	}
+
+	taskRunMetadataKey := TaskRunMetadataKey{taskID: taskID, runID: metadata.ID, date: metadata.StartTime}
+	e = m.storer.Update(taskRunMetadataKey.String(), updateLog)
+	return e
+}
+
+// Finds the highest increment in a list of strings formatted like "/tasks/metadata/task_id=0"
```
anthonylasso, line 443: could probably put this in `TaskMetadata` or maybe create a helpers directory for these functions

locker/memory_locker.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+type MemoryLocker struct {
+	LockedItems sync.Map
+	Mutex       *sync.Mutex
+}
+
+func (m *MemoryLocker) Lock(key string) (Key, fferr.GRPCError) {
+	if key == "" {
+		return &memoryKey{}, fferr.NewInternalError(fmt.Errorf("key is empty"))
```
sdreyer, line 34: Should make the message more specific or make a custom error. Going to be confusing if we just get a message saying key is empty

locker/memory_locker.go
```diff
@@ -0,0 +1,120 @@
...
+					Date: time.Now().UTC(),
+				})
+			}
+		}
+	}
+}
+
+func (m *MemoryLocker) Unlock(key Key) fferr.GRPCError {
+	if key.Key() == "" {
+		return fferr.NewInternalError(fmt.Errorf("key is empty"))
```
sdreyer, line 98: Same treatment for this error

locker/memory_locker.go
```diff
@@ -0,0 +1,120 @@
...
+
+	lockInfo, ok := m.LockedItems.Load(key.Key())
+	if !ok {
+		return fferr.NewKeyNotLockedError(key.Key(), nil)
+	}
+
+	keyLock := lockInfo.(LockInformation)
+	if keyLock.ID != key.ID() {
+		err := fferr.NewKeyAlreadyLockedError(key.Key(), keyLock.ID, nil)
+		err.AddDetail("trying to unlock key with different id", key.ID())
```
sdreyer, line 112: better format for this would be:
```
err := fferr.NewKeyAlreadyLockedError(key.Key(), keyLock.ID, fmt.Errorf("attempting to unlock with incorrect key"))
err.AddDetail("expected key", keyLock.ID())
err.AddDetail("recieved key", key.ID())

File: scheduling/taskmanager_test.go [hunk 24]
```diff
@@ -14,923 +11,923 @@ func TestInitialization(t *testing.T) {
...
-
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.Name, func(t *testing.T) {
-			fn(t, tt)
-		})
-	}
-}
+// func TestCreateTask(t *testing.T) {
```
sdreyer, line 14: uncomment?

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,463 @@
...
+	"time"
+
+	"github.com/featureform/locker"
+	ss "github.com/featureform/storage_storer"
+)
+
+type TaskMetadataList []TaskMetadata
+
+func (tml *TaskMetadataList) ToJSON() string {
+	return ""
```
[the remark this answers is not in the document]
sdreyer: can remove

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,550 @@
...
+	if err != nil {
+		err.AddDetail("task_id", string(taskID))
+		err.AddDetail("run_id", string(runID))
+		return TaskRunMetadata{}, err
+	}
+
+	// Want to move this logic out
+	found := false
+	var runRecord TaskRunSimple
+	for _, run := range runs.Runs {
```
sdreyer, line 271: Throw this in a func

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,550 @@
...
+
+func getHighestRunID(taskRuns TaskRuns) (TaskRunID, fferr.GRPCError) {
+	if len(taskRuns.Runs) == 0 {
+		return 0, nil
+	}
+
+	// Move this logic out
+	highestRunID := taskRuns.Runs[0].RunID
+
+	for _, run := range taskRuns.Runs[1:] {
```
sdreyer, line 242: throw this in a func

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,550 @@
...
+
+func (m *TaskMetadataManager) GetAllTasks() (TaskMetadataList, fferr.GRPCError) {
+	metadata, err := m.storer.List(TaskMetadataKey{}.String())
+	if err != nil {
+		return TaskMetadataList{}, err
+	}
+
+	// Want to move this logic out of this func
+	tml := TaskMetadataList{}
+	for _, m := range metadata {
```
sdreyer, line 149: throw this in a func

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,550 @@
...
+	metadata, e := m.GetRunByID(taskID, runID)
+	if e != nil {
+		return e
+	}
+
+	updateLog := func(runMetadata string) (string, fferr.GRPCError) {
+		metadata := TaskRunMetadata{}
+		err := metadata.Unmarshal([]byte(runMetadata))
+		if err != nil {
+			return "", err
```
sdreyer, line 502: is this supposed to have a structured error?

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,550 @@
...
+	taskRunMetadataKey := TaskRunMetadataKey{taskID: taskID, runID: metadata.ID, date: metadata.StartTime}
+	e = m.storer.Update(taskRunMetadataKey.String(), updateLog)
+	return e
+}
+
+// Finds the highest increment in a list of strings formatted like "/tasks/metadata/task_id=0"
+func getLatestId(taskMetadata map[string]string) (int, fferr.GRPCError) {
+	highestIncrement := -1
+	for path, _ := range taskMetadata {
+		parts := strings.Split(path, "task_id=")
```
sdreyer, line 524: parsing should have its own function

sdreyer: Should see if we can change this to regex

scheduling/taskmetadata_manager_test.go
```diff
@@ -0,0 +1,895 @@
...
+	"testing"
+	"time"
+
+	sp "github.com/featureform/scheduling/storage_providers"
+)
+
+func TestTaskMetadataManager(t *testing.T) {
+	testFns := map[string]func(*testing.T, TaskMetadataManager){
+		"CreateTask": testCreateTask,
+		// "GetTaskMetadata":  TaskMetadataGet,
```
sdreyer, line 15: Should be enabled?

storage_storer/metadata_storer.go
sdreyer, line 1: any reason why this package is storage_storer instead of just storage?

storage_storer/metadata_storer.go
```diff
@@ -0,0 +1,82 @@
...
+	newValue, err := updateFn(currentValue)
+	if err != nil {
+		return err
+	}
+
+	return s.Storer.Set(key, newValue)
+}
+
+func (s *MetadataStorer) List(prefix string) (map[string]string, fferr.GRPCError) {
+	lock, err := s.Locker.Lock(prefix)
```
sdreyer, line 44: Does the locker understand prefixes? Should make a test for it

scheduling/taskmetadata.go
ahmadnazeri: Marshal returns a featureform error

scheduling/taskmetadata.go
ahmadnazeri: they will return fferr types

scheduling/taskmetadata_manager_test.go
ahmadnazeri: I have the other tests directly for Memory, will change them when I add ETCD.

locker/locking.go
```diff
@@ -0,0 +1,70 @@
...
+		return fferr.NewInvalidArgumentError(err)
+	}
+
+	l.ID = tmp.ID
+	l.Key = tmp.Key
+
+	// Parse the date string with UTC time zone
+	parsedTime, err := time.Parse(time.RFC3339, tmp.Date)
+	if err != nil {
+		return fferr.NewInternalError(err)
```
anthonylasso, line 47: nit: I'd add context to the error like `id` and `key` errors above using addDetail
```
parseErr := fferr.NewInternalError(err)
parseErr.AddDetail("parseDate", "lock information cannot parse date field")
```
the regular parse error already has the attempted info, so just adding lock-specific context.

anthonylasso: maybe consider creating a new type. NewParsingError or something.

locker/memory_locker.go
```diff
@@ -0,0 +1,153 @@
...
+func (m *MemoryLocker) Lock(key string) (Key, fferr.GRPCError) {
+	if key == "" {
+		return &memoryKey{}, fferr.NewInternalError(fmt.Errorf("cannot lock an empty key"))
+	}
+
+	m.Mutex.Lock()
+	defer m.Mutex.Unlock()
+
+	if m.isPrefixOfExistingKey(key) || m.hasPrefixLocked(key) {
+		return &memoryKey{}, fferr.NewKeyAlreadyLockedError(key, "", nil)
```
anthonylasso, line 41: passing in an empty string would result in a confusing log for `NewKeyAlreadyLockedError`
"key SomeKey is already locked by: "

Since this is prefix specific, maybe create a new error type or modify the existing so it reads:
"key SomeKey prefix is already locked"

locker/memory_locker.go
```diff
@@ -0,0 +1,153 @@
...
+	"time"
+
+	"github.com/featureform/fferr"
+	"github.com/google/uuid"
+)
+
+type memoryKey struct {
+	id      string
+	key     string
+	Channel *chan error
```
anthonylasso, line 16: nit: identifier smell. rename to `Done`? otherwise reads like it's meant to hold an error

locker/memory_locker.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	keyLock, ok := lockInfo.(LockInformation)
+	if !ok {
+		return fferr.NewInternalError(fmt.Errorf("could not cast lock information"))
+	}
+	if keyLock.ID != key.ID() {
+		err := fferr.NewKeyAlreadyLockedError(key.Key(), keyLock.ID, fmt.Errorf("attempting to unlock with incorrect key"))
+		err.AddDetail("expected key", keyLock.ID)
+		err.AddDetail("recieved key", key.ID())
```
anthonylasso, line 142: typo: "received"

scheduling/taskmanager.go
```diff
@@ -238,7 +216,7 @@ func (tm *TaskManager) CreateTaskRun(name string, taskID TaskID, trigger Trigger
 	}
 
 	// This function could be a method of TaskRuns
```
anthonylasso, line 218: old comment now, should delete

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+func (m *TaskMetadataManager) CreateTask(name string, tType TaskType, target TaskTarget) (TaskMetadata, fferr.GRPCError) {
+	keys, err := m.storage.List(TaskMetadataKey{}.String())
+	if err != nil {
+		err.AddDetail("metadata type", "task metadata")
+		err.AddDetail("task name", name)
+		err.AddDetail("task type", string(tType))
+		return TaskMetadata{}, err
+	}
+
+	// This logic could probably be somewhere else
```
anthonylasso, line 61: could do the same as TaskRuns with getLatestId()

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+// Finds the highest increment in a list of strings formatted like "/tasks/metadata/task_id=0"
+func getLatestId(taskMetadata map[string]string) (int, fferr.GRPCError) {
+	highestTaskId := -1
+
+	// Regular expression pattern to match "task_id=<number>"
+	pattern := regexp.MustCompile(`/tasks/metadata/task_id=(\d+)`)
+
+	for path, _ := range taskMetadata {
```
anthonylasso, line 497: you can shorthand this if you only need the path

`for path := range taskMetadata`

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	return taskRun, nil
+}
+
+func (m *TaskMetadataManager) GetRunsByDate(start time.Time, end time.Time) (TaskRunList, fferr.GRPCError) {
+	// the date range is inclusive
+	var runs []TaskRunMetadata
+
+	// iterate through each day in the date range including the end date
+	for date := start; date.Before(end) || date.Equal(end); date = date.AddDate(0, 0, 1) {
+		dayRuns, err := m.getRunsForDay(date, start, end)
```
simba-git, line 283: Why do we do it like this?

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+func (m *TaskMetadataManager) GetAllTaskRuns() (TaskRunList, fferr.GRPCError) {
+	recs, err := m.storage.List(TaskRunMetadataKey{}.String())
+	if err != nil {
+		return []TaskRunMetadata{}, err
+	}
+
+	var runs []TaskRunMetadata
+	for _, record := range recs {
+		taskRun := TaskRunMetadata{}
+		err = taskRun.Unmarshal([]byte(record))
```
simba-git, line 327: In a lot of situations we are throwing away a lot of these, but are parsing them each time. It's not a problem but maybe leave a comment about it. In case it becomes an issue later.

scheduling/task_run.go
```diff
@@ -67,10 +69,14 @@ type TaskRunMetadata struct {
 }
 
 // Formatting
```
aolfat, line 71: what's this comment?

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+		if err != nil {
+			return []TaskRunMetadata{}, err
+		}
+		runs = append(runs, dayRuns...)
+	}
+
+	return runs, nil
+}
+
+func (m *TaskMetadataManager) getRunsForDay(date time.Time, start time.Time, end time.Time) ([]TaskRunMetadata, fferr.GRPCError) {
```
aolfat, line 293: What is this for?

- How is a day defined? utc? what about timezones?

scheduling/taskmetadata.go
aolfat: assuming we need a "day" function I image the right logic would be that the day uses the get runs by date and just truncates the times. Either way the day function confuses me a bit

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+		err.AddDetail("task name", name)
+		err.AddDetail("task_id", string(taskID))
+		err.AddDetail("run_id", string(metadata.ID))
+		return TaskRunMetadata{}, err
+	}
+
+	return metadata, nil
+}
+
+func (m *TaskMetadataManager) GetRunByID(taskID TaskID, runID TaskRunID) (TaskRunMetadata, fferr.GRPCError) {
```
aolfat, line 232: aren't all runs associated to tasks? I imagine you should be able to just pass in a run in that case, what if you you pass in a run that isn't associated with that task?

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+func (m *TaskMetadataManager) CreateTaskRun(name string, taskID TaskID, trigger Trigger) (TaskRunMetadata, fferr.GRPCError) {
+	// ids will be generated by TM
+	taskRunKey := TaskRunKey{taskID: taskID}
+	taskMetadata, err := m.storage.Get(taskRunKey.String())
+	if err != nil {
+		err.AddDetail("task name", name)
+		return TaskRunMetadata{}, err
+	}
+
+	// Not sold on this naming for this struct. Maybe like RunHistory or something?
```
aolfat, line 168: I like runhistory but the implication with history is that it's usually sorted by timei

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	if err != nil {
+		err.AddDetail("task name", name)
+		err.AddDetail("task_id", string(taskID))
+		return TaskRunMetadata{}, err
+	}
+
+	startTime := time.Now().UTC()
+
+	metadata := TaskRunMetadata{
+		ID:          TaskRunID(latestID + 1),
```
aolfat, line 187: do we not need locks for all this? it's not possible for two taskrunmetadata with the same id to be generated/

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	}
+	return TaskMetadataManager{
+		storage: memoryMetadataStorage,
+	}
+}
+
+func (m *TaskMetadataManager) CreateTask(name string, tType TaskType, target TaskTarget) (TaskMetadata, fferr.GRPCError) {
+	keys, err := m.storage.List(TaskMetadataKey{}.String())
+	if err != nil {
+		err.AddDetail("metadata type", "task metadata")
```
aolfat, line 55: how we create errors is v clunky rn 3 lines to create an error, we should have a err.WithDetails(...) similar to how a logger does it. Don't worry about it on this PR though

locker/locking.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (l *LockInformation) Marshal() ([]byte, fferr.GRPCError) {
+	bytes, err := json.Marshal(l)
+	if err != nil {
+		return nil, fferr.NewInternalError(err)
+	}
+	return bytes, nil
+}
+
+type MultiLock interface {
```
aolfat, line 62: can you add a comment on what a multi lock is?

locker/memory_locker.go
aolfat: nit: channel lowercase

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+		Storage: make(map[string]string),
+	}
+
+	memoryMetadataStorage := ss.MetadataStorage{
+		Locker:  &memoryLocker,
+		Storage: &memoryStorage,
+	}
+	return TaskMetadataManager{
+		storage: memoryMetadataStorage,
+	}
```
aolfat, line 49: what if it's

```
    return TaskMetadataManager{
        storage: ss.NewMemoryMetadataStorage(),
    }
```

and you delegate the creation

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	taskRunMetaKey := TaskRunMetadataKey{taskID: taskID, runID: metadata.ID, date: startTime}
+	err = m.storage.Create(taskRunMetaKey.String(), string(serializedMetadata))
+	if err != nil {
+		err.AddDetail("task name", name)
+		err.AddDetail("task_id", string(taskID))
+		err.AddDetail("run_id", string(metadata.ID))
+		return TaskRunMetadata{}, err
+	}
+
+	return metadata, nil
```
aolfat, line 229: nit: this should probably be it's own

```
    if err := m.saveTaskMetadata(metadata); err != nil {
        return TaskMetadata{}, fmt.Errorf("saving task metadata failed: %w", err)
    }
```

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+		err.AddDetail("task type", string(tType))
+		return TaskMetadata{}, err
+	}
+
+	// This logic could probably be somewhere else
+	var latestID int
+	if len(keys) == 0 {
+		latestID = 0
+	} else {
+		latestID, err = getLatestId(keys)
```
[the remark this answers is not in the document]
aolfat: yup was going to say the same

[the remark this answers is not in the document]
aolfat: I think whatever is storing the keys should probs be in charge of generating these latestIds

scheduling/tasks.go
```diff
@@ -53,11 +55,16 @@ type TaskMetadata struct {
 	DateCreated time.Time  `json:"dateCreated"`
 }
 
-func (t *TaskMetadata) Marshal() ([]byte, error) {
-	return json.Marshal(t)
+func (t *TaskMetadata) Marshal() ([]byte, fferr.GRPCError) {
```
aolfat, line 58: let's use the same receiver name for all these.  goland gives me warnings on these. Does vscode not have a go extension that does static checks/lints on go?

scheduling/tasks.go
```diff
@@ -70,50 +77,61 @@ func (t *TaskMetadata) Unmarshal(data []byte) error {
...
+		return fferr.NewInvalidArgumentError(fmt.Errorf("task metadata is missing name"))
 	}
 	t.Name = temp.Name
 
-	if temp.TaskType == "" || (temp.TaskType != ResourceCreation && temp.TaskType != HealthCheck && temp.TaskType != Monitoring) {
-		return fmt.Errorf("unknown task type: %s", temp.TaskType)
+	if temp.TaskType == "" {
+		return fferr.NewInvalidArgumentError(fmt.Errorf("task metadata is missing TaskType"))
+	}
+	if temp.TaskType != ResourceCreation || temp.TaskType != HealthCheck || temp.TaskType != Monitoring {
```
aolfat, line 96: this always evaluates to true

aolfat: i think you want something like this

```
func isValidTaskType(tType TaskType) bool {
    switch tType {
    case ResourceCreation, HealthCheck, Monitoring:
        return true
    default:
        return false
    }
}
```

aolfat: goland points these little things out for me. Does vscode not do it? If not you should consider switching lol

scheduling/tasks.go
```diff
@@ -53,11 +55,16 @@ type TaskMetadata struct {
...
+	bytes, err := json.Marshal(t)
+	if err != nil {
+		errMessage := fmt.Errorf("failed to serialize task metadata: %w", err)
+		return nil, fferr.NewInternalError(errMessage)
+	}
+	return bytes, nil
 }
 
-func (t *TaskMetadata) Unmarshal(data []byte) error {
+func (t *TaskMetadata) Unmarshal(data []byte) fferr.GRPCError {
```
aolfat, line 67: we don't need to specifically return a GRPCError everywhere, we can just return a regular error

aolfat: since we wrap it anyways we can rely on that

scheduling/tasks.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"time"
+
+	"github.com/featureform/fferr"
```
aolfat, line 8: none of the tasks test pass

scheduling/tasks.go
```diff
@@ -70,50 +77,61 @@ func (t *TaskMetadata) Unmarshal(data []byte) error {
 
 	var temp tempConfig
 	if err := json.Unmarshal(data, &temp); err != nil {
-		return fmt.Errorf("failed to deserialize task metadata: %w", err)
+		errMessage := fmt.Errorf("failed to deserialize task metadata: %w", err)
+		return fferr.NewInternalError(errMessage)
 	}
 	if temp.ID == 0 {
-		return fmt.Errorf("task metadata is missing ID")
+		return fferr.NewInvalidArgumentError(fmt.Errorf("task metadata is missing ID"))
```
aolfat, line 84: nit: for readability we should do all the validations first (maybe even in a separate function) and then the rest

scheduling/tasks.go
```diff
@@ -70,50 +77,61 @@ func (t *TaskMetadata) Unmarshal(data []byte) error {
...
 
 	if temp.DateCreated.IsZero() {
-		return fmt.Errorf("task metadata is missing Date value")
+		return fferr.NewInvalidArgumentError(fmt.Errorf("task metadata is missing DateCreated"))
 	}
 	t.DateCreated = temp.DateCreated
 
 	t.TargetType = temp.TargetType
 
 	targetMap := make(map[string]interface{})
```
aolfat, line 110: Do we need to do this?

can we not just do: `t.TargetType = temp.TargetType`

scheduling/taskmetadata.go
aolfat: never mind, simba pointed these all out

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	taskMetadata := TaskMetadata{}
+	err = taskMetadata.Unmarshal([]byte(metadata))
+	if err != nil {
+		return TaskMetadata{}, err
+	}
+	return taskMetadata, nil
+}
+
+func (m *TaskMetadataManager) GetAllTasks() (TaskMetadataList, fferr.GRPCError) {
+	metadata, err := m.storage.List(TaskMetadataKey{}.String())
```
aolfat, line 134: nit: I'd probably do m.storage.List(TaskMetadataKey.AllKeysPrefix)

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	if err != nil {
+		return TaskMetadataList{}, err
+	}
+
+	tml, err := m.getAllTasksAsMetadataList(metadata)
```
aolfat, line 139: what is this doing?

scheduling/taskmetadata.go
aolfat, line 187: i think this is allllll a race condition

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+		return TaskRunMetadata{}, err
+	}
+
+	taskRunMetaKey := TaskRunMetadataKey{taskID: taskID, runID: metadata.ID, date: startTime}
+	err = m.storage.Create(taskRunMetaKey.String(), string(serializedMetadata))
+	if err != nil {
+		err.AddDetail("task name", name)
+		err.AddDetail("task_id", string(taskID))
+		err.AddDetail("run_id", string(metadata.ID))
+		return TaskRunMetadata{}, err
```
aolfat, line 226: i feel like this needs to be transactional. What happens if this part fails, the above was stored but the metadat failed

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	runs := TaskRuns{}
+	err = runs.Unmarshal([]byte(taskRunMetadata))
+	if err != nil {
+		err.AddDetail("task_id", string(taskID))
+		err.AddDetail("run_id", string(runID))
+		return TaskRunMetadata{}, err
+	}
+
+	// Want to move this logic out
+	found, runRecord := runs.ContainsRun(runID)
```
aolfat, line 250: hmmm i don't think this is going to be inefficient yet it could def be optimized.I think TaskRuns.runs should be a map of runId to Runs so it's easily searched

aolfat: actually now that i'm looking at it more it should definitely be that

scheduling/tasks.go
ahmadnazeri: weirdly enough, they pass for me locally

scheduling/taskmetadata.go
ahmadnazeri: This will be used to display the Runs on the dashboard. It allows a filtering based on date range.

scheduling/taskmetadata.go
ahmadnazeri: the key contains a date so this would get the runs for a particular day

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+func (m *TaskMetadataManager) SetRunStatus(runID TaskRunID, taskID TaskID, status Status, err error) fferr.GRPCError {
+	if taskID <= 0 {
+		e := fferr.NewInvalidArgumentError(fmt.Errorf("task id needs to be greater than 0"))
+		e.AddDetail("task_id", string(taskID))
+		return e
+	}
+
+	metadata, e := m.GetRunByID(taskID, runID)
```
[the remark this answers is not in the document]
ahmadnazeri: we need to get it because we need to set the `date: metadata.StartTime`.

scheduling/taskmetadata.go
```diff
@@ -0,0 +1,533 @@
...
+	if time.IsZero() {
+		errMessage := fmt.Errorf("end time cannot be zero")
+		err := fferr.NewInvalidArgumentError(errMessage)
+		err.AddDetail("task_id", string(taskID))
+		err.AddDetail("run_id", string(runID))
+		err.AddDetail("end_time", time.String())
+		return err
+	}
+
+	metadata, e := m.GetRunByID(taskID, runID)
```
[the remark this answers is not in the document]
ahmadnazeri: I totally agree. The problem is that we need the start date unless we ask for it as a parameter.

locker/memory_locker.go
aolfat: channels are already pass by reference: https://stackoverflow.com/questions/44351159/using-pointer-to-channel

scheduling/tasks.go
aolfat: look at the reciever names, sometimes you use t, sometimes you use tm

scheduling/tasks.go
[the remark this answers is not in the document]
anthonylasso: good update, far more legible. resolving.

locker/memory_locker.go
anthonylasso: semantics nit: everything in go is pass-by-value. in your case Ahmad you could remove the explicit pointer in the struct, you don't really need it. When you pass the channel a copy of its pointer is used anyway, so you have access to the original ch.

storage/memory_storage.go
```diff
@@ -0,0 +1,68 @@
...
+type memoryStorageImplementation struct {
+	storage map[string]string
+}
+
+func (m *memoryStorageImplementation) Set(key string, value string) fferr.GRPCError {
+	if key == "" {
+		return fferr.NewInvalidArgumentError(fmt.Errorf("key is empty"))
+	}
+
+	m.storage[key] = value
```
ahmadnazeri, line 25: fail if key exists

scheduling/taskmetadata.go
anthonylasso: for learnings, what was the reason for no factory functions?

ffsync/locker.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		parsingErr := fmt.Errorf("failed to parse date, '%s': %v", tmp.Date, err)
+		return fferr.NewParsingError(parsingErr)
+	}
+	l.Date = parsedTime.UTC()
+
+	return nil
+}
+
+func (l *LockInformation) Marshal() ([]byte, fferr.GRPCError) {
```
simba-git, line 55: Why are we doing this?

ffsync/memory_locker.go
```diff
@@ -0,0 +1,171 @@
...
+
+	go m.updateLockTime(lockKey)
+
+	return lockKey, nil
+}
+
+func (m *memoryLocker) isPrefixOfExistingKey(key string) (string, bool) {
+	existingKey := ""
+	isPrefix := false
+	m.lockedItems.Range(func(k, v interface{}) bool {
```
simba-git, line 84: This doesn't work, Range isn't guaranteed to be consistent as per their docs

ffsync/ordered_id.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type memoryIdGenerator struct {
+	counters map[string]*Uint64OrderedId
+	mu       sync.Mutex
+}
+
+func (m *memoryIdGenerator) NextId(namespace string) (OrderedId, fferr.GRPCError) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
simba-git, line 74: If you use a sync map you may not need this lock at all, but probably not worth it tbh

metadata/metadata.go
```diff
@@ -1900,7 +1900,8 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	} else {
 		// TODO: create a method to get the task id from the name and variant and type
 		// taskId = res.Proto().GetTaskId()
-		taskId = 1
+		id1 := ffsync.Uint64OrderedId(1)
+		taskId = scheduling.TaskID(&id1)
```
simba-git, line 1904: Shouldn't NextID be used here?

metadata/metadata.go
```diff
@@ -1900,7 +1900,8 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 	} else {
 		// TODO: create a method to get the task id from the name and variant and type
```
simba-git, line 1901: Remove TODO?

scheduling/task_run.go
```diff
@@ -4,9 +4,46 @@ import (
...
+	"github.com/featureform/ffsync"
 )
 
-type TaskRunID int32
+type TaskRunKey struct {
+	taskID TaskID
+}
+
+func (trk TaskRunKey) String() string {
+	if trk.taskID == nil || trk.taskID.Value() == 0 {
```
simba-git, line 17: .Value() returns at inferface{} shouldn't the check be if its nil?

simba-git: Why even check .Value() at all?

scheduling/task_run.go
```diff
@@ -4,9 +4,46 @@ import (
...
+	key := "/tasks/runs/metadata"
+
+	// adds the date to the key if it's not the default value
+	if trmk.date != (time.Time{}) {
+		key += fmt.Sprintf("/%s", trmk.date.Format("2006/01/02"))
+
+		// adds the task_id and run_id to the key if they're not the default value
+		taskIdIsNotNil := trmk.taskID != nil
+		runIdIsNotNil := trmk.runID != nil
+		if (taskIdIsNotNil && trmk.taskID.Value() != 0) && (runIdIsNotNil && trmk.runID.Value() != 0) {
```
simba-git, line 39: .Value() != 0 assumes its an int. What is this really trying to check?

scheduling/task_run_list.go
```diff
@@ -4,30 +4,86 @@ import (
...
+	"github.com/featureform/fferr"
+	"github.com/featureform/ffsync"
 )
 
 type TaskRunSimple struct {
 	RunID       TaskRunID `json:"runID"`
 	DateCreated time.Time `json:"dateCreated"`
 }
 
+func (trs *TaskRunSimple) Marshal() ([]byte, fferr.GRPCError) {
```
simba-git, line 17: Why are we not sticking to the JSON package interface?

ffsync/locker.go
ahmadnazeri: we store it to compare with the key that tries to unlock it

metadata/metadata.go
ahmadnazeri: metadata and coordinator changes are done on a different PR, I will address these comments there.

scheduling/tasks.go
```diff
@@ -70,50 +94,63 @@ func (t *TaskMetadata) Unmarshal(data []byte) error {
...
+		return fferr.NewInternalError(errMessage)
 	}
-	t.ID = temp.ID
+
+	id := ffsync.Uint64OrderedId(temp.ID)
+	t.ID = TaskID(&id)
 
 	if temp.Name == "" {
-		return fmt.Errorf("task metadata is missing Name")
+		return fferr.NewInvalidArgumentError(fmt.Errorf("task metadata is missing name"))
```
anthonylasso, line 105: consistency nit: caps the "Name" property like the others messages.

anthonylasso (approved):

PR #1364, opened 2024-03-01
Feature/implement triggers

RiddhiBagadiaa:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/client.py
+++ b/client/src/featureform/client.py
@@ -188,6 +188,56 @@ class Client(ResourceClient, ServingClient):
         """
         self.impl.close()
 
+    def add_trigger(self, trigger_name, resource_name, resource_variant):
+        """
+        Add a trigger to a resource after creation
+
+        **Example:**
+        ```py title="definitions.py"
+        client.add_trigger("my_trigger", "my_resource", "my_variant")
+        ```
+
+        Args:
+            trigger_name (str): The name of the trigger
+            resource_name (str): The name of the resource
+            resource_variant (str): The variant of the resource
+        """
+        self.impl.add_trigger(trigger_name, resource_name, resource_variant)
+        # raise NotImplementedError("This method is not yet implemented")
+    
+    def remove_trigger(self, trigger_name, resource_name, resource_variant):
+        """
+        Remove a trigger from a resource
+
+        **Example:**
+        ```py title="definitions.py"
+        client.remove_trigger("my_trigger", "my_resource", "my_variant")
+        ```
+
+        Args:
+            trigger_name (str): The name of the trigger
+            resource_name (str): The name of the resource
+            resource_variant (str): The variant of the resource
+        """
+        # self.impl.remove_trigger(trigger_name, resource_name, resource_variant)
+        raise NotImplementedError("This method is not yet implemented")
+    
+    def update_trigger(self, trigger_name, schedule):
+        """
+        Update a trigger
+        """
+        # self.impl.update_trigger(trigger_name, schedule)
+        raise NotImplementedError("This method is not yet implemented")
+    
+    def delete_trigger(self, trigger_name):
+        """
+        Delete a trigger
+        """
+        # self.impl.delete_trigger(trigger_name)
+        raise NotImplementedError("This method is not yet implemented")
+
+
+
     @staticmethod
     def _validate_host(host):
         if host.startswith("http://") or host.startswith("https://"):
--- a/client/src/featureform/client.py
+++ b/client/src/featureform/client.py
@@ -7,6 +7,8 @@ from .register import (
     SourceRegistrar,
     SubscriptableTransformation,
     FeatureColumnResource,
+    TriggerResource,
+    TrainingSetVariant,
 )
 from .serving import ServingClient
 from .enums import ResourceType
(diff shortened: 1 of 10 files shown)


anthonylasso:
Riddhi reopen a PR and point it to [feature/trigger-base](https://github.com/featureform/featureform/tree/feature/trigger-base)

we're going to consolidate all trigger branches there

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+        """
+        req = metadata_pb2.TriggerRequest()
+
+        trigger_req = metadata_pb2.Trigger()
+        if isinstance(trigger, str):
+            trigger_req.name = trigger
+        elif isinstance(trigger, TriggerResource):
+            trigger_req.name = trigger.name
+        else:
+            raise ValueError("Invalid trigger type")
```
epps, line 216: Might be helpful here to print the type passed to the function and then also provide the valid types so it's easier for the user to address the mistake.

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+        req.trigger.CopyFrom(trigger_req)
+
+        resource_req = metadata_pb2.ResourceID()
+        if isinstance(resource, tuple):
```
epps, line 220: You probably want to add a check for the tuple length so that you don't run into an `IndexError` exception when indexing to get the name, variant and type. This will also help provide the use with actionable information to correct the mistake.

metadata/metadata.go
```diff
@@ -1423,6 +1437,53 @@ func (resource *entityResource) Update(lookup ResourceLookup, updateRes Resource
...
+	}
+}
+
+func (resource *triggerResource) Schedule() string {
+	// TODO: Make sure it's returning the correct thing
+	return resource.serialized.GetScheduleTrigger().Schedule
+}
+
+func (resource *triggerResource) Dependencies(lookup ResourceLookup) (ResourceLookup, error) {
+	// TODO: Change later, do we need to return all the resources that use this trigger?
```
anthonylasso, line 1457: the only current validation rule we have is we can't delete triggers that have resources currently, so if it's a performance issue, we only need to know that it has some resources not necessarily all of them.

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+            resource_req.resource.name = resource.name
+            resource_req.resource.variant = resource.variant
+            resource_req.resource_type = resource.resource_type
+        else:
+            raise ValueError("Invalid resource type")
+        req.resource.CopyFrom(resource_req)
+
+        self._stub.AddTrigger(req)
+    
+    def remove_trigger(self, trigger, resource):
```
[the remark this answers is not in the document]
sdreyer: trigger name is unique

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+        else:
+            raise ValueError("Invalid trigger type")
+        req.trigger.CopyFrom(trigger_req)
+
+        resource_req = metadata_pb2.ResourceID()
+        if isinstance(resource, tuple):
+            resource_req.resource.name = resource[0]
+            resource_req.resource.variant = resource[1]
+            # TODO: Need to convert this to a real resource type
+            resource_req.resource_type = resource[2]
```
sdreyer, line 264: Same as above, should check the tuple length so we dont go out of bounds with this. Ideally should just make a function that can be used in all these functions instead of repeating code

client/src/featureform/register.py
```diff
@@ -772,6 +773,7 @@ def to_source(self) -> SourceVariant:
             description=self.description,
             tags=self.tags,
             properties=self.properties,
+            # TODO: Add job task id stuff?
```
sdreyer, line 776: Leave out for now. Dont think theres much use for it since its an internal linking

client/src/featureform/register.py
```diff
@@ -1187,6 +1199,9 @@ def __init__(
         schedule: str = "",
         tags: Optional[List[str]] = None,
         properties: Optional[Dict[str, str]] = None,
+        task_id: int = 0,
```
sdreyer, line 1202: Dont need this in the client. These are internal values

client/src/featureform/register.py
```diff
@@ -3662,6 +3696,26 @@ def avg_user_transactions(client, params, entities):
             return decorator
         else:
             return decorator(fn)
+        
+
+    def register_trigger(self, trigger_name, trigger_type):
```
sdreyer, line 3701: Should just have this `ff.ScheduleTrigger(name=”Monday at Noon”, "0 12 * * MON")` that registers the trigger

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+            resource_req.resource.name = resource.name
+            resource_req.resource.variant = resource.variant
+            resource_req.resource_type = resource.resource_type
+        else:
+            raise ValueError("Invalid resource type")
+        req.resource.CopyFrom(resource_req)
+        
+        self._stub.RemoveTrigger(req)
+    
+    def update_trigger(self, trigger, schedule):
```
sdreyer, line 275: Not every trigger will have a schedule. Should have it take in a type the a trigger object

client/src/featureform/resources.py
```diff
@@ -1103,6 +1103,9 @@ class SourceVariant(ResourceVariant):
     transformation: str = ""
     inputs: list = ([],)
     error: Optional[str] = None
+    task_id: int = 0
```
sdreyer, line 1106: These are internal values, not for the client

client/src/featureform/resources.py
```diff
@@ -1244,6 +1253,38 @@ def proto(self) -> pb.Columns:
 
 ResourceLocation = ResourceColumnMapping
 
+@typechecked
+@dataclass
+class TriggerResource:
```
sdreyer, line 1258: This should probably be a base class that can have different classes on top like ScheduleTrigger, WebhookTrigger, etc

client/src/featureform/resources.py
```diff
@@ -1506,6 +1556,9 @@ class LabelVariant(ResourceVariant):
     created: str = None
     status: str = "NO_STATUS"
     error: Optional[str] = None
+    task_id: int = 0
```
sdreyer, line 1559: internal, remove

client/tests/resources_test.py
```diff
@@ -1002,3 +1002,26 @@ def test_add_all_resources_with_schedule(all_resources_strange_order, redis_conf
             name="primary", variant="abc", resource_type=7, schedule_string="* * * * *"
         ),
     ]
+
+def test_register_trigger_resource():
+    state = ResourceState()
+    state.add(TriggerResource(name="trigger", variant="v1", schedule="* * * * *"))
```
sdreyer, line 1008: Triggers dont have variants

metadata/client.go
```diff
@@ -225,6 +227,9 @@ type FeatureDef struct {
 	IsOnDemand  bool
 	IsEmbedding bool
 	Definition  string
+	JobID       int32
+	TaskID      int32
```
sdreyer, line 231: Lets make this an array of int32

metadata/proto/metadata.proto
```diff
@@ -248,6 +266,10 @@ message LabelVariant {
     }
     Tags tags = 13;
     Properties properties = 14;
+    int32 task_id = 15;
```
sdreyer, line 269: For all resources lets make trigger an array of int32

metadata/metadata.go
```diff
@@ -1423,6 +1437,74 @@ func (resource *entityResource) Update(lookup ResourceLookup, updateRes Resource
...
+	}
+}
+
+func (resource *triggerResource) Schedule() string {
+	// TODO: Make sure it's returning the correct thing
+	return resource.serialized.GetScheduleTrigger().Schedule
+}
+
+func (resource *triggerResource) Dependencies(lookup ResourceLookup) (ResourceLookup, error) {
+	// depIdsProtos := resource.serialized.Resources
```
sdreyer, line 1457: Should this be filled?

metadata/metadata.go
```diff
@@ -1423,6 +1437,74 @@ func (resource *entityResource) Update(lookup ResourceLookup, updateRes Resource
...
+	// return deps, nil
+	return make(LocalResourceLookup), nil
+}
+
+func (resource *triggerResource) Proto() proto.Message {
+	return resource.serialized
+}
+
+func (resource *triggerResource) Notify(lookup ResourceLookup, op operation, that Resource) error {
+	// update/list jobs and tasks associated
```
sdreyer, line 1481: Is this unfinished?

metadata/metadata.go
```diff
@@ -1622,6 +1705,23 @@ func (serv *MetadataServer) ListTrainingSets(_ *pb.Empty, stream pb.Metadata_Lis
...
+		trigger, err := serv.lookup.Lookup(ResourceID{Name: t, Type: TRIGGER})
+		if err != nil {
+			return nil, fmt.Errorf("trigger does not exist: %v", err)
+		}
+		asserted_trigger, ok := trigger.(*triggerResource)
+		if !ok {
+			return nil, fmt.Errorf("resource not of type trigger: %v", err)
+		}
+		//asserted_trigger.serialized.Resources = append(asserted_trigger.serialized.Resources, &pb.ResourceID{Resource: &pb.NameVariant{Name: variant.Name, Variant: variant.Variant}, ResourceType: 6})
+		err = serv.lookup.Set(trigger.ID(), asserted_trigger)
```
sdreyer, line 1719: Is this creating the trigger? Unsure what we need this for?

metadata/metadata.go
```diff
@@ -1727,6 +1827,270 @@ func (serv *MetadataServer) GetEntities(stream pb.Metadata_GetEntitiesServer) er
 	})
 }
 
+func (serv *MetadataServer) CreateTrigger(ctx context.Context, trigger *pb.Trigger) (*pb.Empty, error) {
+	return serv.genericCreate(ctx, &triggerResource{trigger}, nil)
+}
+
+func (serv *MetadataServer) AddTrigger(ctx context.Context, tr *pb.TriggerRequest) (*pb.Empty, error) {
```
sdreyer, line 1834: Lets walk through the add and remove when you get the other stuff cleaned up. Theres probably cleaner ways to do these functions

metadata/metadata.go
```diff
@@ -1727,6 +1827,270 @@ func (serv *MetadataServer) GetEntities(stream pb.Metadata_GetEntitiesServer) er
...
+	}
+	// Save the trigger back
+	err = serv.lookup.Set(triggerID, triggerRecord)
+	if err != nil {
+		return nil, err
+	}
+	return &pb.Empty{}, nil
+}
+
+func removeIntFromList(slice []int32, id int32) ([]int32, error) {
```
sdreyer, line 2028: These should probably in the helpers directory and can use generics instead of having one for each type

metadata/metadata.go
```diff
@@ -1892,6 +2256,59 @@ func (serv *MetadataServer) genericCreate(ctx context.Context, res Resource, ini
 		serv.Logger.Error(errors.WithStack(err))
 		return nil, err
 	}
+
+	triggers := []string{}
```
sdreyer, line 2260: All this logic can probably be in propogate change. Lets go through this after the rest is cleaned up

metadata/metadata.go
```diff
@@ -2025,19 +2442,19 @@ func (serv *MetadataServer) genericGet(stream interface{}, t ResourceType, send
 			serv.Logger.Errorw("Generic Get receive error", "error", recvErr)
 			return recvErr
 		}
-		serv.Logger.Infow("Looking up Resource", "id", id)
+		//serv.Logger.Infow("Looking up Resource", "id", id)
```
sdreyer, line 2445: Remove

client/src/featureform/client.py
```diff
@@ -20,6 +23,7 @@ class Client(ResourceClient, ServingClient):
     ```py title="definitions.py"
     import featureform as ff
     from featureform import Client
+from featureform.client.src.featureform import metadata_pb2
```
aolfat, line 26: nit: indent?

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+        """
+        Add a trigger to a resource
+
+        **Example:**
+        ```py title="definitions.py"
+        client.add_trigger("trigger_name", ("resource_name", "resource_variant, resource_type"))
+        ```
+
+        Args:
+            trigger(Union[str, TriggerResource]): The name of the trigger
```
aolfat, line 205: nit: can we add these types to the params in the function

client/src/featureform/client.py
```diff
@@ -188,6 +192,135 @@ def close(self):
...
+            resource(Union[tuple, FeatureColumnResource, TrainingSetVariant]): The name, variant and type of the resource
+        """
+        req = metadata_pb2.TriggerRequest()
```
aolfat, line 208: nit: you could probably move all this logic to its own function since it's reused in both places

to_trigger_request(resource)

client/src/featureform/register.py
```diff
@@ -4035,6 +4099,7 @@ def register_training_set(
             label = (label, self.__run)
         if not isinstance(label, LabelColumnResource) and label[1] == "":
             label = (label[0], self.__run)
+        trigger = [t if isinstance(t, str) else t.name for t in trigger]
```
aolfat, line 4102: nit: triggers? since it's a list?

client/src/featureform/register.py
```diff
@@ -3971,6 +4031,9 @@ def register_training_set(
         schedule: str = "",
         tags: List[str] = [],
         properties: dict = {},
+        task_id: int = 0,
+        job_id: int = 0,
+        trigger: Union[List[str], List[TriggerResource]] = [],
```
aolfat, line 4036: we shouldn't set to a default list, bad python. It shouldn't cause a bug here but just something to watch out

client/src/featureform/resources.py
aolfat: agreed

metadata/proto/metadata.proto
```diff
@@ -90,6 +99,11 @@ service Api {
     rpc ListProviders(Empty) returns (stream Provider);
     rpc ListEntities(Empty) returns (stream Entity);
     rpc ListModels(Empty) returns (stream Model);
+    rpc ListTriggers(Empty) returns (stream Trigger);
+
+    rpc AddTrigger(TriggerRequest) returns (Empty);
```
aolfat, line 104: It's probably fine for now but generally, with rpcs, you want different requests for each.

So the better approach is to have an AddTriggerRequest and a RemoveTriggerRequest. Both of those can you have a TriggerRequest in them but the reason why you'd split it up is that programmers suck at determining what we need in the future so if you need to change the request for add now you're ALSO changing the request for remove; and with how annoying backwards compatability is within grpc it's best to safeguard against it.

metadata/proto/metadata.proto
```diff
@@ -201,6 +216,9 @@ message FeatureVariant {
     bool is_embedding = 19;
     int32 dimension = 20;
     FeatureParameters additional_parameters = 21;
+    int32 task_id = 22;
+    int32 job_id = 23;
+    repeated string trigger = 24;
```
aolfat, line 221: nit: triggers

metadata/proto/metadata.proto
```diff
@@ -248,6 +266,10 @@ message LabelVariant {
     }
     Tags tags = 13;
     Properties properties = 14;
+    int32 task_id = 15;
+    int32 job_id = 16;
+    repeated string trigger = 17;
```
aolfat, line 271: triggers

metadata/proto/metadata.proto
```diff
@@ -288,6 +310,9 @@ message TrainingSetVariant {
     repeated FeatureLag feature_lags = 15;
     Tags tags = 16;
     Properties properties = 17;
+    int32 task_id = 18;
+    int32 job_id = 19;
+    repeated string trigger = 20;
```
aolfat, line 315: same with all of these, if it's a list it should be plural

metadata/metadata.go
```diff
@@ -1622,6 +1705,23 @@ func (serv *MetadataServer) ListTrainingSets(_ *pb.Empty, stream pb.Metadata_Lis
 
 func (serv *MetadataServer) CreateTrainingSetVariant(ctx context.Context, variant *pb.TrainingSetVariant) (*pb.Empty, error) {
 	variant.Created = tspb.New(time.Now())
+	triggers := variant.Trigger
+	for _, t := range triggers {
+		trigger, err := serv.lookup.Lookup(ResourceID{Name: t, Type: TRIGGER})
+		if err != nil {
+			return nil, fmt.Errorf("trigger does not exist: %v", err)
+		}
+		asserted_trigger, ok := trigger.(*triggerResource)
```
aolfat, line 1714: what is an asserted trigger?

metadata/metadata.go
aolfat: +10 on this, you can delegate the creation to each variant instead of putting all the logic in here

metadata/metadata.go
aolfat: +1 on this

metadata/metadata.go
RiddhiBagadiaa: Once it is verified that trigger is of type Trigger Resource, its named asserted trigger

PR #1375, opened 2024-03-05
Install, update, or delete nightly namespace

anthonylasso:
# Description

This PR adds upgrades and deletes to our publish featureform automation in the nightly box. 

Also fresh installs now come with Redis and Postgres providers out-of-the-box. 

[Documentation](https://www.notion.so/Push-Features-to-Nightly-Box-7043b8db57094f6a91b42e1843c146ef)
[Upgrade Run](https://github.com/featureform/featureform/actions/runs/8209045984)
[Delete Run](https://github.com/featureform/featureform/actions/runs/8209382901)

`deployment upgrade`
Screenshot 2024-03-08 at 3 31 35 PM

### Select type(s) of change

- [x] Documentation update

# Checklist:


.github/workflows/publish-oss-feature-branch.yml
```diff
@@ -364,11 +378,54 @@ jobs:
...
       ]
+    outputs:
+      found: ${{ steps.nscheck.outputs.found }}
     steps:
       - uses: actions/checkout@v3
-      - name: Install helm chart
+      - name: Find the namespace
+        id: nscheck
         uses: WyriHaximus/github-action-helm3@v3
         with:
```
anthonylasso, line 388: technically helm upgrades can be done with `helm upgrade --install` but other steps and jobs needs to know explicitly.

ahmadnazeri (approved):

charts/featureform/charts/metadata/templates/deployment.yaml
```diff
@@ -20,6 +20,8 @@ spec:
         chart: featureform
         run: featureform-metadata-server
         app: featureform-metadata-server
+      annotations:
```
sdreyer, line 23: What do these do?

sdreyer (approved):

charts/featureform/charts/metadata/templates/deployment.yaml
anthonylasso: these are a custom timestamp that change the resulting deployment yaml spec when we call `helm upgrade`. 
if not, when we call `helm upgrade` if k8s sees no diff in the resulting spec it won't cause the deploys to restart (even if technically you posted a new image with the same tag in docker).

PR #1381, opened 2024-03-13
All Train Test Split Changes

aolfat:
# Description

All Training Set Split changes thus far

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/clickhouse.go
```diff
@@ -996,20 +1047,88 @@ func (store *clickHouseOfflineStore) GetTrainingSet(id ResourceID) (TrainingSetI
...
+	trainRows, err := store.db.Query(train)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("could not query train set: %v", err)
+
+	}
+
+	colTypes, err := store.getValueColumnTypes(trainingTestSplitName)
+	fmt.Printf("these are the column types: %v\n", colTypes)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("could not get column types: %v", err)
```
simba-git, line 1113: If this happened, dropFunc would never be called right?

Maybe force the caller to call create and defer close then get the iterators. Would also make the function signature a little cleaner.

provider/clickhouse.go
```diff
@@ -996,20 +1047,88 @@ func (store *clickHouseOfflineStore) GetTrainingSet(id ResourceID) (TrainingSetI
...
+	}
+
+	// return callback to drop view
+	dropFunc := func() error {
+		// two queries to drop split and row number table
```
simba-git, line 1118: Out of date comment?

provider/clickhouse.go
```diff
@@ -996,20 +1047,88 @@ func (store *clickHouseOfflineStore) GetTrainingSet(id ResourceID) (TrainingSetI
...
 	}
 	return store.newsqlTrainingSetIterator(rows, colTypes), nil
 }
 
+func (store *clickHouseOfflineStore) GetTrainingSetTestSplit(
+	id ResourceID,
+	testSize float32,
+	shuffle bool,
+	randomState int,
+) (TrainingSetIterator, TrainingSetIterator, func() error, error) {
```
simba-git, line 1082: Dont love having this many things returned, I have advice in a comment below though

client/src/featureform/serving.py
```diff
@@ -424,9 +447,129 @@ def __init__(self, stream, dataframe=None):
...
+        train_size: float = 0,
+        shuffle: bool = True,
+        random_state: Union[int, None] = None,
+        batch_size: int = 1,
+    ):
+        """
+        (This functionality is currently only available for Clickhouse).
+
+        Splits an existing training set into training and testing iterators. The split is processed on the underlying
+        provider and calculated and serving time.
```
anthonylasso, line 462: replace "calculated **_and_** serving time." with "calculated **_at_** serving time."

client/src/featureform/serving.py
```diff
@@ -424,9 +447,129 @@ def __init__(self, stream, dataframe=None):
...
+            random_state (Union[int, None]): A random state to shuffle the dataset. If None, the dataset will be shuffled
+                randomly on every call. If >0, the value will be used a seed to create random shuffle that can be repeated
+                if subsequent calls use the same seed.
+            batch_size (int): The size of the batch to return from the iterator. Must be greater than 0.
+
+        Returns:
+            train (Iterator): An iterator for training values.
+            test (Iterator): An iterator for testing values.
+        """
+        if batch_size < 1:
```
anthonylasso, line 530: will the min batch size be configurable at some point? if so I'd replace this magic number with a const + interpolate the value error msg with the const. 

if not ignore.

client/src/featureform/serving.py
```diff
@@ -424,9 +447,129 @@ def __init__(self, stream, dataframe=None):
...
+        if random_state == 0:
+            raise ValueError("random_state must be greater than zero or None")
+
+        test_size, train_size = self.validate_test_size(test_size, train_size)
+
+        name = self._stream.name
+        variant = self._stream.version
+        stub = self._stream._stub
+        model = self._stream.model if hasattr(self._stream, "model") else None
+        if random_state is None:
```
anthonylasso, line 542: shouldn't this `random_state` conditional be a part of the first if check on line 533?

also it sets the `random_state` to zero, but that's an error state value in the first conditional

client/src/featureform/serving.py
```diff
@@ -712,6 +857,26 @@ def parse_proto_value(value):
...
+        "int32_value": np.int32,
+        "int64_value": np.int64,
+        "float_value": np.float32,
+        "double_value": np.float64,
+        "bool_value": bool,
+    }
+    return type_mapping[value.WhichOneof("value")]
+
+
+def get_numpy_array_type(types):
```
anthonylasso, line 873: i'd throw a parameterized unit test at this function. reads like it could easily be a source of a hidden logic bug.

proto/serving.proto
```diff
@@ -148,3 +149,33 @@ message ResourceIdRequest {
...
 }
+
+message TrainTestSplitRequest {
+  TrainingDataID id = 1;
+  Model model = 2;
+  float test_size = 3;
+  float train_size = 4;
+  bool shuffle = 5;
+  int32 random_state = 6;  // Seed for shuffling, if shuffle is true
+  RequestType request_type = 7;  // Specify the type of data being requested
```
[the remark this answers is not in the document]
aolfat: i can remove these comments, they were generated

client/src/featureform/train_test_split.py
```diff
@@ -0,0 +1,212 @@
...
+        batch_features_list = []
+        batch_labels_list = []
+
+        result = self._split_stream.send_request(self.request_type)
+        data = result.data
+
+        if self._np_type_parser is None and data.rows:
+            self._np_type_parser = _NpProtoTypeParser.init_types(data.rows[0])
+
+        for i, row in enumerate(data.rows):
```
anthonylasso, line 155: not using the index var "i". can just drop it like: 

`for row in enumerate(data.rows)`

client/src/featureform/train_test_split.py
```diff
@@ -0,0 +1,212 @@
...
+    Returns two iterators, one for the training set and one for the test set, in that order
+    """
+
+    def __init__(
+        self,
+        stub,
+        name: str,
+        version: str,
+        test_size: float,
+        train_size: float,  # unused for now
```
anthonylasso, line 186: if unused remove from `init`?

provider/clickhouse.go
```diff
@@ -954,6 +955,53 @@ func (store *clickHouseOfflineStore) CreateTrainingSet(def TrainingSetDef) error
...
 }
 
+func (store *clickHouseOfflineStore) CreateTrainingTestSplit(
+	trainingSetTable string,
+	testSize float32,
+	shuffle bool,
+	randomState int,
+) (string, error) {
+	// Generate unique suffix for the view names
+	tableNameSuffix := fmt.Sprintf("%s_%d_%t_%d", trainingSetTable, int(testSize*100), shuffle, randomState)
```
anthonylasso, line 965: `int(testSize*100)` why 100, is it not better to use rand int here?

client/src/featureform/serving.py
```diff
@@ -109,8 +111,27 @@ def training_set(
...
+            name (str): The name of the training set
+            variation (str): The variation of the training set
+            model (Union[str, Model]): The model to use for the training set
+
+        Returns:
+            train_set (Dataset): The training set
+            test_set (Dataset): The test set
+            :param variant:
+        """
+        train, test = self.impl.training_set_test_split(name, variant, model)
```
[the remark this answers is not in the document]
aolfat: good catch, this should actually be deleted

client/src/featureform/serving.py
aolfat: min batch size has to be 1 right? effectively meaning no batch

client/src/featureform/serving.py
aolfat: also really good catch

serving/serving.go
```diff
@@ -82,13 +79,176 @@ func (serv *FeatureServer) TrainingData(req *pb.TrainingDataRequest, stream pb.F
...
+	for {
+		if isTrainFinished && isTestFinished {
+			// If both iterators are finished, we can close the stream
+			serv.Logger.Infow("Both iterators are finished, closing stream")
+			return nil
+		}
+
+		req, err := stream.Recv()
+		if err != nil {
+			return err
```
anthonylasso, line 108: int: log that the error happened at recv()

serving/serving.go
```diff
@@ -134,29 +294,54 @@ func (serv *FeatureServer) SourceData(req *pb.SourceDataRequest, stream pb.Featu
...
 		// This means that the provider of the training set isn't an offline store.
 		// That shouldn't be possible.
-		return nil, errors.Wrap(err, "could not open as offline store")
+		return nil, err
 	}
 	serv.Logger.Debugw("Get Training Set From Store", "name", name, "variant", variant)
 	return store.GetTrainingSet(provider.ResourceID{Name: name, Variant: variant})
 }
 
+func (serv *FeatureServer) getTrainingSetTestSplitIterator(name, variant string, testSize float32, shuffle bool, randomState int) (provider.TrainingSetIterator, provider.TrainingSetIterator, func() error, error) {
```
anthonylasso, line 320: nit: too many returns here. maybe use a wrapper `SplitResp` struct or similar.

provider/bigquery.go
```diff
@@ -214,6 +214,15 @@ func (it *bqGenericTableIterator) Values() GenericRecord {
 
 func (it *bqGenericTableIterator) Columns() []string {
 	var columns []string
+	// As the documentation for bigquery.Schema notes:
```
aolfat, line 217: ignore this

client/src/featureform/train_test_split.py
aolfat: sterling said to keep it and I don't care enough lol

serving/serving.go
```diff
@@ -82,13 +79,182 @@ func (serv *FeatureServer) TrainingData(req *pb.TrainingDataRequest, stream pb.F
...
+}
+
+func (serv *FeatureServer) handleFinishedIterator(trainingDataRows []*pb.TrainingDataRow, ctx *splitContext) {
+	if ctx.req.GetRequestType() == pb.RequestType_TEST {
+		*ctx.isTestFinished = true
+	} else if ctx.req.GetRequestType() == pb.RequestType_TRAINING {
+		*ctx.isTrainFinished = true
+	}
+
+	if *ctx.isTestFinished && *ctx.isTrainFinished {
```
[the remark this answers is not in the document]
aolfat: good q.

a request is either a testtype OR a traintype, never both, if one of them finishes we return an iterator finished response so that the iteration is stopped but the stream isn't closed

serving/serving.go
```diff
@@ -82,13 +79,182 @@ func (serv *FeatureServer) TrainingData(req *pb.TrainingDataRequest, stream pb.F
...
+			}
+			trainingDataRows = append(trainingDataRows, sRow)
+			rows++
+		} else {
+			// if we reach the end of the iterator mid-batch, we'll send the processed rows so far and end the iteration
+			serv.handleFinishedIterator(trainingDataRows, ctx)
+		}
+	}
+
+	response := &pb.BatchTrainTestSplitResponse{
```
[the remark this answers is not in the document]
aolfat: yeahhh it's kindaaaa weird

so the flow is this ->

we make a trainsplitreq -> recieving by api/main -> send to serving.go -> recieve req by serving -> create response by serving and SEND response -> recv resp on main -> send resp on main

serving/serving.go
```diff
@@ -82,13 +79,182 @@ func (serv *FeatureServer) TrainingData(req *pb.TrainingDataRequest, stream pb.F
...
+		}
+
+		id := req.GetId()
+		name, variant := id.GetName(), id.GetVersion()
+		logger := serv.Logger.With("Name", name, "Variant", variant, "RequestType", req.GetRequestType().String())
+
+		featureObserver := serv.Metrics.BeginObservingTrainingServe(name, variant)
+		defer featureObserver.Finish()
+
+		ctx := splitContext{
```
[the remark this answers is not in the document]
aolfat: yeahhh i thought about that, I'm a big advocate of just typing things out so I'll do that

serving/serving.go
```diff
@@ -82,13 +79,182 @@ func (serv *FeatureServer) TrainingData(req *pb.TrainingDataRequest, stream pb.F
 	return nil
 }
 
+type splitContext struct {
```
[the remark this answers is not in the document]
aolfat: what is the convention?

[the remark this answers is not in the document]
aolfat: I mostly agree and thought about it a bit too but with it being an internal struct and literally being the context for all the functions I thought it would do.

PR #1389, opened 2024-03-18
Bugfix: Can't materialize more than 100k rows/entities from BigQuery to Redis 

epps:
# Description

This PR introduces a fix for a user-reported issue that identified that only 100K rows/entities were being materialized to the online store despite there being 500K records.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/runner/materialize.go
+++ b/runner/materialize.go
@@ -158,7 +158,7 @@ func (m MaterializeRunner) Run() (types.CompletionWatcher, error) {
 			numChunks += 1
 		}
 	}
-	m.Logger.Infow("Creating chunks", "name", m.ID.Name, "variant", m.ID.Variant, "count", numChunks)
+	m.Logger.Infow("Creating chunks", "name", m.ID.Name, "variant", m.ID.Variant, "count", numChunks, "chunkSize", chunkSize)
 	config := &MaterializedChunkRunnerConfig{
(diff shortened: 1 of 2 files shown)


runner/materialize.go
```diff
@@ -197,7 +197,13 @@ func (m MaterializeRunner) Run() (types.CompletionWatcher, error) {
 		m.Logger.Infow("Making Local Runner", "name", m.ID.Name, "variant", m.ID.Variant)
 		completionList := make([]types.CompletionWatcher, int(numChunks))
 		for i := 0; i < int(numChunks); i++ {
-			localRunner, err := Create(COPY_TO_ONLINE, serializedConfig)
+			m.Logger.Infow("Creating materialization chunk", "name", m.ID.Name, "variant", m.ID.Variant, "chunkIndex", i, "chunkSize", chunkSize)
+			config.ChunkIdx = int64(i)
+			serializedChunkConfig, err := config.Serialize()
```
sdreyer, line 202: Was the other config not passing the chunk size in resulting in 0 everytime?

epps: Exactly, so while the number of chunks was being correctly evaluated,  the `ChunkIndex` for each one was always 0, which means they all query the first page of data.

ahmadnazeri (approved):

PR #805, opened 2023-05-15
FeatureForm V1 Dashboard

RedLeader16:
# Description
This PR implements the [V1 dashboard design](https://www.figma.com/file/vm29ey3litU8XsYCZBAlcI/v1-Dashboard?type=design&node-id=0-1&t=nZLbLvFAEu6qmVha-0), fixes a search bar navigation bug, and fixes an issue where revisiting a resources does not fire a new fetch request. 

## Type of change

Example layout: 
Screenshot 2023-05-16 at 5 08 53 PM

Search history sync:

https://github.com/featureform/featureform/assets/34227093/d7e2b3d4-1bc7-44d9-870d-8ea823c6113d

### Select type(s) of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have fixed any merge conflicts

PR #1393, opened 2024-03-19
Bugfix: Fix Pyspark stats dataframe date

anthonylasso:
# Description

This PR fixes the issue where the `execute_df_job` does not take into account date timestamps in the row inputs.

The fix is to detect and format the datetime objects in the stats dataframe before using it to write the json stats file.

This avoids seeing the dates represented as java  util date strings in the stats file. `java.util.GregorianCalendar[time=?,areFieldsSet=false,areAllFieldsSet=false,lenient=true...` 

closes [ent-#805](https://github.com/featureform/featureform-enterprise/issues/805)

`example`
Screenshot 2024-03-20 at 2 47 15 PM

`test_execute_df_job success`
image

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -7,7 +7,7 @@
 import sys
 import types
 import uuid
-from datetime import datetime
+import datetime
```
anthonylasso, line 10: need access to `isinstance`

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -71,8 +71,12 @@ def display_data_metrics(df, spark):
     if len(rows_as_list) > 100:
         rows_as_list = rows_as_list[:100]
     for row in rows_as_list:
-        row_list = row.asDict().values()
-        rows.append(list(row_list))
+        row_values = list(row.asDict().values())
+        for index in range(len(row_values)):
+            curr_value = row_values[index]
+            if isinstance(curr_value, datetime.datetime):
+                row_values[index] = curr_value.strftime("%m/%d/%Y")
```
sdreyer, line 78: Is this the highest precision it can have? Should try to maintain the given precision

anthonylasso: yeah I can add the time itself. one sec.

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -71,8 +71,12 @@ def display_data_metrics(df, spark):
     if len(rows_as_list) > 100:
         rows_as_list = rows_as_list[:100]
     for row in rows_as_list:
-        row_list = row.asDict().values()
-        rows.append(list(row_list))
+        row_values = list(row.asDict().values())
+        for index in range(len(row_values)):
```
ahmadnazeri, line 75: small change:
```
for i, curr_value in enumerate(row_values):
...
```

provider/scripts/spark/offline_store_spark_runner.py
ahmadnazeri: why not keep it `from datetime import datetime`?

ahmadnazeri (approved):
LGTM, just one question and small suggestion

provider/scripts/spark/offline_store_spark_runner.py
anthonylasso: yeah much cleaner

anthonylasso: done

provider/scripts/spark/offline_store_spark_runner.py
anthonylasso: unfortunately that causes `datetime.date` to refer to a method of the `datetime` class (which doesn't work with `isinstance`)

so to fix you instead pull the module, so then `datetime.date` refers to the class, and `isinstance` will work.

anthonylasso: I always forget that and keep re-googling to the same stack post lol 

https://stackoverflow.com/questions/5665597/python-isinstance-returning-error-with-datetime-date

provider/scripts/spark/offline_store_spark_runner.py
```diff
@@ -71,8 +71,11 @@ def display_data_metrics(df, spark):
     if len(rows_as_list) > 100:
         rows_as_list = rows_as_list[:100]
     for row in rows_as_list:
-        row_list = row.asDict().values()
-        rows.append(list(row_list))
+        row_values = list(row.asDict().values())
+        for i, curr_value in enumerate(row_values):
+            if isinstance(curr_value, datetime.datetime):
+                row_values[i] = curr_value.strftime("%m/%d/%Y, %H:%M:%S")
```
sdreyer, line 77: wondering if this makes more sense to use a dataframe method here instead of iterating through everything

sdreyer (approved):

PR #1405, opened 2024-03-28
Variant Control pre-work

anthonylasso:
# Description

This PR serves as a bit of cleanup before implementing the main variant rework. 

It cleans up parts of the tags, entity view, and code view. 

`Jest Tests`
image

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/dashboard/src/components/entitypage/EntityPageView.js
+++ b/dashboard/src/components/entitypage/EntityPageView.js
@@ -646,9 +646,13 @@ const EntityPageView = ({
                             variant='body1'
                             className={classes.typeTitle}
                           >
+                            <Typography
+                              variant='body1'
+                              fontWeight='bold'
+                            ></Typography>
                             <strong>Columns:</strong>{' '}
                           </Typography>
-                          <Typography variant='body2'>
+                          <Typography variant='body2' style={{ marginTop: 2 }}>
                             &nbsp;<strong>Entity:</strong>{' '}
                             {metadata['location'].Entity}
                             &nbsp;<strong>Value:</strong>{' '}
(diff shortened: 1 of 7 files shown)


dashboard/src/components/entitypage/EntityPageView.js
```diff
@@ -39,7 +40,7 @@ SyntaxHighlighter.registerLanguage('json', json);
 export function getFormattedSQL(sqlString = '') {
   let stringResult = sqlString;
   try {
-    stringResult = format(sqlString.replace('{{', '').replace('}}', ''), {
+    stringResult = format(sqlString.replace('{{ ', '"').replace(' }}', '"'), {
```
sdreyer, line 43: Do we want this? Will it get confusing if a table actually has quotes?

sdreyer (approved):

dashboard/src/components/entitypage/statusChip.js
```diff
@@ -0,0 +1,20 @@
+import { Chip } from '@mui/material';
+import React from 'react';
+
+export default function StatusChip({ status = '' }) {
+  console.log(status);
+  const chipMap = {
+    NO_STATUS: '#FFFFFF',
+    CREATED: '#FFFFFF',
```
ahmadnazeri, line 8: just curious why `NO_STATUS` and `CREATED` are both black?

ahmadnazeri (approved):

dashboard/src/components/entitypage/EntityPageView.js
anthonylasso: that is a possibility, but without these quotes, a table like `transactions.2024-03-27t14-53-01` will cause the sql formatter to fail and it spits out an ugly one line string.

dashboard/src/components/entitypage/statusChip.js
anthonylasso: good catch (both are white) I'll quickly just set something for `created`

dashboard/src/components/entitypage/EntityPageView.js
anthonylasso: worst case, if a table does have quotes (like in Oracle or something) we can just check for pairs and remove one

PR #1419, opened 2024-04-05
Feature/fast materialize/simba

simba-git:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:


provider/dynamodb.go
```diff
@@ -801,7 +850,10 @@ func castNumberToFloat32(value any) (float32, error) {
 		return float32(typed), nil
 	case string:
 		f64, err := strconv.ParseFloat(typed, 32)
-		return float32(f64), err
+		if err != nil {
+			return 0, fmt.Errorf("Type error: Expected numerical type and got %T", typed)
+		}
```
epps, line 855: Do you mean to use `fferr` here?

provider/dynamodb.go
```diff
@@ -431,6 +443,43 @@ func (store *dynamodbOnlineStore) GetImport(id ImportID) (Import, error) {
...
+	for i, serItem := range serialized {
+		reqs[i] = types.WriteRequest{PutRequest: &types.PutRequest{Item: serItem}}
+	}
+	batchInput := &dynamodb.BatchWriteItemInput{
+		RequestItems: map[string][]types.WriteRequest{
+			table.key.ToTableName(): reqs,
+		},
+	}
+	if _, err := table.client.BatchWriteItem(context.TODO(), batchInput); err != nil {
+		return err
```
epps, line 474: Do you mean to use `fferr` here given we'd be returning the client's error instead of ours?

provider/dynamodb.go
```diff
@@ -825,7 +877,11 @@ func castNumberToFloat64(value any) (float64, error) {
 	case float64:
 		return typed, nil
 	case string:
-		return strconv.ParseFloat(typed, 64)
+		f64, err := strconv.ParseFloat(typed, 64)
+		if err != nil {
+			return 0, fmt.Errorf("Type error: Expected numerical type and got %T", typed)
+		}
```
epps, line 883: Do you mean to use `fferr` here, too?

provider/filestore.go
```diff
@@ -976,12 +1004,12 @@ func (store *genericFileStore) Write(path filestore.Filepath, data []byte) error
...
 		func() error {
 			blob, errRetr := store.bucket.ReadAll(ctx, path.Key())
-			fmt.Printf("Read (%d) bytes from bucket (%s) after write\n", len(data), path.Key())
 			if errRetr != nil {
-				return re.Unrecoverable(errRetr)
+				return errRetr
 			} else if !bytes.Equal(blob, data) {
 				return fmt.Errorf("blob read from bucket does not match blob written to bucket")
 			}
+			fmt.Printf("Read (%d) bytes from bucket (%s) after write\n", len(blob), path.Key())
```
epps, line 1012: Do you want to convert this to a `Debugw` log?

simba-git: Will leave for Riddhi to figure out

provider/dynamodb.go
simba-git: these get wrapped in the method above

provider/dynamodb.go
```diff
@@ -327,6 +338,7 @@ func (store *dynamodbOnlineStore) CheckHealth() (bool, error) {
 	return true, nil
 }
 
+// TODO(simba) Make this work with Serialize V1
```
anthonylasso, line 341: todo?

provider/k8s.go
```diff
@@ -1175,6 +1175,7 @@ func (mat FileStoreMaterialization) NumRows() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
+	// TODO(simba) THIS IS INCORRECT!
```
anthonylasso, line 1178: todo?

provider/k8s.go
```diff
@@ -1206,6 +1207,7 @@ func (mat FileStoreMaterialization) IterateSegment(begin, end int64) (FeatureIte
 	}
 	i := int64(0)
 	for i = 0; i < begin; i++ {
+		// TODO(simba) Can't actually skip the error here
```
anthonylasso, line 1210: error + todo

provider/k8s.go
```diff
@@ -1215,17 +1217,69 @@ func (mat FileStoreMaterialization) IterateSegment(begin, end int64) (FeatureIte
 	}, nil
 }
 
+func (mat FileStoreMaterialization) NumChunks() (int, error) {
+	searchPath, err := mat.store.CreateFilePath(fileStoreResourcePath(mat.id))
+	if err != nil {
+		return -1, err
```
anthonylasso, line 1223: convert -1 to a const. easy to fat finger a diff value between the 8 uses.

provider/offline.go
```diff
@@ -28,6 +28,9 @@ import (
 	"golang.org/x/text/language"
 )
 
+// defaultRowsPerChunk is the number of rows in a chunk when using materializations.
+const defaultRowsPerChunk int64 = 100000
```
anthonylasso, line 32: nit: use `100_000` to make it more legible.

provider/offline.go
```diff
@@ -953,24 +964,44 @@ func (table *memoryOfflineTable) WriteBatch(recs []ResourceRecord) error {
 	return nil
 }
 
-type memoryMaterialization struct {
-	id   MaterializationID
-	data []ResourceRecord
+// Used in runner/copy_test.go
+type MemoryMaterialization struct {
+	// Lowercase do so it doesn't name thrash ID()
```
anthonylasso, line 969: ??

coordinator/coordinator_test.go
```diff
@@ -1350,17 +1337,25 @@ func testCoordinatorTrainingSet(addr string) error {
...
-		fmt.Printf("waiting for job %v to be deleted\n", tsID)
 	}
-	if elapsed >= time.Duration(10)*time.Second {
+	if elapsed >= waitTime {
 		return fmt.Errorf("timed out waiting for job to delete")
 	}
 	ts_complete, err := client.GetTrainingSetVariant(ctx, metadata.NameVariant{Name: tsName, Variant: ""})
 	if err != nil {
-		return fmt.Errorf("could not get training set variant")
+		return fmt.Errorf("could not get training set variant: %v", err)
```
aolfat, line 1358: should we fferror it

provider/filestore.go
```diff
@@ -270,6 +277,27 @@ func NewS3FileStore(config Config) (FileStore, error) {
 		return nil, wrapped
 	}
 
+	args := []func(*awsv2cfg.LoadOptions) error{
```
aolfat, line 280: what is all this new stuff for? so we can connect with retries?

runner/copy.go
```diff
@@ -258,39 +282,64 @@ func (m *MaterializedChunkRunnerConfig) Deserialize(config Config) error {
...
-		return nil, err
-	}
-	offlineProvider, err := provider.Get(runnerConfig.OfflineType, runnerConfig.OfflineConfig)
-	if err != nil {
-		return nil, err
-	}
-	onlineStore, err := onlineProvider.AsOnlineStore()
-	if err != nil {
-		return nil, err
+	cachedOnline, found := onlineProviderCache.Load(onlineCfgStr)
```
aolfat, line 303: I wonder if it makse sense to move the caching logic into provider.Get

aolfat: @sdreyer did some optimization work for redis I wonder if any of that work applies

provider/filestore.go
aolfat: just a thought, this looks like code I would copy from here and paste in a scratch file to connect to s3 to test out their api's locally, might be good to have it as a separate function so folks can easily search/use it.

provider/dynamodb.go
simba-git: Left on purpose

provider/k8s.go
simba-git: Old code

provider/k8s.go
simba-git: Old code

provider/k8s.go
simba-git: Yeah but its actually the error that should be checked, the -1 is to hopefully force some sort of error if you forget to check the err

simba-git: If this was a special value that meant error, I'd agree, but I just arbritrarily decided -1 instead of 0

provider/offline.go
simba-git: I'll make clearer, it can't be named Id because there is already an ID method on this struct

runner/copy.go
simba-git: I actually turned this off, because it breaks all tests and has really gnarly edge cases where you can get two different providers with the same config

provider/filestore.go
simba-git: I want to move all the connection stuff into one place soon, makes testing easier too

sdreyer (approved):

provider/dynamodb.go
```diff
@@ -431,6 +443,43 @@ func (store *dynamodbOnlineStore) GetImport(id ImportID) (Import, error) {
 	return S3Import{id: id, status: string(output.ImportTableDescription.ImportStatus), errorMessage: errorMessage}, nil
 }
 
+const maxDynamoBatchSize = 25
```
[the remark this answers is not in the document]
simba-git: dynamo limit

provider/k8s.go
anthonylasso: it's less about the error checking logic and more on the syntax consistency of using a sentinel value in multiple spots. Like if someone fat fingers `-11` then that could cause a bug if a conditional explicitly checks for -1

ex. `if iter.maxIdx != -1` -> some random explosion or data bug

but if you think it can't happen then you're good to resolve!

aolfat: I think what Simba says makes sense, the  value doesn't matter if there's an error

simba-git: Yeah but shouldn't depend on that value at all, should check the error. I see your point, but I think having a const and exporting it is overkill since it'd have to be documented and would make the behavior fixed.

PR #1424, opened 2024-04-16
Tests: Adding additional behave tests

ahmadnazeri:
# Description
The goal of this PR is add some additional tests in order to test out our client API functionality more robustly. 

This is one PR with behave tests, I will be adding additional behave tests in a separate PR. 

### Select type(s) of change

- [ ] Documentation update

# Checklist:


tests/end_to_end/features/autovariants.feature
```diff
@@ -0,0 +1,47 @@
+Feature: AutoVariants
+    Scenario: Same auto variant for same transformation 
+        Given Featureform is installed
+        When I turn on autovariants
```
aolfat, line 4: i was wrong about this, simba pointed out a good observation, since things are run in parallel this won't have the affect we want

tests/end_to_end/features/steps/autovariants.py
```diff
@@ -0,0 +1,132 @@
...
+
+@then("I turn off autovariants")
+def step_impl(context):
+    os.environ["FF_GET_EQUIVALENT_VARIANTS"] = "false"
+
+    context.variant = ff.get_run()
+
+
+@when("I register a transformation with auto variant")
+def step_impl(context):
```
aolfat, line 23: Ideally this is "I register a transformation with {{variant}}"

And if it's empty we default to autovariants.

tests/end_to_end/features/autovariants.feature
```diff
@@ -0,0 +1,47 @@
...
+        And I create a "hosted" "insecure" client for "localhost:7878"
+        And I register "s3" filestore with bucket "featureform-spark-testing" and root path "data"
+        And I register databricks
+        And I register transactions_short.csv
+        And I register a transformation with auto variant
+        Then I should be able to reuse the same variant for the same transformation
+        And I can get the transformation as df
+        And I turn off autovariants
```
anthonylasso, line 12: may be a good idea to convert `I turn on/off autovariants` into @before & @after hooks to avoid test pollution if a scenario fails in a weird way. Those get defined in the step definitions files.

tests/end_to_end/features/autovariants.feature
```diff
@@ -0,0 +1,47 @@
...
+    
+    Scenario: Different auto variant for different transformation
+        Given Featureform is installed
+        When I turn on autovariants
+        And I create a "hosted" "insecure" client for "localhost:7878"
+        And I register "s3" filestore with bucket "featureform-spark-testing" and root path "data"
+        And I register databricks
+        When I register transactions_short.csv
+        And I register a transformation with auto variant
+        Then I should be able to register a modified transformation with new auto variant
```
aolfat, line 34: a bit of a vent but maybe you share the same opinion: I don't like that behave tests are just becoming glorified comments for code. I think having a definitions file with asserts might be better. I've been finding it even slightly difficult to navigate to see what the previous step is doing, often times I'd rather just look at code.

Are we going to register the same databricks for each one, the health check has to run and it's going to take awhile? should we use the same one?

aolfat: I'm not advocating you completely change how things are done here but just saying with the path a lot of these behave tests are going it might be better to do what sterling talked about and shove definitions file into the behave framework

tests/end_to_end/features/steps/autovariants.py
```diff
@@ -0,0 +1,132 @@
...
+        context.new_transformation.name_variant()[0]
+        == context.transformation.name_variant()[0]
+    )
+    assert (
+        context.new_transformation.name_variant()[1]
+        != context.transformation.name_variant()[1]
+    )
+
+
+@then("I should be able to register a source with user-defined variant")
```
aolfat, line 118: what do you think about "register file (with variant {variant}" or something more reusable

anthonylasso (approved):

aolfat (approved):

PR #1503, opened 2024-11-12
Add ArgoCD and Argo Rollout capabilities to charts

ff-kamal:
Adds Argo Rollout files, which are the Argo equivalent of Deployments. These are required in order to use the capabilities of Argo rollouts.

Additionally, adds configuration so that /argocd/ can be forwarded to the Argo web server.

Note that installing ArgoCD and Argo Rollouts on the cluster is a prerequisite for these to work, since these are cluster-wide dependencies (this can be done through the infrastructure repo, through the use of Terraform modules that provision it).

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/charts/featureform/templates/api-server/rollout.yaml
+++ b/charts/featureform/templates/api-server/rollout.yaml
@@ -1,0 +1,20 @@
+{{ if .Values.argo.rollouts.enabled }}
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
+  name: featureform-api-server
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: featureform-api-server
+  workloadRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: featureform-api-server
+    scaleDown: onsuccess
+  strategy:
+    blueGreen:
+      activeService: featureform-api-server
+      autoPromotionEnabled: false
+{{ end }}
--- a/charts/featureform/templates/api-server/rollout.yaml
+++ b/charts/featureform/templates/api-server/rollout.yaml
@@ -1,3 +1,10 @@
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+#  Copyright 2024 FeatureForm Inc.
+#
+
 {{ if .Values.argo.rollouts.enabled }}
 apiVersion: argoproj.io/v1alpha1
 kind: Rollout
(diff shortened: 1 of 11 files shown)


charts/featureform/templates/api-server/rollout.yaml
```diff
@@ -0,0 +1,27 @@
...
+      app: featureform-api-server
+  workloadRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: featureform-api-server
+    scaleDown: onsuccess
+  strategy:
+    blueGreen:
+      activeService: featureform-api-server
+      autoPromotionEnabled: false
```
anthonylasso, line 26: q: `autoPromotionEnabled`  set to `false` keeps the actual flip to the new version manual right?

charts/featureform/templates/ingress/argo.yaml
```diff
@@ -0,0 +1,13 @@
+# This is necessary because ingress doesn't work across namespaces boundaries,
```
anthonylasso, line 1: good to know. also missing the license tag?

charts/featureform/values.yaml
```diff
@@ -57,6 +57,16 @@ dashboardUrl: ""
 # If true, restarts pods on update even if no changes have been made
 restartOnUpdate: false
 
+argo:
+  # Enables path forwarding of /argocd/ to the Argo web server. Note this still requires installation of Argo,
+  # done outside of this chart itself as it is a global service across the cluster.
+  enabled: false
```
anthonylasso, line 63: couple of q's: 
- is the argo update an all-or-nothing prop? is it easy to revert back and forth between argo being enabled or not (given the flag in the values file)

- we have a few jobs resources that we use in the app, do we use them normally or do they need to be "argo-fied" too.

charts/featureform/templates/api-server/rollout.yaml
ff-kamal: That's correct, in this case there's a manual step before switching over to the green deployment, as outlined here https://argo-rollouts.readthedocs.io/en/stable/generated/kubectl-argo-rollouts/kubectl-argo-rollouts_promote/. Figured it would be a safer default to allow customers to switch at whatever point they're comfortable.

anthonylasso: I agree that's probably the better approach to avoid hiccups between the 2 versions.

anthonylasso: resolve at will.

charts/featureform/values.yaml
anthonylasso: Gotcha, ok I'll pass along the jobs to you, although it's possible we might change how the data_loader works in the future.

PR #1504, opened 2024-11-12
Cleanup cert management in Helm charts

ff-kamal:
Cleanup some of the cert management around the Helm charts. This includes:

- Removing hardcoded secret names (and use the previously defined one in config)
- Add config for the endpoint of Let's Encrypt that we want to target, for testing infrastructure purposes
- Grouping cert configs together.

Note: These changes require backwards incompatible changes (as we are changing the location of some of the configuration values to group them together). If we feel like this would be an unnecessary burden, we can change back the location of the configuration values and leave the rest of the changes.

### Select type(s) of change

- [ ] Documentation update
- [x] Cleanup
# Checklist:

### DIFF
--- a/charts/featureform/templates/ingress/issuer.yaml
+++ b/charts/featureform/templates/ingress/issuer.yaml
@@ -5,14 +5,19 @@
 #  Copyright 2024 FeatureForm Inc.
 #
 
-{{ if .Values.publicCert }}
+{{ if .Values.cert.publicCert }}
 apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
-  name: letsencrypt-prod
+  name: letsencrypt
 spec:
   acme:
-    server: https://acme-v02.api.letsencrypt.org/directory
+    server: {{
+              ternary
+              "https://acme-v02.api.letsencrypt.org/directory"
+              "https://acme-staging-v02.api.letsencrypt.org/directory"
+              .Values.cert.letsencryptProd
+            }}
     email: sterling@featureform.com
     privateKeySecretRef:
       name: account-key-prod
--- a/charts/featureform/values.yaml
+++ b/charts/featureform/values.yaml
@@ -20,16 +20,25 @@ repository: "featureformenterprise"
 # Global pull policy
 pullPolicy: "Always"
 
-# If a public certificate should be generated automatically through lets encrypt and used
-publicCert: false
+cert:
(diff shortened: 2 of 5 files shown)


charts/featureform/templates/ingress/issuer.yaml
```diff
@@ -5,19 +5,24 @@
...
 apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
-  name: letsencrypt-prod
+  name: letsencrypt
 spec:
   acme:
-    server: https://acme-v02.api.letsencrypt.org/directory
+    server: {{
+              ternary
```
anthonylasso, line 16: q: how do these options work? is it just a set of cert servers?

charts/featureform/values.yaml
```diff
@@ -20,16 +20,25 @@ repository: "featureformenterprise"
...
+  # Will automatically create a self-signed cert that can be used if enabled
+  selfSignedCert: true
 
-# Note: If post publicCert and selfSignedCert are disabled, an existing certificate will need to be supplied
+  # Whether to use the prod letsencrypt endpoint or the staging. The production endpoint returns
+  # certificates that are trusted be well-known Certificate Authorities, however they have a low
+  # request limit (approximately 5 / week at the time of writing). During testing of infrastructure,
+  # this might be too low, so they also have a staging endpoint, with a much higher limit, but
+  # by default they're not trusted by any well-known CA (but you can add the certs manually).
+  # See https://letsencrypt.org/docs/staging-environment/
```
anthonylasso, line 35: note: that limit restriction _might_ be ok for now. we don't normally stand up or tear down the internal box as much, not even the nightly box saw 5 times a week cycles.

anthonylasso (approved):

anthonylasso:
The compatibility issue shouldn't be a big deal with new installs from the OSS repo. Anyone else upgrading will just need to take note of the option changes and edit their values file accordingly.

charts/featureform/templates/ingress/issuer.yaml
ff-kamal: That's correct, it's a couple of hardcoded cert servers, with the staging one specified at https://letsencrypt.org/docs/staging-environment/. The production one is just the normal one we've been using before.

anthonylasso: Gotcha make sense.

PR #1515, opened 2024-11-21
Feature: adds session_params to Snowflake config

epps:
# Description

This PR introduces the ability for users to provider [session parameters](https://docs.snowflake.com/en/sql-reference/sql/alter-session) to a Snowflake provider to ensure these params are set when establishing a connection (e.g. `query_tag`).

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/client/src/featureform/enums.py
+++ b/client/src/featureform/enums.py
@@ -322,3 +322,47 @@ class Initialize(Enum):
             raise ValueError(f"Initialize value not supported: {value}")
         except AttributeError:
             raise ValueError(f"Initialize value required: received {value}")
+
+
+class SnowflakeSessionParamKey(Enum):
+    ABORT_DETACHED_QUERY = "abort_detached_query"
+    AUTOCOMMIT = "autocommit"
+    BINARY_INPUT_FORMAT = "binary_input_format"
+    BINARY_OUTPUT_FORMAT = "binary_output_format"
+    DATE_INPUT_FORMAT = "date_input_format"
+    DATE_OUTPUT_FORMAT = "date_output_format"
+    ERROR_ON_NONDETERMINISTIC_MERGE = "error_on_nondeterministic_merge"
+    ERROR_ON_NONDETERMINISTIC_UPDATE = "error_on_nondeterministic_update"
+    GEOGRAPHY_OUTPUT_FORMAT = "geography_output_format"
+    HYBRID_TABLE_LOCK_TIMEOUT = "hybrid_table_lock_timeout"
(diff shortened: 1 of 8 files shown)


client/src/featureform/resources.py
```diff
@@ -1094,8 +1095,17 @@ class SnowflakeConfig:
     warehouse: str = ""
     role: str = ""
     catalog: Optional[SnowflakeCatalog] = None
+    session_params: Optional[Dict[str, str]] = None
 
     def __post_init__(self):
+
+        if self.session_params:
+            for key, _ in self.session_params.items():
+                if not SnowflakeSessionParamKey.validate_key(key):
```
simba-git, line 1104: Should this be done in Go? I actually dont mind it in the client, but just curious

client/src/featureform/enums.py
```diff
@@ -322,3 +322,47 @@ def from_string(cls, value):
             raise ValueError(f"Initialize value not supported: {value}")
         except AttributeError:
             raise ValueError(f"Initialize value required: received {value}")
+
+
+class SnowflakeSessionParamKey(Enum):
```
simba-git, line 327: Rough

simba-git (approved):

client/src/featureform/resources.py
epps: I waffled on this. I decided to go with a warning instead of an exception to avoid blocking a user down the line in the event Snowflake adds a session param that we don't take into account. I'm happy to move validation to the backend and fail. Thoughts?

client/src/featureform/register.py
```diff
@@ -3382,6 +3382,7 @@ def register_snowflake(
         tags: List[str] = [],
         properties: dict = {},
         catalog: Optional[SnowflakeCatalog] = None,
+        session_params: Optional[Dict[str, str]] = None
```
aolfat, line 3385: should it be Dict[SnowflakeSessionParamKey, str]]?

aolfat: we can even make it a dataclass and get the typing right too?

client/src/featureform/resources.py
aolfat: my thots, since the only way we're interacting with our backend is python client it's probz fine to have validation on the client, but if we were sending this from dxboard as well we'd want backend validation. probz doesn't hurt to keep it here to save a nwork call lol

client/src/featureform/enums.py
aolfat: lol

PR #1525, opened 2024-12-06
Add support for Goose migrations

ff-kamal:
# Description

- Adds a new folder where we'll store db migrations.
- Create a Dockerfile with an image that can be run as a job to 'up' migrate the db.
- Creates Helm pre-install, pre-upgrade hooks for this new image.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced a database migration tool integrated into the deployment process for PostgreSQL.
	- Added a new configuration for specifying the migration image in the Helm chart.
	- Enhanced management of PostgreSQL credentials with a connection string in Kubernetes Secrets.
	- Added a new job in the GitHub Actions workflow for building Docker images for database migrations.
	- New Minikube-related commands for building and managing Docker images.

- **Chores**
	- Created a Dockerfile for the database migration tool using Goose.
	- Configured a Kubernetes Job for executing database migrations during deployment phases.
	- Added an entrypoint script for managing database migrations.

### DIFF
--- a/charts/featureform/templates/db/upgrade.yaml
+++ b/charts/featureform/templates/db/upgrade.yaml
@@ -28,5 +28,8 @@ spec:
             - name: GOOSE_DRIVER
               value: postgres
             - name: GOOSE_DBSTRING
-              value: postgres://{{ .Values.psql.user }}:{{ .Values.psql.password }}@{{ .Values.psql.host }}:{{ .Values.psql.port }}/{{ .Values.psql.db }}
+              valueFrom:
+                secretKeyRef:
+                  name: psql-secret-literal
+                  key: connection-string
 {{ end }}
--- a/charts/featureform/templates/db/upgrade.yaml
+++ b/charts/featureform/templates/db/upgrade.yaml
@@ -22,7 +22,7 @@ spec:
       restartPolicy: Never
       containers:
         - name: db-migration-up
-          image: "{{ .Values.repository | default .Values.repository }}/{{ .Values.psql.migration.image.name }}:{{ .Values.versionOverride | default .Chart.AppVersion }}"
+          image: "{{ .Values.repository }}/{{ .Values.psql.migration.image.name }}:{{ .Values.versionOverride | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.pullPolicy }}
           env:
             - name: GOOSE_DRIVER
--- a/charts/featureform/templates/db/upgrade.yaml
+++ b/charts/featureform/templates/db/upgrade.yaml
@@ -1,0 +1,32 @@
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+#  Copyright 2024 FeatureForm Inc.
+#
+
+{{ if eq .Values.stateProvider "psql" }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: featureform-db-migration-up
+  labels:
+    chart: featureform
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: db-migration-up
+          image: "{{ .Values.repository | default .Values.repository }}/{{ .Values.psql.migration.image.name }}:{{ .Values.versionOverride | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.pullPolicy }}
+          env:
+            - name: GOOSE_DRIVER
+              value: postgres
+            - name: GOOSE_DBSTRING
+              value: postgres://{{ .Values.psql.user }}:{{ .Values.psql.password }}@{{ .Values.psql.host }}:{{ .Values.psql.port }}/{{ .Values.psql.db }}
+{{ end }}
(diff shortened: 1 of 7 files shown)


charts/featureform/templates/db/upgrade.yaml
```diff
@@ -0,0 +1,32 @@
...
+      restartPolicy: Never
+      containers:
+        - name: db-migration-up
+          image: "{{ .Values.repository | default .Values.repository }}/{{ .Values.psql.migration.image.name }}:{{ .Values.versionOverride | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.pullPolicy }}
+          env:
+            - name: GOOSE_DRIVER
+              value: postgres
+            - name: GOOSE_DBSTRING
+              value: postgres://{{ .Values.psql.user }}:{{ .Values.psql.password }}@{{ .Values.psql.host }}:{{ .Values.psql.port }}/{{ .Values.psql.db }}
```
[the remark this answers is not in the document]
ff-kamal: This is probably sound advice actually, will push out a new commit pulling the secret we already have for the DB.

db/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.21 AS builder
+
+WORKDIR /app
+
+RUN GOBIN=/app go install github.com/pressly/goose/v3/cmd/goose@v3.23.0
+
+FROM ubuntu:22.04
+
+ENV GOOSE_DRIVER=postgres
+ENV GOOSE_DBSTRING='postgres://postgres:password@localhost:5432/postgres?sslmode=disable'
```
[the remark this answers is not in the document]
ff-kamal: This is just a default for local development, so unless there are major objections will leave this in.

charts/featureform/templates/db/upgrade.yaml
```diff
@@ -0,0 +1,32 @@
...
+    chart: featureform
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
```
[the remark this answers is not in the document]
ff-kamal: Will need to do this for all deployments at a separate time.

charts/featureform/values.yaml
```diff
@@ -104,6 +104,9 @@ psql:
   # If true, automatically creates a secret stored in templates/secrets/postgres.yaml.
   # If false, does not apply the template. The template can be created manually
   createSecret: false
+  migration:
+    image:
+      name: db-migration-up
```
anthonylasso, line 109: think you need to update the image github actions as well to include the new dockerfile.

charts/featureform/templates/db/upgrade.yaml
```diff
@@ -0,0 +1,35 @@
...
+        - name: db-migration-up
+          image: "{{ .Values.repository | default .Values.repository }}/{{ .Values.psql.migration.image.name }}:{{ .Values.versionOverride | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.pullPolicy }}
```
anthonylasso, line 26: probably need to include [ttlSecondsAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) so the job doesn't stop the next helm upgrade from completing, unless goose auto deletes it?

anthonylasso:
q: how are we handling migration reverts? I assume there's a `down` feature that goose provides?

charts/featureform/values.yaml
anthonylasso: also look at `minikube_install_featureform` in the `makefile` as well.

charts/featureform/templates/db/upgrade.yaml
ff-kamal: The `"helm.sh/hook-delete-policy": hook-succeeded` annotation does, but that raises a question whether we want to also delete it if it failed? I'd think no so someone can go and debug it, but interested in hearing feedback.

anthonylasso: it's a good question, because we'd at least like to see the logs in case it does blow up. can it be conditional? if the job succeeds, then delete, otherwise keep the job alive so we can see the logs?

simba-git: I think always delete it, but it should fail the deployment or something right

charts/featureform/templates/db/upgrade.yaml
```diff
@@ -0,0 +1,35 @@
...
+          env:
+            - name: GOOSE_DRIVER
```
simba-git, line 28: Don't do this, use the ones we use to set postgres in other places

charts/featureform/templates/db/upgrade.yaml
```diff
@@ -0,0 +1,35 @@
...
+              value: postgres
+            - name: GOOSE_DBSTRING
```
simba-git, line 30: Also how do we decide which version to upgrade or downgrade to?

ff-kamal: Just to the latest migration that is present, but do we want that to be configurable?

charts/featureform/templates/db/upgrade.yaml
ff-kamal: So to be clear, it currently is conditional (so it's deleted if it's successful bc the annotation mentioned above is applied).

It should also fail the deployment, but we'd have no logs as to why the job failed, which is a situation we'd want to avoid. I think that the current approach of deleting only successful job executions should be good, unless there's major pushback.

anthonylasso (approved):

PR #1533, opened 2024-12-17
Iceberg Proxy Part 2

anthonylasso:
# Description

This PR is the iceberg proxy for our application's cluster. It will transmit requests from various callers (go modules, python client) and will pass-through the requests to the streamer. 

closes https://github.com/featureform/featureform-enterprise/issues/1550

`Proxy Benchmark`
image

Notes:
The proxy adds about ~1.5 seconds of latency vs directly using the streamer, however that's consistent with the overhead from the additional grpc initialization and calls. Overall performance seems reasonable and memory usage appears minimal.

Todo: in the near future will test out with an extremely larger dataset (1 million or more record).

### Select type(s) of change

- [ ] Documentation update

# Checklist:


charts/featureform/templates/streamer_proxy/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+      {{ if .Values.restartOnUpdate }}
+      annotations:
+        timestamp: {{ now | quote }}
+      {{ end }}
+    spec:
+      imagePullSecrets:
+        - name: {{ .Values.k8sImagePullSecret }}
+      containers:
+      - name: iceberg-proxy
+        image: "{{ .Values.repository | default .Values.repository }}/{{ .Values.icebergProxy.image.name }}:{{ .Values.versionOverride | default .Chart.AppVersion }}"
```
ff-kamal, line 32: `{{ .Values.repository | default .Values.repository }}` is a no-op to default to `Values.repository`. I notice this issue with a few other templates, can bring up a separate PR to fix the other ones.

streamer_proxy/main.go
```diff
@@ -0,0 +1,92 @@
...
+		if sendErr != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func main() {
+	baseLogger := logging.NewLogger("iceberg-proxy")
+	serverAddress := "0.0.0.0:8086"
```
ff-kamal, line 68: Pull from environment, and use `8086` as default.

streamer_proxy/main.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	for {
+		flightData, err := flightStream.Recv()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				gps.logger.Infof("Reached the end of the stream")
+				break
+			}
+			gps.logger.Infof("An error occurred: %v", err)
```
ff-kamal, line 53: nit: `Errorf`

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/featureform
 
-go 1.21
+go 1.22
+
+toolchain go1.23.3
```
ff-kamal, line 5: Any reason we're setting a newer toolchain? And if so, any reason not to just bump the version to 1.23?

streamer_proxy/main.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		gps.logger.Fatalf("Failed to connect to the iceberg-streamer: %v", err)
+		return err
+	}
+	defer client.Close()
+
+	// fetch and pass stream back to the caller
+	flightStream, err := client.DoGet(context.Background(), ticket)
+	if err != nil {
+		gps.logger.Infof("Error fetching the data from the iceberg-streamer: %v", err)
```
ff-kamal, line 42: nit: `Errorf`.

charts/featureform/templates/streamer_proxy/deployment.yaml
anthonylasso: good eye. too much copy-pasta 🍝

ff-kamal:
General question, but not sure the method of benchmarking that you used, if not already included it might be useful to put include it? If it's very rudimentary and prone to easily breaking wouldn't worry about it too much.

go.mod
anthonylasso: Was modified when pulling the arrow go library v17. But...I'm going to try and revert to `1.21` and see if it still works, since their repo's [go.mod](https://github.com/apache/arrow/blob/go/v17.0.0/go/go.mod) apparently only needs 1.21.

anthonylasso:
> General question, but not sure the method of benchmarking that you used, if not already included it might be useful to put include it? If it's very rudimentary and prone to easily breaking wouldn't worry about it too much.

Check this out, go has a pretty sweet [benchmarking tool](https://gobyexample.com/testing-and-benchmarking). Essentially it wraps the code you want to test and fires of N instances while keeping track of all the metrics and averages.

go.mod
```diff
@@ -10,10 +10,8 @@ require (
 	github.com/aws/aws-sdk-go v1.50.36 // indirect
 	github.com/aws/aws-sdk-go-v2/config v1.27.11
 	github.com/aws/aws-sdk-go-v2/service/emr v1.39.5
-	github.com/casbin/casbin/v2 v2.20.2
```
anthonylasso, line 13: note: go mod tidy

streamer_proxy/main.go
```diff
@@ -0,0 +1,92 @@
...
+	logger          logging.Logger
+}
+
+func (gps *GoProxyServer) DoGet(ticket *flight.Ticket, stream flight.FlightService_DoGetServer) error {
+	gps.logger.Infof("Received request, forwarding to iceberg-streamer at: %v", gps.streamerAddress)
+	insecureOption := grpc.WithTransportCredentials(insecure.NewCredentials())
+	client, err := flight.NewClientWithMiddleware(gps.streamerAddress, nil, nil, insecureOption)
+	if err != nil {
+		gps.logger.Fatalf("Failed to connect to the iceberg-streamer: %v", err)
+		return err
```
ff-kamal, line 35: `Fatalf` calls `os.Exit`, so this would halt the program. I imagine this is probably not what we want to do, so keep it at `Errorf`.

If it is actually what we want, would recommend deleteing `return err` to make it clear nothing is returning.

ff-kamal (approved):

streamer_proxy/main.go
anthonylasso: very good catch, and yes definitely want the app to keep running since it already passed main's initialization steps.

go.mod
anthonylasso: cleaned up the mod file and 1.21 works just fine.

PR #1564, opened 2025-01-27
Bugfix/sort task page (#1691)

RiddhiBagadiaa:
# Description

This PR allows the user to find the correct name/variant row on the Task Run page when they navigate from the individual resource page using the task run button

### Select type(s) of change

- [ ] Documentation update

# Checklist:

## Release Notes

- **New Features**
  - Added dynamic routing for the task runs page.
  - Introduced a new filter panel for search results.

- **Improvements**
  - Enhanced task runs search functionality with variant-based filtering.
  - Updated task runs page to support dynamic name and variant parameters.
  - Improved component props handling and data flow.

- **Technical Updates**
  - Updated metadata structure to support more flexible search options.

### DIFF
--- a/dashboard/src/components/searchresults/FilterPanel.js
+++ b/dashboard/src/components/searchresults/FilterPanel.js
@@ -1,0 +1,98 @@
+import ArrowDropDownIcon from '@mui/icons-material/ArrowDropDown';
+import {
+  AccordionDetails,
+  AccordionSummary,
+  Checkbox,
+  FormGroup,
+  Typography,
+  Accordion,
+  Divider,
+  FormControlLabel,
+  Box,
+} from '@mui/material';
+import React, { useState } from 'react';
+import { styled } from '@mui/system';
+import { FEATUREFORM_DARK_GRAY } from '../../styles/theme';
+
+const FILTER_PANEL_GREY = "#F9F9FA";
+
+const STATUS_OPTIONS = [
+    { key: 'No Status', value: 'NO_STATUS' },
+    { key: 'Created', value: 'CREATED' },
+    { key: 'Pending', value: 'PENDING' },
+    { key: 'Ready', value: 'READY' },
+    { key: 'Failed', value: 'FAILED' },
+    { key: 'Running', value: 'RUNNING' },
+    { key: 'Cancelled', value: 'CANCELLED' },
+];
+
+const filterItemTxtSize = '0.875rem';
+
+const StyledFormControlLabel = styled(FormControlLabel)({
+  '& .MuiFormControlLabel-label': {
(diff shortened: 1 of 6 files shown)


dashboard/src/components/searchresults/FilterPanel.js
```diff
@@ -0,0 +1,98 @@
+import ArrowDropDownIcon from '@mui/icons-material/ArrowDropDown';
```
ff-kamal, line 1: Add license.

dashboard/src/components/tasks/tableDataWrapper.test.js
```diff
@@ -45,6 +45,7 @@ describe('Task table data wrapper tests', () => {
   const CLEAR_ICON_ID = 'clearIcon';
   const DEFAULT_PARAMS = {
     searchText: '',
+    variantSearch: '',
```
ff-kamal, line 48: We should probably have a test that tests a non-empty `variantSearch`.

metadata/dashboard/dashboard_metadata.go
```diff
@@ -2558,7 +2559,26 @@ func (m MetadataServer) getTaskRunsQuery(requestBody PaginateRequest, isCount bo
 		}
 	}
 
-	if requestBody.SearchText != "" {
+	if requestBody.VariantSearch != "" && requestBody.SearchText != "" {
```
ff-kamal, line 2562: Just to confirm that this is the behavior we want. Do we expect a use case where only the `VariantSearch` is populated to be a valid search case, and handle that, or just ignore it as we currently do?

anthonylasso: This is probably fine for now. Later on if we add a variant input, we can have separate logic to do variant-only searches.

dashboard/src/components/tasks/tableDataWrapper.test.js
anthonylasso: Yes good idea, a test that sends both `name` and `variant` can verify that both items are sent as part of the initial request. I'd add it to this PR.

dashboard/src/components/tasks/tableDataWrapper.test.js
```diff
@@ -149,4 +150,27 @@ describe('Task table data wrapper tests', () => {
     //then: the api is invoked twice. on initial load and refresh
     expect(dataAPIMock.getTaskRuns).toHaveBeenCalledTimes(2);
   });
+
+  test('Initial load sends both name and variant from router query params', async () => {
```
anthonylasso, line 154: good test 🏆

anthonylasso (approved):

PR #1583, opened 2025-02-06
Bugfix/fix flaky tests/simba

simba-git:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

## Summary by CodeRabbit

- **New Features**
	- Spark job execution now automatically retries on transient issues for improved reliability.
	- Health check operations use dynamic file isolation and enhanced cleanup logging to reduce conflicts and boost overall system robustness.
	- Extended default timeout for DynamoDB table readiness to improve operation reliability.
- **Bug Fixes**
	- Improved error handling during health check path deletion to ensure better logging and traceability.
	- Enhanced error reporting in lock management to capture multiple errors during concurrent operations.


provider/databricks.go
```diff
@@ -143,28 +143,77 @@ func (db *DatabricksExecutor) RunSparkJob(cmd *spark.Command, store SparkFileSto
...
+		func() error {
+			deadline, ok := ctx.Deadline()
+			if !ok {
+				errMsg := "Deadline not set on context, exiting infinite loop running databricks job"
+				logger.Error(errMsg)
+				// Do not loop infinitely without a timeout
+				return re.Unrecoverable(fferr.NewInternalErrorf(errMsg))
+			}
+			retryTimeout := retries.Timeout[jobs.Run](time.Until(deadline))
+			_, err = db.client.Jobs.RunNowAndWait(ctx, jobs.RunNow{
```
ff-kamal: Nit: RunNowAndWait deprecated, replace with `WaitGetRunJobTerminatedOrSkipped`

provider/databricks.go
```diff
@@ -143,28 +143,77 @@ func (db *DatabricksExecutor) RunSparkJob(cmd *spark.Command, store SparkFileSto
...
+	return re.Do(
+		func() error {
+			deadline, ok := ctx.Deadline()
+			if !ok {
+				errMsg := "Deadline not set on context, exiting infinite loop running databricks job"
+				logger.Error(errMsg)
+				// Do not loop infinitely without a timeout
+				return re.Unrecoverable(fferr.NewInternalErrorf(errMsg))
+			}
+			retryTimeout := retries.Timeout[jobs.Run](time.Until(deadline))
```
ff-kamal, line 173: Just to be clear, `deadline` doesn't represent a retry timeout, but the full timeout including all retries, right? If so, small nit: to change the naming of it to something like `jobTimeout`.

provider/databricks.go
```diff
@@ -143,28 +143,77 @@ func (db *DatabricksExecutor) RunSparkJob(cmd *spark.Command, store SparkFileSto
...
+		},
+		re.Context(ctx),                   // Stop retries when context times out
+		re.Attempts(0),                    // Infinite retries (until context cancels)
+		re.LastErrorOnly(true),            // Only return the last error
+		re.DelayType(re.CombineDelay(   // Use exponential backoff + jitter
+			re.BackOffDelay,              // Exponential backoff
+			re.RandomDelay,               // Random jitter
+		)),
+		re.Delay(5*time.Second),           // Initial delay of 5 seconds
+		re.MaxDelay(3*time.Minute),        // Cap max delay to 5 min
```
ff-kamal, line 205: Comment says 5 minutes, code says 3 minutes.

provider/databricks.go
ff-kamal: Also, is there any reason to set this here if the overall context we pass to `re` already handles it?

provider/databricks.go
simba-git: nice catch

provider/databricks.go
simba-git: So the RunNowAndWait has its own retry mechanism but it doesn't retry on some things that we want it to retry on like hte driver errors

simba-git: This is sort of a mash up of their retry logic and our retry logic

provider/databricks.go
simba-git: nice catch

provider/databricks.go
ff-kamal: Are there things it retries that re doesn't? Why can't we just get rid of the databricks retry entirely?

simba-git: Turns out that retry is part of their deprecated func anyway. My original though was that they might have smarter retry mechanisms based on knowing how their system works. But a moot point now

provider/databricks.go
```diff
@@ -143,28 +143,77 @@ func (db *DatabricksExecutor) RunSparkJob(cmd *spark.Command, store SparkFileSto
...
+	)
+}
+
+func (db *DatabricksExecutor) isEphemeralError(err error) bool {
+	// This happens when the driver goes OOM sometimes
+	if strings.Contains(err.Error(), "Could not reach driver of cluster") {
```
anthonylasso, line 215: nit: seems a bit flimsy, do we know the error _type_ coming in?

ghost (approved):

anthonylasso (approved):

provider/databricks.go
simba-git: not sure, the problem is I can't always make this error happen on command and the go library doesn't document how to check for this particular error

caldempsey:
Hi, not sure why I was mentioned but I updated my Databricks-go client with an example in the issue I created that pulls data at about 100k rows/s

PR #1611, opened 2025-03-10
SQL Types Serialize with Datasets

aolfat:
# Description

### Select type(s) of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/fftypes/types.go
+++ b/fftypes/types.go
@@ -321,6 +321,8 @@ type Value struct {
 type NativeType string
 type ColumnName string
 
+type NativeToValueTypeMapper map[NativeType]ValueType
+
 type Schema struct {
 	Fields []ColumnSchema
 	// todo: can include more state or behavior, etc.
--- a/provider/dataset/dataset.go
+++ b/provider/dataset/dataset.go
@@ -16,7 +16,7 @@ import (
 	pl "github.com/featureform/provider/location"
 )
 
-// Dataset is the base interface required by most of the
+// SqlDataset is the base interface required by most of the
 // OfflineStore interface. It provides a location, iterator,
 // and schema
 type Dataset interface {
@@ -25,7 +25,7 @@ type Dataset interface {
 	Schema() (types.Schema, error)
 }
 
-// WriteableDataset is a Dataset that you can write to. For some datasets,
+// WriteableDataset is a SqlDataset that you can write to. For some datasets,
 // especially file based one, writing single rows is expensive. It's
 // preferable to write as many things as possible in a single batch.
 type WriteableDataset interface {
@@ -33,7 +33,7 @@ type WriteableDataset interface {
 	WriteBatch(Context, []types.Row) error
 }
 
-// SizedDataset is a Dataset where the size can be cheaply calculated.
+// SizedDataset is a SqlDataset where the size can be cheaply calculated.
 type SizedDataset interface {
 	Dataset
 	Len() (int64, error)
@@ -42,7 +42,7 @@ type SizedDataset interface {
 // SegmentableDataset allows a user to easily iterate
 // an arbitrary segment of the dataset. Note that it
 // doesn't guarantee the end of the iterator is actually
-// in range of the dataset. You can try casting Iterator
+// in range of the dataset. You can try casting SqlIterator
 // to SizedIterator to see if the length is available.
 type SegmentableDataset interface {
 	Dataset
@@ -131,7 +131,7 @@ func (adapter *ChunkedDatasetAdapter) ChunkIterator(ctx Context, idx int) (Sized
 	}, nil
 }
 
-// Iterator is the generic interface to loop through any dataset in
+// SqlIterator is the generic interface to loop through any dataset in
 // Featureform.
 type Iterator interface {
 	Next(Context) bool
@@ -141,7 +141,7 @@ type Iterator interface {
 	Close() error
 }
 
-// SizedIterator is an Iterator where we can cheaply check
+// SizedIterator is an SqlIterator where we can cheaply check
 // the full length.
 type SizedIterator interface {
 	Iterator
--- a/provider/dataset/sql_dataset.go
+++ b/provider/dataset/sql_dataset.go
@@ -1,0 +1,203 @@
+package dataset
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/featureform/fferr"
+	types "github.com/featureform/fftypes"
+	"github.com/featureform/provider/location"
+)
+
+type SqlDataset struct {
+	db       *sql.DB
+	location location.SQLLocation
+	schema   types.Schema
+	mapper   types.NativeToValueTypeMapper
+	limit    int
+}
+
+func NewSqlDataset(db *sql.DB, location location.SQLLocation, limit int) (SqlDataset, error) {
+	var lmt = limit
+	if limit == 0 {
+		lmt = -1
+	}
+	schema, err := getSchema(db, location.GetTable())
+	if err != nil {
+		return SqlDataset{}, err
+	}
+	return SqlDataset{location: location, schema: schema, limit: lmt}, nil
+}
+
+func getSchema(db *sql.DB, tableName string) (types.Schema, error) {
+	qry := "SELECT column_name, data_type FROM information_schema.columns WHERE table_name = ? AND table_schema = CURRENT_SCHEMA() ORDER BY ordinal_position"
+
+	rows, err := db.Query(qry, tableName)
+	if err != nil {
+		wrapped := fferr.NewExecutionError("SQL", err)
+		wrapped.AddDetail("table_name", tableName)
+		return types.Schema{}, wrapped
+	}
+	defer rows.Close()
+
+	fields := make([]types.ColumnSchema, 0)
+	for rows.Next() {
+		var columnName, dataType string
+		if err := rows.Scan(&columnName, &dataType); err != nil {
+			wrapped := fferr.NewExecutionError("SQL", err)
+			wrapped.AddDetail("table_name", tableName)
+			return types.Schema{}, wrapped
+		}
+
+		column := types.ColumnSchema{
+			Name:       types.ColumnName(columnName),
+			NativeType: types.NativeType(dataType),
+		}
+		fields = append(fields, column)
+	}
+
+	if err := rows.Err(); err != nil {
+		wrapped := fferr.NewExecutionError("SQL", err)
+		wrapped.AddDetail("table_name", tableName)
+		return types.Schema{}, wrapped
+	}
+
+	schema := types.Schema{
+		Fields: fields,
+	}
+
+	return schema, nil
+}
+
+func (ds SqlDataset) Location() location.Location {
+	return &ds.location
+}
+
+func (ds SqlDataset) Iterator(ctx context.Context) (Iterator, error) {
+	var query string
+	colNames := ds.schema.ColumnNames()
+	tableName := ds.location.GetTable()
+	if ds.limit == -1 {
+		query = fmt.Sprintf("SELECT %s FROM %s", colNames, tableName)
+	} else {
+		query = fmt.Sprintf("SELECT %s FROM %s LIMIT %d", colNames, tableName, ds.limit)
+	}
+	rows, err := ds.db.Query(query)
+	if err != nil {
+		return nil, fferr.NewExecutionError("SQL", err)
+	}
+	return NewSqlIterator(rows, ds.mapper, ds.schema), nil
+}
+
+func (ds SqlDataset) Schema() (types.Schema, error) {
+	return ds.schema, nil
+}
+
+type SqlIterator struct {
+	rows          *sql.Rows
+	currentValues types.Row
+	mapper        types.NativeToValueTypeMapper
+	schema        types.Schema
+	err           error
+}
+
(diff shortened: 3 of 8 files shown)


provider/biguery/value_converter.go
```diff
@@ -0,0 +1,267 @@
...
+		default:
+			wrapped := fferr.NewTypeError(t.String(), value, nil)
+			return nil, wrapped
+		}
+	default:
+		wrapped := fferr.NewInternalErrorf("Type conversion not supported")
+		wrapped.AddDetail("type", fmt.Sprintf("%v", t))
+		return nil, wrapped
+	}
+}
```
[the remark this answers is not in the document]
aolfat: this is there is the value_tests

provider/snowflake/types.go
```diff
@@ -0,0 +1,46 @@
...
+import . "github.com/featureform/fftypes"
+
+var TypeMap = NativeToValueTypeMapper{
+	// Integer types
+	"INTEGER":  Int32,
+	"BIGINT":   Int64,
+	"SMALLINT": Int32,
+
+	// Floating point types
+	"NUMBER":           Float64,
```
aolfat, line 12: maybe map to a common value converter

fftypes/types.go
```diff
@@ -32,10 +32,10 @@ const (
 	Bool      ScalarType = "bool"
 	Timestamp ScalarType = "time.Time"
 	Datetime  ScalarType = "datetime"
+	Unknown   ScalarType = "unknown"
```
aolfat, line 35: keep this but don't use it

provider/clickhouse/value_converter.go
```diff
@@ -0,0 +1,106 @@
...
+func convertScalar(t ValueType, value any) (any, error) {
+	if value == nil {
+		return nil, nil
+	}
+
+	switch t {
+	case Int32:
+		return int(value.(int32)), nil
+	case Int64:
+		if intValue, ok := value.(int64); ok && intValue >= math.MinInt && intValue <= math.MaxInt {
```
aolfat, line 65: check ok's

provider/clickhouse/types.go
```diff
@@ -0,0 +1,31 @@
+package clickhouse
+
+import . "github.com/featureform/fftypes"
```
aolfat, line 3: don't use '.' imports

provider/bigquery/value_converter.go
```diff
@@ -0,0 +1,166 @@
...
+		convertedValue, err := types.ConvertNumberToFloat64(value)
+		if err != nil {
+			return types.Value{}, err
+		}
+		return types.Value{
+			NativeType: nativeType,
+			Type:       types.Float64,
+			Value:      convertedValue,
+		}, nil
+
```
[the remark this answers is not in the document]
aolfat: going to change this a bit when I add the integration test

provider/dataset/dataset.go
```diff
@@ -131,18 +131,17 @@ func (adapter *ChunkedDatasetAdapter) ChunkIterator(ctx Context, idx int) (Sized
 	}, nil
 }
 
-// Iterator is the generic interface to loop through any dataset in
-// Featureform.
+// Iterator is the generic interface to loop through any dataset in Featureform.
 type Iterator interface {
-	Next(Context) bool
+	Next() bool
```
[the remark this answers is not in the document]
aolfat: tests would fail if this wasn't fixed

fftypes/value_converter.go
```diff
@@ -0,0 +1,298 @@
...
+			return time.Time{}, wrapped
+		}
+		return time.Unix(unixTime, 0).UTC(), nil
+	default:
+		wrapped := fferr.NewTypeError(Datetime.String(), v, nil)
+		return time.Time{}, wrapped
+	}
+}
+
+func ConvertVectorValue(value any) (ScalarType, any, error) {
```
ff-kamal, line 194: nit: Is this function expected to be used by anything other than Snowflake? If so, I'd remove explicit mentions of Snowflake from the error messages and comments.

provider/bigquery/value_converter.go
```diff
@@ -0,0 +1,166 @@
...
+import (
+	"github.com/featureform/fferr"
+	types "github.com/featureform/fftypes"
+	"github.com/featureform/logging"
+	"github.com/featureform/provider/provider_type"
+)
+
+// SupportedTypes is a set of all supported BigQuery types
+// NOTE: If you add a new type here, make sure to add a corresponding conversion function in ConvertValue
```
ff-kamal, line 11: Maybe a better way of doing this pattern is that you change the type of this map to `map[types.NativeType]types.Value` (i.e. the conversion functions in `ConvertValue` are stored in this map), and the presence of an entry implies that the type is supported. This ensures that the two don't go out of sync, and there's no possibility of forgetting to update one or the other.

fftypes/value_converter.go
aolfat: good catch

fftypes/types.go
```diff
@@ -51,6 +51,7 @@ var ScalarTypes = map[ScalarType]bool{
 	Bool:      true,
 	Timestamp: true,
 	Datetime:  true,
+	Unknown:   true,
```
simba-git, line 54: Don't put this here

fftypes/types.go
```diff
@@ -309,7 +310,7 @@ func (vt ValueTypeJSONWrapper) MarshalJSON() ([]byte, error) {
 type ColumnSchema struct {
 	Name       ColumnName
 	NativeType NativeType
-	Type       ValueType
+	//Type       ValueType // I dont' think we need this
```
simba-git, line 313: We do

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+type SqlIterator struct {
+	rows          *sql.Rows
+	currentValues types.Row
+	converter     types.ValueConverter[any]
+	schema        types.Schema
+	err           error
+	closed        bool
+	scanBuffers   []interface{} // Reusable buffer for SQL Scan
```
simba-git, line 160: use any instead of interface{}

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	return NewSqlIterator(rows, ds.converter, ds.schema), nil
+}
+
+func (ds SqlDataset) Schema() (types.Schema, error) {
+	return ds.schema, nil
+}
+
+func (ds SqlDataset) ColumnNames() []string {
```
simba-git, line 149: Kill this method, just have types.Schema have a ColumnNames method or something

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+	return types.Schema{Fields: fields}, nil
+}
+
+func (ds SqlDataset) Location() location.Location {
+	return &ds.location
+}
+
+func (ds SqlDataset) Iterator(ctx context.Context) (Iterator, error) {
+	colNames := ds.schema.ColumnNames()
+	cols := strings.Join(colNames[:], ", ")
```
simba-git, line 127: why are you doing colNames[:] here? just use colNames

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+	var query string
+	if ds.limit == -1 {
+		query = fmt.Sprintf("SELECT %s FROM %s", cols, loc)
+	} else {
+		query = fmt.Sprintf("SELECT %s FROM %s LIMIT %d", cols, loc, ds.limit)
+	}
+
+	rows, err := ds.db.QueryContext(ctx, query)
+	if err != nil {
+		return nil, fferr.NewInternalErrorf("Failed to execute query: %v", err)
```
simba-git, line 139: log

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return types.Schema{Fields: fields}, nil
+}
+
+func (ds SqlDataset) Location() location.Location {
+	return &ds.location
+}
+
+func (ds SqlDataset) Iterator(ctx context.Context) (Iterator, error) {
+	colNames := ds.schema.ColumnNames()
```
simba-git, line 126: Get the logger from ctx

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (ds SqlDataset) Schema() (types.Schema, error) {
+	return ds.schema, nil
+}
+
+func (ds SqlDataset) ColumnNames() []string {
+	return ds.schema.ColumnNames()
+}
+
+type SqlIterator struct {
```
simba-git, line 153: Context should be passed into the iterator

provider/dataset/in_memory_dataset.go
```diff
@@ -72,6 +72,10 @@ func (it *InMemoryIterator) Schema() (types.Schema, error) {
 	return types.Schema{}, nil
 }
 
+func (it *InMemoryIterator) Columns() []string {
+	return make([]string, 0)
```
ff-kamal, line 76: Is this right, shouldn't we actually return the columns it's initialized with? And to push on that a bit, shouldn't there be a test that fails with this implementation?

provider/dataset/in_memory_dataset.go
```diff
@@ -72,6 +72,10 @@ func (it *InMemoryIterator) Schema() (types.Schema, error) {
 	return types.Schema{}, nil
```
ff-kamal, line 72: Below's comment probably applies to this too even though you didn't modify it. This should really be returning the schema initialized on the dataset, no?

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,258 @@
...
+
+	// Corrected Query: Ensure both `table_name` and `table_schema` are matched
+	qry := `SELECT column_name, data_type 
+	        FROM information_schema.columns 
+	        WHERE table_name = ? 
+	        AND table_schema = ? 
+	        ORDER BY ordinal_position`
+
+	// Debugging print (remove in production)
+	fmt.Printf("Querying schema: %q, table: %q\n", schema, tblName)
```
ff-kamal, line 74: Reminder to remove.

provider/dataset/sql_dataset.go
```diff
@@ -0,0 +1,264 @@
...
+func (ds SqlDataset) Location() location.Location {
+	return &ds.location
+}
+
+func (ds SqlDataset) Iterator(ctx context.Context) (Iterator, error) {
+	logger := logging.GetLoggerFromContext(ctx)
+	colNames := ds.schema.ColumnNames()
+	cols := strings.Join(colNames, ", ")
+
+	loc := postgres.SanitizeLocation(ds.location)
```
ff-kamal, line 135: The fact that we tie sanitization to how Postgres does it isn't great, but I don't think we have a better solution until we refactor `Location` related stuff (ideally we'd just do something like `ds.Location()` which implicitly sanitizes, but that's not how it works right now). Can you add a `// TODO: ...` here for it (unless you can think of a better solution?).

provider/bigquery/value_converter.go
```diff
@@ -0,0 +1,178 @@
...
+	types "github.com/featureform/fftypes"
+	"github.com/featureform/logging"
+	"github.com/featureform/provider/provider_type"
+)
+
+var bqConverter = Converter{}
+
+func init() {
+	Register()
+}
```
ff-kamal, line 14: I think this is less clear than what we had before. Is there any reason why we can't have it just be a `map[types.NativeType] func(any)types.Value`, and in the `ConvertValue` function below we simply just do a lookup on the map and call the associated function on the `value` being passed in?

ff-kamal: Let's sync up tomorrow to just to make sure we're on the same page about this, and that I properly understand the problem here.

fftypes/value_converter.go
```diff
@@ -0,0 +1,298 @@
...
+	case int32:
+		return float64(casted), nil
+	case int64:
+		return float64(casted), nil
+	case float32:
+		return float64(casted), nil
+	case float64:
+		return casted, nil
+	case string:
+		return strconv.ParseFloat(casted, 64)
```
ff-kamal, line 121: Handle this the same way it's handled above in `Float32`, by wrapping the error around it.

fftypes/value_converter.go
```diff
@@ -0,0 +1,298 @@
...
+
+	if scalarType == "" || scalarType == Unknown {
+		err := fferr.NewInternalErrorf("Type conversion failed due to unknown scalar type")
+		return Unknown, nil, err
+	}
+
+	// Step 4: Convert to a typed slice based on inferred scalar type
+	var converted any
+	var err error
+	switch scalarType {
```
ff-kamal, line 235: What about the remaining scalar types (ex. `DateTime`)

fftypes/value_converter.go
```diff
@@ -0,0 +1,298 @@
...
+		return Unknown, nil, err
+	}
+
+	return scalarType, converted, nil
+}
+
+func convertList[T any](scalar ScalarType, values []any) ([]T, error) {
+	result := make([]T, len(values))
+	for i, value := range values {
+		casted, ok := value.(T)
```
ff-kamal, line 266: Do we want to cast here? Should we instead be using the convert scalar functions instead? If I'm misunderstanding the point of this function let me know (and if I am, I suggest we rename this so it's clear we're not convert'ing, we're just casting).

ff-kamal (approved):

PR #1622, opened 2025-03-25
Add startup probes to deployments

ff-kamal:
# Description

During an upgrade, we currently don't have any probes on any pods (besides the `api-server`). This means when we do an upgrade, Kubernetes will bring up a new pod and then immediately terminate the old one, assuming that the new one is able to serve traffic (where in most cases it isn't, since these containers take a bit of time to start up).

[Kubernetes startup probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) are the solution to this, where the probes first call into the container, and only when the probe returns successfully does Kubernetes then know that the pod is ready to serve traffic.

We utilize gRPC probes for most services, except in the case where some services don't use gRPC (such as `dashboard`), and in this case use HTTP probes. However, we implement both types for any relevant services still.

Additionally, created a new directory `tests/kubernetes/downtime`, which contains an example helper script to test whether a deployment has downtime during an upgrade or not.

### Select type(s) of change

- [x] Chore (non-breaking change that modifies some code)
- [ ] Documentation update

# Checklist:

## Summary by CodeRabbit

- **New Features**
  - Transitioned to a modern gRPC-based health check system for improved service monitoring.
  - Introduced dynamic startup probes in Kubernetes deployments to enhance container readiness.
  - Enhanced configuration options with environment-driven status ports and flexible container port settings.
  - Updated Docker images to include essential health resources.

- **Tests**
  - Added an automated downtime verification tool to monitor service availability during rolling updates.
  - Introduced a new gRPC service definition for health checks using Protocol Buffers.


tests/kubernetes/downtime/health.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package grpc.health.v1;
```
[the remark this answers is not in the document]
ff-kamal: This is the health proto as according to the grpc spec. We're not actually going to be using this anywhere, this is here to be used manually (ex. `grpcurl`) / get an idea of what the health check is.

api/api.go
```diff
@@ -1513,6 +1512,11 @@ func (serv *ApiServer) ServeOnListener(lis net.Listener) error {
 		grpc.KeepaliveParams(kasp),
 	}
 	grpcServer := grpc.NewServer(opt...)
+
+	healthServer := grpc_health.NewServer()
+	grpc_health_v1.RegisterHealthServer(grpcServer, healthServer)
+	healthServer.SetServingStatus("", grpc_health_v1.HealthCheckResponse_SERVING)
```
anthonylasso, line 1518: safe to not include the service name?

charts/featureform/templates/coordinator/deployment.yaml
```diff
@@ -97,19 +97,13 @@ spec:
...
-#          readinessProbe:
-#            httpGet:
-#              path: /
-#              port: http
           resources:
             {{- toYaml .Values.coordinator.resources | nindent 12 }}
+          startupProbe:
+            httpGet:
+              path: /
+              port: 8443
```
anthonylasso, line 109: should this not be configurable?

charts/featureform/templates/dashboard/deployment.yaml
```diff
@@ -41,3 +41,7 @@ spec:
                   name: featureform-env
             - name: FEATUREFORM_DEBUG_LOGGING
               value: {{ .Values.debug | quote }}
+          startupProbe:
+            httpGet:
+              path: /
+              port: 80
```
anthonylasso, line 47: similar to above this port is configurable from the values file

charts/featureform/templates/metadata/deployment.yaml
```diff
@@ -72,3 +72,6 @@ spec:
                   name: psql-secret-literal
                   key: password
           {{ end }}
+          startupProbe:
+            grpc:
+              port: 8080
```
anthonylasso, line 77: might want to check this too

health/server.go
```diff
@@ -0,0 +1,100 @@
...
+	// handles possible trailing slash
+	mux.HandleFunc("/status/", handleStatus)
+	// Health check endpoint will handle all /_ah/* requests
+	// e.g. /_ah/live, /_ah/ready and /_ah/lb
+	// Create separate routes for specific health requests as needed.
+	mux.HandleFunc("/_ah/", handleHealthCheck)
+	mux.HandleFunc("/", handleIndex)
+	// Add more routes as needed.
+
+	addr := fmt.Sprintf(":%s", port)
```
anthonylasso, line 79: should check for a zeroed port, and if so log an error (can't rely on a caller playing nice with this function)

health/server.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Set timeouts so that a slow or malicious client doesn't hold resources forever.
```
anthonylasso, line 81: good idea 🏆

health/server.go
```diff
@@ -0,0 +1,100 @@
...
+		IdleTimeout:  60 * time.Second,
+		Handler:      mux,
+		Addr:         addr,
+	}
+
+	logger.Infof("starting health check HTTP server on port %s", port)
+
+	go func() {
+		err := httpsSrv.ListenAndServe()
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
```
simba-git, line 94: how about in the other case of an error?

api/api.go
ff-kamal: Yup, the name is not required according the health spec (which then represents the overall health of the service, and not just a single service that's being hosted).

anthonylasso (approved):
lgtm, good job with the configurable fields.

PR #1630, opened 2025-04-10
Fix float conversion for Postgres

ff-kamal:
# Description

When a column type is `NUMERIC` in Postgres, the SQL interface we have will return it as a `[]uint8` when we scan it in. This is because `NUMERIC` is arbitrary-precision. Previously, we didn't handle this case, and a panic was thrown when attempting to convert the value.

### Select type(s) of change

- [ ] Documentation update

# Checklist:

## Summary by CodeRabbit

- **New Features**
  - Enhanced handling of `pgFloat` column type for improved numeric value conversion from byte arrays.
- **Tests**
  - Added a test case for verifying the conversion of byte arrays to numeric values.

### DIFF
--- a/provider/postgres.go
+++ b/provider/postgres.go
@@ -10,6 +10,7 @@ package provider
 import (
 	"database/sql"
 	"fmt"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


provider/postgres_test.go
```diff
@@ -112,6 +112,12 @@ func TestPostgresCastTableItemType(t *testing.T) {
 			typeSpec: pgFloat,
 			expected: 3.14,
 		},
+		{
+			name:     "Numeric column type",
```
aolfat, line 116: wonder if it's better to have a specific numeric type for now? this conversion still works for for the other floats right? I'll fix this in my stuff but just curious

aolfat (approved):

provider/postgres.go
```diff
@@ -269,6 +271,16 @@ func (q postgresSQLQueries) castTableItemType(v interface{}, t interface{}) inte
...
+		// If the column type is NUMERIC, the SQL interface will return the value as
+		// a []uint8 type. This is the ASCII-formatted value of the float, and is
+		// done because the NUMERIC type has arbitrary precision.
+		byteArray := v.([]uint8)
+		floatVal, err := strconv.ParseFloat(string(byteArray), 64)
+		if err != nil {
+			return nil
+		}
+		return floatVal
 	case pgString:
```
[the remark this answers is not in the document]
ff-kamal: It's fine for now, the whole logic is being revisited as part of the types project. This is just a stopgap.

provider/postgres_test.go
ff-kamal: I did change it, but apparently the SQL interface does actually use `interface {}` as the scan type for a few different floats, so it now makes more sense in my head just to group them all together.

